PR #3, opened 2026-04-01
Add worktree isolation and update stage pipeline

verveguy:
## Summary

- **Worktree isolation**: Each issue gets its own git worktree (`.fabrik/worktrees/issue-N/`) and branch (`fabrik/issue-N`) forked from main. Claude Code runs inside the worktree so multiple issues can be in flight without conflicting.
- **Updated stages**: Replace Triage/Design/Implement/Review with **Research → Plan → Implement → Review → Validate** pipeline, matching the project board columns.

## How worktrees work

```
.fabrik/worktrees/
├── issue-2/    ← branch: fabrik/issue-2
├── issue-5/    ← branch: fabrik/issue-5
└── issue-7/    ← branch: fabrik/issue-7
```

- Created on first processing of an issue
- Persist across polls for Claude session continuity
- `.fabrik/` is gitignored

## Test plan

- [ ] Build with `go build ./...`
- [ ] Run Fabrik, drag an issue to Research
- [ ] Verify `.fabrik/worktrees/issue-N/` is created with its own branch
- [ ] Verify Claude Code runs inside the worktree (check log output)
- [ ] Verify a second issue gets a separate worktree
- [ ] Verify stage names match board columns (Research, Plan, Implement, Review, Validate)


verveguy:
## Copilot Review Response

All 5 comments addressed in 29f062c. Here's the breakdown:

### 1. `os.Getwd()` error ignored in `Engine.New()` — **Fixed**
Good catch. Replaced with `git rev-parse --show-toplevel` which properly resolves the repo root even from subdirectories, and now returns an error if we're not in a git repo.

### 2. Stale worktree directory silently accepted — **Fixed**
Now validates that an existing directory is actually a git worktree on the expected branch (`git rev-parse --abbrev-ref HEAD`). If it's stale or invalid, removes and recreates it.

### 3. Branch creation from local ref may fail — **Fixed**
Now prefers `origin/<base>` when creating issue branches, falling back to the bare name if the remote ref doesn't exist. This handles the common case where only `origin/main` exists locally.

### 4. `--force` flag ordering in `git worktree remove` — **Fixed**
Moved `--force` before the path argument: `git worktree remove --force <path>`.

### 5. `DefaultBaseBranch()` fallback too naive — **Fixed**
Added fallback that checks for existence of `main`/`master` (both local and `origin/` variants) before defaulting to `"main"`. Didn't go as far as `git remote show origin` — that requires a network call on every startup which feels excessive for a fallback.

PR #8, opened 2026-04-01
Add .env config file to reduce cmd line toggles

### LINKED ISSUE #2: Add a config file to reduce cmd line toggles required
## Summary

Add a simple config file so we can put all the required config bits in one place. Make it a `.env` file to discourage it from getting checked in by mistake.

### Approach

- Use `joho/godotenv` (off-the-shelf, standard) to load a `.env` file from the current working directory (`./.env`)
- No `--config` flag; `./.env` in CWD is the only supported location
- On startup, if `.env` exists but is **not listed in `.gitignore`**, emit a fatal error (not just a warning) to prevent accidental check-in
- Preserve backward compatibility: `GITHUB_TOKEN` env var continues to work alongside the new canonical `FABRIK_TOKEN` key

### Design Decisions

| Decision | Choice | Rationale |
|---|---|---|
| Parser | `joho/godotenv` | Off-the-shelf, robust; confirmed preference over hand-rolled zero-dep parser |
| Config location | `./.env` in CWD | Keeps it simple; no `--config` flag needed |
| Missing `.gitignore` entry | **Fatal error** | Stronger than a warning — prevents accidental token leak; confirmed by user |
| `GITHUB_TOKEN` compat | Preserved | Existing workflows should not break; confirmed by user |

### Task Checklist

- [x] Add `joho/godotenv` dependency (`go get github.com/joho/godotenv`)
- [x] On startup, attempt to load `./.env` if it exists
- [x] Check that `.env` is listed in `.gitignore`; if not, exit with a clear fatal error message
- [x] Map `FABRIK_TOKEN` → GitHub token, with fallback to `GITHUB_TOKEN` for backward compatibility
- [x] Move all existing required env vars (`GITHUB_OWNER`, `GITHUB_REPO`, `GITHUB_PROJECT_NUMBER`, etc.) into `.env` documentation / example
- [x] Add a `.env.example` file showing all supported keys
- [x] Update `README.md` to document the `.env` setup and `.gitignore` requirement
- [x] Link the implementation PR to this issue

## Implementation PR

https://github.com/jmatthewpryor/fabrik/pull/8

verveguy:
## Summary
- Adds `.env` file support using `joho/godotenv` to load config from CWD
- All required settings (`FABRIK_TOKEN`, `FABRIK_OWNER`, `FABRIK_REPO`, `FABRIK_PROJECT_NUMBER`, `FABRIK_USER`) can be specified via env vars instead of flags
- Enforces `.gitignore` safety check — fatal error if `.env` exists but is not gitignored
- Preserves backward compatibility with `GITHUB_TOKEN` env var
- Includes `.env.example` and updated README documentation

## Test plan
- [ ] Verify `fabrik` starts with only a `.env` file (no flags)
- [ ] Verify flags override `.env` values
- [ ] Verify fatal error when `.env` exists but not in `.gitignore`
- [ ] Verify `GITHUB_TOKEN` still works as fallback
- [ ] Verify graceful behavior when no `.env` file exists

Closes #2

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,19 @@
+# Fabrik configuration
+# Copy this file to .env and fill in your values.
+# Make sure .env is listed in .gitignore (Fabrik will refuse to start otherwise).
+
+# Required: GitHub token with repo and project access
+FABRIK_TOKEN=ghp_your_token_here
+# GITHUB_TOKEN is also accepted for backward compatibility
+
+# Required: GitHub repository owner
+FABRIK_OWNER=your-org-or-username
+
+# Required: GitHub repository name
+FABRIK_REPO=your-repo
+
+# Required: GitHub project number
+FABRIK_PROJECT_NUMBER=1
+
+# Required: Your GitHub username (only process issues assigned to this user)
+FABRIK_USER=your-github-username
--- a/.env.example
+++ b/.env.example
@@ -17,3 +17,15 @@ FABRIK_PROJECT_NUMBER=1
 
 # Required: Your GitHub username (only process issues assigned to this user)
 FABRIK_USER=your-github-username
+
+# Optional: Directory containing stage YAML configs (default: ./stages)
+# FABRIK_STAGES=./stages
+
+# Optional: Auto-advance issues through stages without waiting for human input (default: false)
+# FABRIK_YOLO=false
+
+# Optional: Polling interval in seconds (default: 30)
+# FABRIK_POLL=30
+
+# Optional: Maximum concurrent issue workers (default: 5)
+# FABRIK_MAX_CONCURRENT=5
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .history/*
 .claude/
 .fabrik/
+.env
 stages/mystages/
 fabrik
 *.exe
--- a/README.md
+++ b/README.md
@@ -12,19 +12,19 @@ are user input, and the board columns define the workflow.
 # Build
 go build -o fabrik .
 
-# Set up your GitHub token
-export GITHUB_TOKEN=ghp_...
+# Create your .env file from the example
+cp .env.example .env
+# Edit .env with your GitHub token and repo details
+# Make sure .env is in your .gitignore (Fabrik enforces this)
 
 # Copy and customize stage configs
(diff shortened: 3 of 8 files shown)


verveguy:
🏭 **Fabrik — stage: Review**

Error: Reached max turns (30)

verveguy:
🏭 **Fabrik — stage: Review**

Error: Reached max turns (30)

verveguy:
🏭 **Fabrik — stage: Review**

Pushed successfully. Here's a summary of findings and actions:

---

**Findings:**
1. **`isInGitignore` false positives** (security bug): `strings.Contains` would match `.envrc` or `.env.example` in `.gitignore` as if they protected `.env`. Users running `direnv` commonly have `.envrc` gitignored — this would silently bypass the safety check, allowing `.env` with credentials to be loaded without being git-ignored.
2. **`FABRIK_YOLO` case sensitivity** (usability bug): Only `"true"`, `"1"`, `"yes"` matched; `"TRUE"`, `"True"`, `"YES"` silently evaluated to false.

**Actions taken:**
- Fixed `isInGitignore` to reject patterns where the filename appears embedded in a longer component (only allows end-of-pattern, `*`, or `/` after the match).
- Fixed `FABRIK_YOLO` to normalise the env value to lowercase before comparison.
- Added `TestIsInGitignore_SimilarName` covering the `.envrc`/`.env.example` false-positive cases.
- All 11 tests pass; changes pushed to `fabrik/issue-2`.

FABRIK_SUMMARY_BEGIN
Two bugs fixed on the issue-2 branch: (1) `isInGitignore` used `strings.Contains` which caused false positives — `.envrc` or `.env.example` in `.gitignore` would incorrectly bypass the security check that prevents `.env` from being loaded without git protection; the fix rejects matches where the filename is embedded inside a longer filename component. (2) `FABRIK_YOLO` only accepted lowercase values (`"true"`, `"yes"`), silently ignoring `"TRUE"` or `"True"`; fixed with a `strings.ToLower` normalisation. A new test `TestIsInGitignore_SimilarName` was added to cover the false-positive cases. All 11 tests pass; changes committed and pushed.
FABRIK_SUMMARY_END

FABRIK_STAGE_COMPLETE

PR #47, opened 2026-04-01
Add labels to indicate "editing" and "stage:Plan:in_progress" etc

### LINKED ISSUE #10: Add labels to indicate "editing" and "stage:Plan:in_progress" etc
Let's enrich the label usage to provide indication of when an issue or PR is being edited (active Fabrik task in progress to review comments, etc.) and when an Issue stage is underway - stage:Plan:in_progress and stage:Research:in_progress for example.

## Implementation

The following labels are now in use:

- **`fabrik:locked:<user>`** — Applied when a driver instance begins processing an issue. Prevents other driver instances from processing the same issue concurrently. Deferred-removed on all exit paths (including abnormal exits).
- **`stage:<Name>:in_progress`** — Applied when a stage begins executing for an issue. Deferred-removed on all exit paths. Only registered for cleanup when the label was successfully added, to avoid spurious removal warnings.
- **`stage:<Name>:complete`** — Applied when a stage finishes successfully.
- **`fabrik:editing`** — Causes Fabrik to skip an issue, signaling it is currently being manually edited.

## PR #47 — Copilot feedback resolved (commit cce35d1)

Two issues flagged during review were addressed:

1. **Stale lock label**: `fabrik:locked:<user>` is now deferred-removed on all exit paths, preventing permanent locks when a driver instance exits abnormally.
2. **Noisy deferred cleanup**: `removeInProgressLabel` (and `removeLockLabel`) are only deferred when the corresponding `AddLabelToIssue` call succeeds, eliminating spurious warnings. 404 responses from GitHub are treated as no-ops in both helpers.

verveguy:
Closes #10

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -349,6 +349,13 @@ func (e *Engine) processItem(board *gh.ProjectBoard, item gh.ProjectItem) error 
 		logf(item.Number, "warn", "could not add lock label: %v\n", err)
 	}
 
+	// Add in_progress label for this stage and ensure it's removed on all exit paths
+	inProgressLabel := fmt.Sprintf("stage:%s:in_progress", stage.Name)
+	if err := e.client.AddLabelToIssue(e.cfg.Owner, e.cfg.Repo, item.Number, inProgressLabel); err != nil {
+		logf(item.Number, "warn", "could not add in_progress label: %v\n", err)
+	}
+	defer e.removeInProgressLabel(item.Number, stage.Name)
+
 	// Ensure worktree exists for this issue
 	baseBranch := e.worktrees.DefaultBaseBranch()
 	workDir, err := e.worktrees.EnsureWorktree(item.Number, baseBranch)
@@ -582,6 +589,13 @@ func (e *Engine) postOutputToPR(item gh.ProjectItem, stageName, output string) {
 func (e *Engine) removeEditingLabel(issueNumber int) {
 	if err := e.client.RemoveLabelFromIssue(e.cfg.Owner, e.cfg.Repo, issueNumber, "fabrik:editing"); err != nil {
 		logf(issueNumber, "warn", "could not remove editing label: %v\n", err)
+	}
+}
+
+func (e *Engine) removeInProgressLabel(issueNumber int, stageName string) {
+	label := fmt.Sprintf("stage:%s:in_progress", stageName)
+	if err := e.client.RemoveLabelFromIssue(e.cfg.Owner, e.cfg.Repo, issueNumber, label); err != nil {
+		logf(issueNumber, "warn", "could not remove in_progress label: %v\n", err)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


verveguy:
🏭 **Fabrik — stage: Review**

The implementation is clean. Here's my review summary:

**Findings:**

1. **Branch state**: Clean, already rebased onto main (merge commit `0a128c5`).

2. **Implementation correctness**: The `stage:<name>:in_progress` label is added after lock acquisition at line 308, and removed via `defer` at line 312 — guaranteeing cleanup on all exit paths (EnsureWorktree failure, Claude invocation failure, normal completion).

3. **Early-return paths**: All returns before line 312 (lock check, cooldown, complete label, comment processing) occur before the defer is registered, so no label is leaked in those cases.

4. **Stale label resilience**: If the engine crashes with a stale `in_progress` label, the next run will re-add it (idempotent) and remove it via defer — no skip logic needed since `in_progress` is purely informational.

5. **Style consistency**: `removeInProgressLabel` uses `logf` (not `fmt.Printf`) and follows the same pattern as `removeEditingLabel` — consistent.

6. **Build and tests**: `go build ./...` and `go test ./...` both pass cleanly.

7. **No issues found** — the implementation is correct and complete.

FABRIK_SUMMARY_BEGIN
The implementation is clean and correct. `stage:<name>:in_progress` labels are added to issues after lock acquisition in `processItem()` and removed via `defer` to guarantee cleanup on all exit paths. No bugs, security issues, or style inconsistencies were found; build and tests pass.
FABRIK_SUMMARY_END

FABRIK_STAGE_COMPLETE

verveguy:
## Copilot feedback addressed

Two issues raised in the Copilot review have been fixed in commit cce35d1:

**1. Stale lock label (fabrik:locked:<user>)**
Previously the lock label was added but never removed, which could permanently block other driver instances from processing the same issue. Fixed by deferring `removeLockLabel` (analogous to the existing `removeInProgressLabel` pattern), but only when the `AddLabelToIssue` call succeeded.

**2. Noisy double-warning on failed add + deferred remove**
If `AddLabelToIssue` failed for the in-progress label, the existing unconditional `defer` would also fire on exit and log a second spurious warning about a label that was never added. Fixed by only registering the `defer` inside the `else` branch (i.e., when the add succeeded). Same guard applied to the new lock label cleanup.

**Additional hardening:** Both `removeLockLabel` and `removeInProgressLabel` now suppress 404 errors from the GitHub API, so a "label not present" response is treated as a no-op rather than a warning.

verveguy:

Validation complete. All tests pass, implementation meets all requirements.

FABRIK_SUMMARY_BEGIN
All tests pass and the implementation is correct. `stage:<name>:in_progress` labels are added after lock acquisition in `processItem()` and removed via `defer` on all exit paths; both lock and in-progress cleanup are guarded so defers are only registered on successful add, eliminating spurious warnings; 404 responses are treated as no-ops. The branch was rebased onto main and pushed.
FABRIK_SUMMARY_END

FABRIK_STAGE_COMPLETE

PR #172, opened 2026-04-06
Auto-upgrade from GitHub Releases for binary installations

### LINKED ISSUE #162: Auto-upgrade from GitHub Releases for binary installations
## Summary

The current `--auto-upgrade` mechanism only works when Fabrik is running from a git checkout of its own source — it does `git pull`, `go build`, and re-execs. When Fabrik is used to manage other projects (e.g., Liminis) it runs as a downloaded binary and cannot self-upgrade. This issue adds a release-based upgrade path: detect when running as a release binary, check the GitHub Releases API for a newer version, download the correct platform asset, atomically replace the running binary, and re-exec. The re-exec'd process refreshes embedded plugin skills before entering the poll loop.

## Requirements

- On idle polls (same `idleUpgradeThreshold` trigger as existing upgrade), perform the appropriate upgrade path based on how the binary was built:
  - **Git-based path** (existing, unchanged): version starts with `"dev"` (i.e., local build) AND Fabrik's own source repo is available as a worktree → `git pull → go build → run upgrade → re-exec`
  - **Release-based path** (new): version does not start with `"dev"` (injected via ldflags by GoReleaser) → GitHub Releases API check
- Note: `cmd.Version` for dev builds may be `"dev"` or `"dev(abc1234)"` (SHA appended by `init()`). Detection must use a prefix check, not equality. Release builds always have a tag value (e.g., `"v0.0.2"`).
- Version comparison: strip leading `v` from both the running version and the latest release tag, then compare as semver strings; only upgrade if `latestVersion > runningVersion`
- If the latest release version equals the running version, log nothing and continue
- If a newer release exists:
  - Identify the correct asset by matching `fabrik_VERSION_GOOS_GOARCH.tar.gz` against `runtime.GOOS` / `runtime.GOARCH`
  - If no matching asset is found, log a warning and continue (graceful fallback)
  - Download the asset to a temp file; extract the `fabrik` binary from the tarball
  - Verify the download is complete before replacing (atomic: download to temp → rename over existing)
  - Set env var `FABRIK_AUTO_UPGRADED=1` in the environment passed to `syscall.Exec`
  - Re-exec via `syscall.Exec` with original `os.Args` and modified `os.Environ()`
- The re-exec'd process detects `FABRIK_AUTO_UPGRADED=1` at startup, runs `fabrik upgrade` to refresh embedded plugin skills, then unsets the env var and continues into the poll loop
- If any step fails (API unreachable, asset missing, download error, rename error), log a warning with `logf(0, "upgrade", ...)` and continue the poll loop without crashing
- Rate-limit the GitHub Releases API check: skip the check if it was last performed within 30 minutes; track last-check time in memory (not persisted to disk)
- Use the existing GitHub token from config for authenticated API requests (avoids GitHub's unauthenticated rate limit)
- Multi-repo mode: if `primaryWorktrees()` returns nil, skip the release-based upgrade the same as the git-based upgrade is currently skipped (log a warning and return)

## Scope

**In scope:**
- Release-based auto-upgrade triggered by version-not-dev detection
- Semver comparison of running version vs. latest GitHub Release tag
- Platform-aware asset selection and tarball extraction
- Atomic binary replacement and re-exec
- Plugin skill refresh in the re-exec'd process via `FABRIK_AUTO_UPGRADED` env var detection
- Rate-limiting the release API check (in-memory, hardcoded 30 min)
- Graceful fallback on any failure

**Out of scope:**
- Making the rate-limit interval user-configurable via flag or config file (hardcoded default only)
- Checksum/signature verification of the downloaded asset
- Windows support (same constraint as the existing git-based path: `syscall.Exec` is unix-only)
- Package manager integration (homebrew, etc.)
- Downgrade support
- Pre-release/beta channel selection
- Expanding GoReleaser to build additional platforms (separate concern — current darwin/arm64-only config produces a graceful "no matching asset" warning on other platforms)
- Skill refresh on manual binary replacement or manual `fabrik upgrade` invocation (tracked separately in #168)

## Prior Art / Context

- **`cmd/version.go`**: `Version` defaults to `"dev"` for local builds; `init()` enriches it with the short VCS SHA producing e.g. `"dev(abc1234)"`. GoReleaser injects the release tag via `-X github.com/verveguy/fabrik/cmd.Version={{.Version}}`. Detection must check `strings.HasPrefix(Version, "dev")`, not `Version == "dev"`.
- **`engine/poll.go` `checkAndUpgrade()`**: Existing git-based flow runs `exec.Command(exe, "upgrade")` on the newly built binary *before* `syscall.Exec`. The release-based path differs: re-exec happens first with `FABRIK_AUTO_UPGRADED=1` in the environment, and the re-exec'd process handles skill refresh.
- **`.goreleaser.yaml`**: Archive naming is `fabrik_{{ .Version }}_{{ .Os }}_{{ .Arch }}` (`.tar.gz`). `{{ .Os }}` = `darwin`, `{{ .Arch }}` = `arm64` — these match `runtime.GOOS` / `runtime.GOARCH` exactly. Asset name for v0.0.2 would be `fabrik_v0.0.2_darwin_arm64.tar.gz`.
- **GitHub Releases REST API**: `GET /repos/{owner}/{repo}/releases/latest` returns JSON with `tag_name` and `assets[]` (each asset has `name`, `browser_download_url`, `size`). Auth header uses the existing config token.

## Risks / Dependencies

- **`syscall.Exec` is unix-only**: Same constraint as the existing path. No Windows support.
- **darwin/arm64 only in GoReleaser**: The feature silently does nothing on other platforms until GoReleaser is expanded. A logged warning makes this visible without crashing.
- **Binary in use on rename**: On Linux and macOS, `os.Rename` over an in-use executable is safe (new inode).
- **Token scope**: The GitHub PAT must have at least `public_repo` read access. The existing token already covers project/issue operations so scope is likely sufficient.
- **Dev build detection**: Version comparison must use `strings.HasPrefix(Version, "dev")`, not equality — the SHA-enriched form (`"dev(abc1234)"`) must also be treated as a dev build.
- **User-customized skills**: The `FABRIK_AUTO_UPGRADED` env var approach is intentional — skill refresh only runs when explicitly triggered by auto-upgrade, not on every launch, preserving any user customizations to `.fabrik/skills`. Manual upgrade skill refresh is tracked in #168.

verveguy:
Closes #162

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -293,12 +293,23 @@ func Execute() error {
 	}
 	fmt.Printf("  debug-output: %v\n", cfg.DebugOutput)
 
+	// If the process was re-exec'd after an auto-upgrade, refresh embedded plugin
+	// skills before entering the poll loop. The env var is unset immediately so
+	// the re-exec'd process doesn't loop.
+	if os.Getenv("FABRIK_AUTO_UPGRADED") == "1" {
+		os.Unsetenv("FABRIK_AUTO_UPGRADED")
+		if _, err := refreshPlugin(); err != nil {
+			fmt.Fprintf(os.Stderr, "[upgrade] warning: refreshPlugin failed: %v\n", err)
+		}
+	}
+
 	eng, err := engine.New(engine.Config{
 		Owner:         cfg.Owner,
 		Repo:          cfg.Repo,
 		ProjectNum:    cfg.ProjectNum,
 		User:          cfg.User,
 		Token:         cfg.Token,
+		Version:       Version,
 		Yolo:          cfg.Yolo,
 		AutoUpgrade:   cfg.AutoUpgrade,
 		PollSeconds:   cfg.PollSeconds,
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -20,6 +20,7 @@ type Config struct {
 	ProjectNum    int
 	User          string
 	Token         string
+	Version       string
 	Yolo          bool
 	AutoUpgrade   bool
 	PollSeconds   int
@@ -53,6 +54,7 @@ type Engine struct {
 	lastBlocked        map[string]bool       // key: issueKey; per-issue blocked-on-input from last processItem (for TUI)
 	lastUpdatedAt      map[string]time.Time  // key: issueKey; tracks last-seen updatedAt per issue
 	idleCount          int                   // consecutive idle polls; triggers self-upgrade at threshold
+	lastReleaseCheck   time.Time             // last time we checked GitHub Releases for a newer binary (in-memory rate limit)
 	sem                chan struct{}         // semaphore bounding concurrent workers across poll cycles
 	wg                 sync.WaitGroup        // tracks in-flight workers for graceful shutdown
 	inFlight           sync.Map              // key: issueKey string, value: bool (isPR)
--- a/engine/interfaces.go
+++ b/engine/interfaces.go
@@ -25,6 +25,7 @@ type GitHubClient interface {
 	CreateDraftPR(owner, repo, title, head, base string, issueNumber int) (int, error)
 	MarkPRReady(owner, repo string, prNumber int) error
 	MergePR(owner, repo string, prNumber int) error
+	FetchLatestRelease(owner, repo string) (*gh.LatestRelease, error)
 	RateLimitStats() (rest, graphql gh.RateLimitStats)
 }
 
--- a/engine/mocks_test.go
+++ b/engine/mocks_test.go
@@ -30,6 +30,7 @@ type mockGitHubClient struct {
 	getIssueBodyFn            func(owner, repo string, issueNumber int) (string, error)
 	markPRReadyFn             func(owner, repo string, prNumber int) error
 	createDraftPRFn           func(owner, repo, title, head, base string, issueNumber int) (int, error)
+	fetchLatestReleaseFn      func(owner, repo string) (*gh.LatestRelease, error)
 
 	// Track calls — access under mu when accessed from concurrent goroutines.
 	addLabelCalls      []addLabelCall
@@ -235,6 +236,16 @@ func (m *mockGitHubClient) MarkPRReady(owner, repo string, prNumber int) error {
 		return fn(owner, repo, prNumber)
 	}
 	return nil
+}
+
+func (m *mockGitHubClient) FetchLatestRelease(owner, repo string) (*gh.LatestRelease, error) {
+	m.mu.Lock()
+	fn := m.fetchLatestReleaseFn
+	m.mu.Unlock()
+	if fn != nil {
+		return fn(owner, repo)
+	}
+	return nil, nil
 }
 
 func (m *mockGitHubClient) RateLimitStats() (gh.RateLimitStats, gh.RateLimitStats) {
--- a/engine/poll.go
+++ b/engine/poll.go
@@ -1,12 +1,18 @@
 package engine
 
 import (
+	"archive/tar"
+	"compress/gzip"
 	"context"
 	"fmt"
+	"io"
+	"net/http"
 	"os"
 	"os/exec"
 	"os/signal"
 	"path/filepath"
+	"runtime"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
@@ -16,6 +22,15 @@ import (
 )
 
 const idleUpgradeThreshold = 2
+
+// fabrikOwner and fabrikRepo are the canonical owner/repo for fabrik itself.
+// These are used when checking the GitHub Releases API for a newer binary —
+// the release always targets verveguy/fabrik, not the user's managed project.
+const (
+	fabrikOwner           = "verveguy"
+	fabrikRepo            = "fabrik"
+	releaseCheckInterval  = 30 * time.Minute
+)
 
 // isTTY reports whether stdout is connected to a terminal.
 var isTTY = func() bool {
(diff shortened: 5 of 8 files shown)


verveguy:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-162 | commit: 160a1a96 | main: 2bdfd8fe86f7e010c0e5e4240bd051f355d5ce32 | 2026-04-06 01:48 UTC*

## Review Findings

### Fixed

- **Issue**: `TestCheckReleaseUpgrade_DownloadAttempted` was misleadingly named — it set up an HTTP server to "verify a download is attempted" but then used a `plan9_arm` asset that would never match, so the server was never hit. The function name and inline comments directly contradicted each other. **Fix**: Changed the asset name to use `runtime.GOOS`/`runtime.GOARCH` so it matches the current test platform. The HTTP server now receives the request (returns HTTP 500), and the test correctly asserts `downloaded == true`. This actually covers the "matching asset found → download attempted → graceful failure on server error" path.

- **Issue**: `extractBinaryFromTarball` — the most complex new function (gzip decompression, tar parsing, chmod, temp file lifecycle with error-path cleanup) had zero unit tests. **Fix**: Added `TestExtractBinaryFromTarball` (happy path: creates an in-memory tar.gz, verifies content and 0755 mode) and `TestExtractBinaryFromTarball_NotFound` (no `fabrik` entry in archive → error returned).

### Verified

- All tests pass (9 packages, race detector clean)
- `go vet ./...` clean — no static analysis issues
- `go build ./...` clean
- Rebased onto latest main (branch was already current)
- No external PR comments to address
- Spec compliance: version prefix detection (`strings.HasPrefix` not `==`), semver integer comparison handles `0.0.2` vs `0.0.10`, atomic rename with same-dir temp file, `FABRIK_AUTO_UPGRADED` unset before `engine.New()`, `primaryWorktrees()` nil guard, 30-minute in-memory rate limit, auth token on download request, graceful fallback on every error path

Reviewed release-based auto-upgrade implementation. Fixed 2 issues: a test named `TestCheckReleaseUpgrade_DownloadAttempted` that actually tested the opposite of what its name implied (corrected to use a matching platform asset and verify the download is attempted), and missing unit tests for `extractBinaryFromTarball` (added happy path and not-found error path). All tests pass with race detector clean.

---
Used 35/50 turns, 0k input / 14k output tokens.

verveguy:
*branch: fabrik/issue-162 | commit: 69a12683 | main: c1448c523025f029964f359a8a78f2c889eff38b | 2026-04-06 01:50 UTC*

Branch is updated. Here's the full validation report:

---

## Validation Report

### Requirements: 9/9 passed

- [x] **Version prefix detection**: `strings.HasPrefix(e.cfg.Version, "dev")` in `checkAndUpgrade()` (poll.go:471) — not equality, handles `"dev(abc1234)"` correctly
- [x] **Release path triggers on non-dev version**: `checkAndUpgrade()` branches to `checkReleaseUpgrade()` when version doesn't start with `"dev"`
- [x] **Semver comparison (integer-based)**: `semverGreater()` strips `v` prefix, compares segments as integers — handles `0.0.2` vs `0.0.10` edge case, verified by `TestSemverGreater` with 12 cases
- [x] **Equal version → log nothing**: `semverGreater` returns false for equal versions, `checkReleaseUpgrade` returns silently
- [x] **Platform-aware asset matching**: `fabrik_VERSION_GOOS_GOARCH.tar.gz` with `runtime.GOOS`/`runtime.GOARCH` (poll.go:706)
- [x] **No matching asset → warn and continue**: logs warning and returns (poll.go:711-714), verified by `TestCheckReleaseUpgrade_NoMatchingAsset`
- [x] **Atomic binary replacement**: `os.CreateTemp(filepath.Dir(exe), ...)` for same-filesystem rename, `defer` guards temp file cleanup, `os.Rename` over existing binary
- [x] **FABRIK_AUTO_UPGRADED=1 env var**: appended to `os.Environ()` before `syscall.Exec`; detected in `cmd/root.go:299-303` before `engine.New()`, calls `refreshPlugin()`, immediately unsets
- [x] **30-minute rate limit**: `releaseCheckInterval = 30 * time.Minute` constant, tracked in `Engine.lastReleaseCheck`, protected by mutex; verified with rate-limit and expiry tests
- [x] **Auth token used**: `Authorization: Bearer` header set on download request (poll.go:747)
- [x] **Multi-repo nil guard**: `primaryWorktrees()` nil check before API call (poll.go:668-671)
- [x] **All failures graceful**: every error path logs via `logf(0, "upgrade", ...)` and returns without crashing
- [x] **Version threaded through Config**: `engine.Config.Version` field added, set from `cmd.Version` in `cmd/root.go:312`

### Test Suite: PASSED

- 9 packages tested (1 with no test files: plugin, watch)
- Race detector: clean
- Build: clean
- Vet: clean
- New tests: `TestSemverGreater` (12 cases), `TestCheckReleaseUpgrade_UpToDate`, `TestCheckReleaseUpgrade_NoMatchingAsset`, `TestCheckReleaseUpgrade_RateLimitSuppressesSecondCall`, `TestCheckReleaseUpgrade_RateLimitExpiry`, `TestCheckReleaseUpgrade_DownloadAttempted`, `TestExtractBinaryFromTarball`, `TestExtractBinaryFromTarball_NotFound`

Existing tests all pass. The `GitHubClient` interface addition is backward-compatible (mock updated). `engine.Config.Version` is additive.

### Issues Found: None

### Branch State

- Rebased onto latest main (local had already been rebased; remote updated via force-push)

### Verdict: READY TO MERGE

Validation passed. All 9 requirements verified against the implementation: version prefix detection (`HasPrefix` not equality), integer semver comparison, platform-aware asset selection, atomic binary replacement, `FABRIK_AUTO_UPGRADED` env var detection before `engine.New()`, 30-minute in-memory rate limit, graceful fallback on all errors, and multi-repo nil guard. Full test suite passes with race detector clean (9 packages). Branch rebased onto latest main and pushed.

---
Used 30/50 turns, 0k input / 5k output tokens.

PR #184, opened 2026-04-06
Auto-upgrade skills on manual fabrik upgrade

### LINKED ISSUE #168: Auto-upgrade skills on manual fabrik upgrade
## Summary

When a user installs a new `fabrik` binary (via `fabrik upgrade` or manually), the embedded plugin skills may differ from what is on disk. Rather than a manifest-tracking approach (which would require the new binary to know hashes of all previous versions), the fix should compare the embedded skills in the running binary directly against the on-disk files at startup. If they differ, fabrik warns the user and offers to upgrade them interactively. This must not trigger on a fresh install or when no `fabrik init` has been run.

This issue covers the **manual upgrade path** only. The automated re-exec path (issue #162) via `FABRIK_AUTO_UPGRADED` is a separate concern.

## Requirements

- At startup, if `.fabrik/plugin/` exists (i.e., `fabrik init` has been run), compare the SHA256 of each embedded skill file against its on-disk counterpart.
- If any embedded file differs from the on-disk file, warn the user: "Plugin skills on disk don't match this version of fabrik. Do you want to upgrade them?"
- If the user responds yes, overwrite only the skill files that differ (not all plugin files) with the embedded versions.
- If the user responds no, proceed without changes and do not re-prompt in the same session.
- In non-interactive mode (no TTY), skip the prompt and print a warning to stderr: "Plugin skills on disk don't match this version of fabrik. Run `fabrik upgrade` to update them." Then proceed without changes.
- The check must NOT trigger when `.fabrik/plugin/` does not exist (no init done).
- The check must NOT trigger during `fabrik init` (the init command handles installation separately).
- The check must NOT trigger during `fabrik upgrade` (that command already performs an explicit overwrite).
- The check must NOT trigger during `fabrik watch` or `fabrik resume` (tactical utility subcommands, not daemon startup).

## Where the Check Runs

The startup check runs only in the main daemon path (`fabrik` with no subcommand), i.e., in `cmd/root.go`'s `Execute()` function, before the engine starts. It does NOT run in `fabrik init`, `fabrik upgrade`, `fabrik watch`, `fabrik resume`, or internal subcommands (`stream-filter`, `_debug-server`).

**`fabrik upgrade` retains its existing role** as an explicit, force-overwrite command for users who want to update plugin skills without starting the daemon.

## Scope

**In scope:**
- Startup detection in the main daemon path only: compare embedded skills vs. on-disk skills before starting the engine
- Interactive prompt when mismatch is detected and stdin is a TTY
- Non-interactive fallback: print a warning to stderr and continue
- Skip check when `.fabrik/plugin/` does not exist
- Overwrite only the differing files when user confirms upgrade

**Out of scope:**
- Manifest-based tracking of install-time hashes (rejected as too complex)
- Distinguishing "user-modified" from "version mismatch" — the prompt gives the user explicit control
- Stage YAML files in `.fabrik/stages/` — plugin skills only
- The auto-upgrade path (issue #162)
- Running the check in `fabrik watch`, `fabrik resume`, or any other subcommand

## Prior Art / Context

- `fabrik init` skips existing files unless `--force` is passed — the startup check complements this by catching the case where init was done with an older binary.
- `fabrik upgrade` currently blindly overwrites all plugin files via `refreshPlugin()` — this issue adds a startup check that only overwrites differing files when the user consents.
- `cmd/root.go` and `cmd/init.go` already import `github.com/mattn/go-isatty` for TTY detection — the non-interactive mode check follows the same pattern.
- Comparing embedded content directly against on-disk content (rather than tracking install-time hashes) is simpler and requires no persistent state.

## Risks / Dependencies

- If fabrik is run frequently (e.g., in a tight loop or CI), the startup check should be fast — SHA256 of a small set of text files is negligible.
- The check must be skipped in non-interactive environments (CI, cron, piped output) to avoid blocking automated pipelines — handled by the TTY check.
- Users who have intentionally customized plugin skill files will see the mismatch warning. The prompt gives them explicit control to keep or overwrite their customizations.

verveguy:
Closes #168

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -308,6 +308,13 @@ func Execute() error {
 		}
 	}
 
+	// Check whether on-disk plugin files match the embedded versions. This runs
+	// only in the main daemon path (all subcommands returned early above).
+	// If .fabrik/plugin/ does not exist (no fabrik init), the check is a no-op.
+	if err := checkPluginSkills(".fabrik/plugin"); err != nil {
+		fmt.Fprintf(os.Stderr, "[upgrade] warning: plugin skill check failed: %v\n", err)
+	}
+
 	eng, err := engine.New(engine.Config{
 		Owner:         cfg.Owner,
 		Repo:          cfg.Repo,
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -1,12 +1,16 @@
 package cmd
 
 import (
+	"bufio"
+	"crypto/sha256"
 	"fmt"
+	"io"
 	"io/fs"
 	"os"
 	"path/filepath"
 	"strings"
 
+	"github.com/mattn/go-isatty"
 	"github.com/verveguy/fabrik/config"
 	"github.com/verveguy/fabrik/engine"
 	gh "github.com/verveguy/fabrik/github"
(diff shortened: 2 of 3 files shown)


verveguy:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-168 | commit: 05d0321f | main: 11416aaec2f47b7bf78c824507c2960cf6b803e7 | 2026-04-06 03:15 UTC*

## Review Findings

### Fixed
No issues found — nothing to fix.

### Verified
- **All plan tasks complete**: All 5 tasks from the plan checklist are implemented exactly as specified.
- **Spec compliance**: All 8 requirements met — TTY prompt, non-TTY stderr warning, skip when `.fabrik/plugin/` absent, natural skip for init/upgrade/watch/resume/stream-filter via early-return dispatch.
- **Tests pass**: 6 test cases cover all branches (no-dir no-op, all-match no-op, non-TTY warning, TTY yes/no/N/empty).
- **`go test -race ./...`**: All packages pass, no races detected.
- **`go vet ./...`**: Clean.
- **`go build ./...`**: Clean.
- **Code follows project conventions**: Same patterns as `refreshPlugin()`, `isatty` pattern from `init.go`, error wrapping, `logf`-free (pure cmd layer).
- **Placement in `root.go`**: After `FABRIK_AUTO_UPGRADED` block — if auto-upgrade just ran, `checkPluginSkills` finds zero diffs and is a no-op. No interference between the two paths.

### Minor Note
The implementation compares all embedded plugin files (including `plugin.json` and `README.md`) rather than skills only. This was an explicit plan decision — these files version together, and `fabrik upgrade` already overwrites all of them. The prompt message "Plugin skills on disk don't match" is verbatim from the spec.

Reviewed startup plugin skill mismatch check implementation for issue #168. No issues found — all 5 plan tasks are complete, the implementation correctly skips for all subcommands via early-return dispatch, handles TTY/non-TTY paths, and all tests pass clean with the race detector.

---
Used 19/50 turns, 0k input / 5k output tokens.

verveguy:
*branch: fabrik/issue-168 | commit: 16e68772 | main: b609bf858a85c412eeac99b31c2c58114b2cd8b0 | 2026-04-06 03:18 UTC*

All checks pass. Here's the validation report:

---

## Validation Report

### Requirements: 8/8 passed

- [x] At startup, compare SHA256 of embedded plugin files against on-disk counterparts — implemented in `diffingPluginFiles()` (`cmd/upgrade.go`)
- [x] Warn "Plugin skills on disk don't match this version of fabrik. Do you want to upgrade them?" — `cmd/upgrade.go:78`
- [x] If yes, overwrite only differing files — `cmd/upgrade.go:86-99`
- [x] If no, proceed without changes — `cmd/upgrade.go:82-84`
- [x] Non-TTY: print warning to stderr "...Run `fabrik upgrade` to update them." and continue — `cmd/upgrade.go:73-76`
- [x] Check skipped when `.fabrik/plugin/` doesn't exist — `os.Stat` guard at `cmd/upgrade.go:61-63`
- [x] Check not triggered for `fabrik init`, `fabrik upgrade`, `fabrik watch`, `fabrik resume`, `stream-filter`, `_debug-server` — all return early in `Execute()` before `checkPluginSkills` is called (`cmd/root.go:45-64`)
- [x] Check placed after `FABRIK_AUTO_UPGRADED` block — `cmd/root.go:311-316`; if auto-upgrade ran first, `refreshPlugin()` brings files up to date and the check finds zero diffs

### Test Suite: PASSED

- 6 tests in `cmd/upgrade_test.go` covering: no-dir no-op, all-match no-op, non-TTY warning, TTY yes (overwrites), TTY no/N/empty (preserves)
- All packages pass with race detector: `ok cmd`, `ok engine`, `ok github`, `ok stages`, etc.
- `go build ./...`: clean
- `go vet ./...`: clean

Pre-existing tests unaffected. Branch rebased cleanly onto origin/main and force-pushed.

### Verdict: READY TO MERGE

---

All 8 requirements verified against the implementation. The `checkPluginSkills` function correctly gates on `.fabrik/plugin/` existence, handles TTY/non-TTY paths, and is naturally excluded from all subcommands via early-return dispatch. All 6 tests pass with the race detector; build and vet are clean. Branch rebased onto origin/main and pushed.

---
Used 24/50 turns, 1k input / 4k output tokens.

PR #290, opened 2026-04-10
Race condition in ensureBareClone when multiple workers discover a new repo simultaneously

### LINKED ISSUE #288: Race condition in ensureBareClone when multiple workers discover a new repo simultaneously
## Problem

When multiple issues from the same (new) repo appear on the board simultaneously, several workers can enter `ensureBareClone` concurrently for the same `owner/repo`. The first worker's `git clone --bare` succeeds, but subsequent workers also attempt to clone into the same directory, hitting:

```
fatal: cannot copy '.../templates/info/exclude' to '.../shadoworg-fabrik.git/info/exclude': File exists
```

This causes those issues to get `fabrik:paused` + `fabrik:awaiting-input` labels, requiring manual intervention to remove the labels and retry.

## Observed

Added 5 issues from `shadoworg/fabrik` to the Fabrik PM project board. Issue #2 cloned successfully, but #3–#6 all failed with the above error and were paused. The bare clone was healthy after #2 completed — the others just needed a retry.

## Root Cause

`ensureBareClone` checks `os.Stat(bareDir)` to see if the clone exists, then either fetches (exists) or clones (doesn't exist). But there's no mutex protecting this check-then-act sequence for the same repo. The worktree creation mutex (`worktreeMu`) serializes worktree operations but doesn't cover the initial bare clone.

The race window: two workers both see `bareDir` doesn't exist via `os.Stat`, both proceed to clone — the second clone attempt fails because the first has already partially or fully written the directory.

## Summary

Serialize `ensureBareClone` calls per repo so the check-then-clone sequence is atomic for a given `owner/repo`. Workers that arrive while a clone is already in progress should wait for it to complete and then use the successfully-cloned bare repo rather than failing.

## Requirements

- The check-then-clone sequence in `ensureBareClone` must be atomic per `owner/repo` — no two concurrent callers can race on the same repo path
- Workers that arrive while another worker is already cloning the same repo must wait for the first clone to complete, then proceed with the result
- If the first clone fails, waiting workers should also fail with the same error (rather than attempting their own clone into a partially-written directory)
- The existing `worktreeMu` on `WorktreeManager` protects per-worktree git config operations; the fix must not regress that serialization
- `ensureBareClone` is a package-level function (not a method on `WorktreeManager`) — the serialization mechanism must be accessible at the call site in `engine.go`'s `ensureRepoReady`
- The fix must handle the case where multiple repos are being cloned simultaneously — serialization should be per-repo, not global, to preserve concurrency across different repos

## Scope

**In scope:**
- Fixing the race condition in `ensureBareClone` for concurrent callers on the same repo
- Ensuring workers that lose the race wait for and reuse the winner's result

**Out of scope:**
- Retrying failed clones automatically (existing behavior: pause + `fabrik:awaiting-input`)
- Changing the bare-clone directory layout or naming convention
- Fixing any other concurrency issues unrelated to `ensureBareClone`

## Prior Art / Context

The existing `worktreeMu` in `WorktreeManager` uses a `sync.Mutex` to serialize per-worktree git operations. A similar approach per repo — using a `sync.Map` of per-repo mutexes or a singleflight group — would extend this pattern to bare clones.

The standard Go `golang.org/x/sync/singleflight` package is designed exactly for this pattern (deduplicate concurrent calls for the same key, share the result), but adding a new dependency should be weighed against a simple `sync.Map` of `sync.Mutex` values or channel-based approach using stdlib only. The project currently uses only `gopkg.in/yaml.v3` as an external dependency.

## Risks / Dependencies

- The fix introduces new shared state for coordinating bare-clone attempts; this state must itself be concurrency-safe
- Using a global map of per-repo mutexes (as opposed to singleflight) means the map is never cleaned up — acceptable given the number of repos is small and bounded by the project board

verveguy:
Closes #288

### DIFF
--- a/adrs/022-bare-clone-singleflight.md
+++ b/adrs/022-bare-clone-singleflight.md
@@ -1,0 +1,76 @@
+# ADR 022: Per-Repo Singleflight Coordination for Bare Clones
+
+## Status
+
+Accepted
+
+## Context
+
+When multiple issues from the same new repository appear on the project board simultaneously, concurrent workers all enter `ensureRepoReady` for the same `owner/repo`. The function checks whether a `WorktreeManager` is registered for that repo, finds none, and proceeds to call `ensureBareClone`. With N workers, N concurrent `git clone --bare` calls target the same destination directory.
+
+The first clone succeeds; subsequent clones fail with:
+
+```
+fatal: cannot copy '.../templates/info/exclude' to '.../shadoworg-fabrik.git/info/exclude': File exists
+```
+
+This causes those issues to receive `fabrik:paused` + `fabrik:awaiting-input` labels, requiring manual intervention to retry.
+
+The existing `worktreeMu` on `WorktreeManager` serializes per-worktree git operations, but a `WorktreeManager` does not exist until after the bare clone completes — so that mutex cannot protect the clone itself. Holding `e.mu` (the engine's general mutex) for the duration of a network-bound `git clone` would serialize all engine operations on all repos, which is unacceptable.
+
+## Decision
+
+Implement a **singleflight-style pattern using `sync.Map` + channels** (stdlib only) to serialize `ensureBareClone` calls per repo within `ensureRepoReady`.
+
+A `cloneInFlight sync.Map` field on `Engine` stores `*cloneCall` entries keyed by `"owner/repo"`. Each `cloneCall` holds:
+- `done chan struct{}` — closed when the clone completes (success or failure)
+- `dir string` — the bare clone directory on success
+- `err error` — the clone error on failure
+
+### Protocol
+
+**Election**: `cloneInFlight.LoadOrStore(nameWithOwner, call)` elects exactly one goroutine as the owner. The first to store wins; all others receive the existing entry.
+
+**Owner behavior**:
+1. Call `ensureBareClone`
+2. Store result in `call.dir` / `call.err`
+3. On failure: `close(call.done)` → `cloneInFlight.Delete(nameWithOwner)` → post comment/labels → return `ErrSkipItem`
+4. On success: `registerWorktrees(...)` → `close(call.done)` → return nil
+
+**Waiter behavior**:
+1. Block on `<-existing.done`
+2. On failure: log a brief message, return `ErrSkipItem` (no duplicate GitHub API calls)
+3. On success: call `registerWorktrees(existing.dir, ...)` (idempotent) → return nil
+
+### Failure-path cleanup
+
+On clone failure the entry is deleted from `cloneInFlight` before returning, so future poll cycles (after the user removes `fabrik:paused`) can retry by electing a new owner. If the entry were left in place, the next poll cycle would find a closed channel with a non-nil error and immediately return `ErrSkipItem` forever.
+
+### Success-path no-cleanup
+
+On success the entry is left in place (closed channel, nil error). Future callers exit at the fast-path `registered` check in `ensureRepoReady` (the `WorktreeManager` is now registered) before ever consulting `cloneInFlight`. The stale entry is harmless and is never read after the first successful clone.
+
+### Ordering invariant on failure
+
+`close(call.done)` happens before `cloneInFlight.Delete(nameWithOwner)`. Waiters that unblock on `close(call.done)` read `call.err` (which is set before `close`) and return `ErrSkipItem`. A re-entrant goroutine from the *next* poll cycle calling `LoadOrStore` after the delete sees no existing entry and becomes the new owner — correct behavior for retry.
+
+## Alternatives Considered
+
+### Map of `sync.Mutex` (one per repo)
+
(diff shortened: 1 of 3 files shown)


verveguy:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-288 | commit: 120e1444 | main: 393f76f8ef3fa911a0f6e2732118393e3cfeb503 | 2026-04-10 02:12 UTC*

The implementation is complete and correct. Here's my assessment:

## Review Findings

### Verified

**Correctness** — The singleflight pattern is implemented correctly:
- `call.dir` and `call.err` are set before `close(call.done)`, satisfying Go's memory model happens-before requirement for waiters reading those fields after `<-existing.done`
- On failure: `close(done)` → `Delete` → post comment/labels → return `ErrSkipItem` — correct ordering ensures waiters unblock before the entry is deleted, and the entry is gone before a future poll cycle can retry
- On success: `registerWorktrees` → `close(done)` — waiters see `existing.err == nil` and call `registerWorktrees` (idempotent); the stale entry is never re-read because the fast-path `registered` check exits early
- Waiters don't post duplicate comments or labels — only the owner does
- `LoadOrStore` atomically elects exactly one owner per repo

**Testing** — The concurrent test uses a proper N-goroutine barrier (`WaitGroup.Done` + `WaitGroup.Wait`) to maximize interleaving, asserts all 8 goroutines return `ErrSkipItem`, and asserts exactly 1 `AddComment` call. Passes under `-race`.

**Build/test/vet** — All pass clean:
- `go build ./...` — clean
- `go test -race ./...` — all packages pass (77s for engine package)
- `go vet ./...` — clean

**All plan tasks completed**:
- [x] `cloneCall` type + `cloneInFlight sync.Map` added to Engine
- [x] `ensureRepoReady` reworked with singleflight coordination
- [x] Concurrent test added (`TestEnsureRepoReady_ConcurrentSameRepo_OnlyOneCloneAttempt`)
- [x] ADR 022 created

**No fixes needed.** The implementation is clean and ready for human review.

Reviewed the singleflight coordination fix for the bare-clone race condition. Implementation is correct: `call.dir`/`call.err` are written before `close(call.done)` satisfying Go's memory model, failure path deletes the `cloneInFlight` entry to allow future retries, and waiters never post duplicate GitHub API calls. All tests pass under `-race`, build and vet are clean, and all four plan tasks are complete.

---
Used 15/50 turns, 0k input / 4k output tokens.

verveguy:
*branch: fabrik/issue-288 | commit: 120e1444 | main: 393f76f8ef3fa911a0f6e2732118393e3cfeb503 | 2026-04-10 02:14 UTC*

All validation checks complete. Here's the full report:

## Validation Report

### Requirements: 6/6 passed

- [x] **Check-then-clone atomic per `owner/repo`**: Verified — `cloneInFlight.LoadOrStore` elects exactly one owner; the owner is the only goroutine that calls `ensureBareClone`. No two concurrent callers can race on the same repo path.
- [x] **Waiting workers share the result**: Verified — waiters block on `<-existing.done` and use `existing.dir` to call `registerWorktrees` on success.
- [x] **First-clone failure propagates to waiters**: Verified — `call.err` is set before `close(call.done)`; waiters read it and return `ErrSkipItem`. Only the owner posts the comment/labels.
- [x] **`worktreeMu` serialization not regressed**: Verified — changes are confined to `engine.go`; `WorktreeManager.mu` and its usage in `worktree.go` are untouched.
- [x] **Serialization accessible at `ensureRepoReady` call site**: Verified — `cloneInFlight` is a field on `Engine`, used directly in `ensureRepoReady`. `ensureBareClone` signature is unchanged.
- [x] **Per-repo, not global**: Verified — `cloneInFlight` is a `sync.Map` keyed by `"owner/repo"`; different repos proceed concurrently.

### Test Suite: PASSED

- All packages pass across `go test -race ./...`
- `engine` package: 78s (most time; real git operations)
- Race detector: clean
- Vet: clean (`go vet ./...`)
- New concurrent test `TestEnsureRepoReady_ConcurrentSameRepo_OnlyOneCloneAttempt`: 8 goroutines with barrier, asserts all return `ErrSkipItem` and exactly 1 `AddComment` call

- Only 3 files changed: `engine/engine.go`, `engine/engine_extra_test.go`, `adrs/022-bare-clone-singleflight.md`
- All pre-existing tests pass

### Code completeness

- No TODOs/FIXMEs left
- No debug logging
- All plan tasks completed and checked off
- Failure/success ordering invariant (`close` before `Delete`) correctly implemented

### Branch state

- Rebased onto latest main (already up to date)

### Verdict: READY TO MERGE

Validated the singleflight fix for the bare-clone race condition. All 6 requirements verified: `cloneInFlight.LoadOrStore` atomically elects one clone owner per repo, waiters share the result via a channel, failure propagates without duplicate GitHub API calls, and per-repo concurrency is preserved. Full test suite passes clean under the race detector including the new 8-goroutine concurrent test. Branch is rebased and ready to merge.

---
Used 16/50 turns, 0k input / 2k output tokens.

Issue #412, opened 2026-04-16
Add per-stage wall-clock timeout to prevent indefinite hangs on stuck Claude processes


verveguy:
## Problem

Fabrik waits indefinitely for the Claude CLI process to exit via `cmd.Wait()` in `runClaude`. There is no wall-clock timeout. If a child process spawned during a Claude session hangs (stuck test suite, dangling git lock, process waiting on stdin), the Claude CLI stays alive and Fabrik's worker goroutine blocks forever — consuming a semaphore slot and preventing other work.

Observed on verveguy/graphiti#26: the Review stage completed (FABRIK_STAGE_COMPLETE emitted, 65 turns, $2.90), but Claude ran background pytest invocations that hung. The TUI showed the job at 39+ minutes "In Progress" even though the stage was done. The only fix was manual process kill.

`max_turns` bounds the conversation length but NOT wall time. A single turn with a hung Bash background task can block indefinitely.

## Summary

Add two complementary timeout mechanisms to prevent indefinitely stuck Claude processes. First, an optional per-stage `max_wall_time` field caps total process runtime. Second, a global hardcoded inactivity timeout kills the process if no streamed output has been received in 15 consecutive minutes — indicating a stuck session regardless of elapsed wall time. In both cases, output collected up to the kill point is processed normally using markers detected from the live stream, not only from the final result JSON. Kill signals target the entire process group so hung background children are also reaped. This issue builds the proactive-kill layer on top of issue #429's passive-recovery baseline (`WaitDelay`).

## Requirements

- **R1.** New optional stage config field `max_wall_time` (duration string, e.g. `30m`, `1h`). When absent or zero, no wall-clock timeout is applied (backward compatible). There is no global default.
- **R2.** The `max_wall_time` timeout starts when the Claude process is spawned (after worktree setup, context file writing, etc.) — not from poll dispatch time.
- **R3.** On timeout (either `max_wall_time` or inactivity): send SIGTERM to the entire process group, wait a hardcoded 10-second grace period, then SIGKILL to the entire process group if the process hasn't exited. This reaps hung background children (dangling pytest, `tail -f`, polling loops, etc.) in addition to the Claude CLI itself. Log a clear message: `[#N warn] stage %q exceeded max_wall_time (%s) — killing Claude process` or `[#N warn] stage %q idle for 15m with no output — killing Claude process`, respectively.
[...]

Issue #285, opened 2026-04-09
Stage bookkeeping not finalized when agent continues working past FABRIK_STAGE_COMPLETE


totalslacker:
## Problem

When a stage agent emits `FABRIK_STAGE_COMPLETE` correctly but then continues doing additional work (e.g. extra verification test runs) before its session ends, the Fabrik engine fails to finalize the stage's bookkeeping. The completion marker is observed and present in output, but never acted upon:

- No `history.json` entry is written for the stage
- The `stage:<Name>:in_progress` label is not replaced with `stage:<Name>:complete`
- The `fabrik:locked:<user>` lock is not released
- No stage summary comment is posted

The issue ends up in a permanent in-progress / locked state requiring manual recovery. The work itself is real and committed — the agent did everything correctly from a code perspective.

**Root cause confirmed via code reading** (`engine/claude.go:405-410`):

```go
if runErr != nil {
    return text, false, usage, fmt.Errorf("claude exited with error: %w", runErr)
}
completed := stageCompleteRE.MatchString(text)
```

When Claude exits with a non-zero exit code, `runClaude` forces `completed=false` regardless of whether `FABRIK_STAGE_COMPLETE` is present in the output. When the agent continues past the marker and the session ends with an error (max_turns exceeded, CCSDK timeout, API error), the engine treats the whole run as incomplete. `handleStageComplete` is never called.

## Concrete Reproduction

**Repo:** `totalslacker/SafariUnfucker`
**Issue:** #84 — Crash on quit: backfill task outlives exit() and faults in MPSGraph
**Stage:** Implement
**Session log:** `bc092344-14da-4f80-94e8-5e5847b60469.jsonl` (441 events, 712KB)
**Fabrik version:** Apr 2026 build with auto-upgrade enabled

### Timeline

| Time (UTC) | Event |
|---|---|
| 20:24:02 | Session starts |
| ... | ~430 events: agent edits source files, writes tests, runs test suite, opens PR #85 |
| **21:01:35** | **Agent emits `FABRIK_STAGE_COMPLETE`** (correct format — own line, preceded by `FABRIK_SUMMARY_END`) |
| 21:03:08 | Agent emits: "The direct SPM test run also completed with exit code 0 — confirming all tests pass." |
| 21:07:02 | Agent emits final text; session ends with non-zero exit (likely max_turns or API error) |

The agent did **5 minutes 27 seconds of additional post-completion verification work** after emitting the marker, then the session ended with a non-zero exit code.

[...]

PR #436, opened 2026-04-23
Engine silently drops completed stage results — no error, no completion, perpetual polling

### LINKED ISSUE #429: Engine silently drops completed stage results — no error, no completion, perpetual polling
## Problem

The Fabrik engine intermittently fails to process a completed Claude session, leaving the issue stuck with `fabrik:locked` and `stage:<Name>:in_progress` labels while polling endlessly. No error is logged — the engine simply never acknowledges the completion.

This is distinct from #285 (non-zero exit masking the completion marker) and #412 (hung subprocess preventing `cmd.Wait()` from returning). In this case, the Claude process exits cleanly with `FABRIK_STAGE_COMPLETE`, zero exit code, and `terminal_reason: completed`.

## Evidence

Observed on three separate issues in `totalslacker/SafariUnfucker` within 24 hours:

### Issue #121 (Review stage)
- Claude invoked at `23:51:33Z`
- Session log shows clean completion: 24 turns, $1.25, `terminal_reason: completed`, `is_error: false`
- `FABRIK_STAGE_COMPLETE` present in output
- **No `[#121 claude] completed` entry in fabrik.log** — the engine never processed the result
- Engine continues polling `deep-fetching details for #121` indefinitely

### Issue #126 (Implement stage, retry after turn-limit)
- First attempt hit 50-turn limit (correctly handled as incomplete)
- Retry invoked at `23:30:08Z`
- Session log shows clean completion: 13 turns, $1.86, `terminal_reason: completed`, `is_error: false`, `duration_ms: 728388`
- `FABRIK_STAGE_COMPLETE` present in output
- **No completion logged** — engine polls indefinitely

### Issue #130 (similar pattern, earlier occurrence)

All three required manual label cleanup to unstick.

## Summary

The engine worker goroutine gets permanently stuck inside `runClaude` even after the Claude process exits, because `cmd.Wait()` blocks indefinitely. The absence of ANY log entry after `[#N claude] invoking` — including the `[#N stats]` line that only appears after `e.claude.Invoke()` returns — confirms the goroutine never leaves `cmd.Run()`.

**Root cause confirmed:** Claude uses the Bash tool with `run_in_background: true` to start test runs and polling loops. These spawn grandchild processes (notably `tail -f` from the Monitor tool) that inherit Fabrik's stdout pipe file descriptor. When the Claude CLI process exits, these grandchildren are reparented to PID 1 (launchd/init) but keep the pipe's write end open. Go's `exec.Cmd.Wait()` reads stdout until EOF, which never arrives while any process holds the pipe open — so `cmd.Wait()` blocks indefinitely. There is no timeout, so the goroutine stays stuck forever.

This is a variant of #412, distinguished by the fact that the *Claude process itself* has exited (evidenced by `duration_ms` in the session log and the `terminal_reason: completed` field); only the grandchild processes remain.

## Requirements

- **(R1)**: When the Claude process exits (regardless of grandchild subprocess behavior), `cmd.Wait()` MUST return within a bounded time so the engine can process the result.
- **(R2)**: If `cmd.Wait()` does not return within a configurable deadline after the Claude process exits, the engine MUST recover: use the output buffered so far, process the result (including any `FABRIK_STAGE_COMPLETE` marker), and release the lock and in_progress labels.
- **(R3)**: The recovery path MUST be logged with enough detail to diagnose future occurrences (e.g., "cmd.Wait() exceeded deadline — processing buffered output").
- **(R4)**: An issue stuck due to this bug MUST not require manual label cleanup to unstick; restarting the engine with the fix applied MUST be sufficient for recovery.
- **(R5)**: Any grandchild process cleanup MUST be scoped to subprocesses of the specific Claude invocation, not the entire Fabrik process group.
- **(R6)**: The fix must not break stages where Claude legitimately commits files or leaves meaningful state in the worktree (i.e., killing orphaned monitoring processes must not discard in-progress work).
- **(R7)**: Existing tests for `runClaude` must continue to pass; a regression test reproducing the stuck-pipe scenario (grandchild holds pipe open after parent exits) MUST be added.

## Scope

**In scope:**
- Preventing `cmd.Wait()` from blocking indefinitely when grandchild processes hold the stdout pipe open
- Bounded timeout on `cmd.Wait()` after the Claude process exits (with recovery from the buffered output)
- Cleaning up grandchild process trees spawned by the Claude invocation, scoped per-invocation
- A regression test covering the grandchild-holds-pipe scenario
- Logging improvements so future occurrences are diagnosable without OS-level debugging

**Out of scope:**
- Fixing #412 (hung *main* Claude process — this fix specifically targets grandchildren outliving the parent)
- Handling the case where Claude itself hangs (that's the `max_turns` + retry path)
- Changes to the poll loop's deep-fetch throttling
- Preventing Claude from using `run_in_background` or the Monitor tool (that's a prompt/stage-config concern, not an engine fix)

## Prior Art / Context

- **#412**: Hung subprocess preventing `cmd.Wait()` return — same pipe-EOF mechanism, but reportedly the *main* Claude process is alive. This issue is about grandchildren only.
- **#285**: Fixed — non-zero exit code masking the completion marker (different root cause).
- **Go docs**: `exec.Cmd.Wait()` "waits for the command to exit and waits for any copying to stdin or copying from stdout or stderr to complete." Any process inheriting the pipe write fd will block EOF and thus block `Wait()`.
- **Go 1.20 `exec.Cmd.WaitDelay`**: Configures a duration to wait after the process exits before forcibly closing the pipe and returning from `Wait()`. This is the purpose-built fix for this exact scenario. Research should verify the module's minimum Go version and evaluate this approach.
- **Confirmed trigger pattern**: Claude uses `run_in_background: true` on Bash tool calls for test runs and polling loops (e.g., `until ! pgrep -f "swift test"`). The Monitor tool spawns `tail -f` processes that inherit the stdout fd. When Claude exits, these tail processes are reparented to PID 1 but keep the pipe open indefinitely.

## Risks / Dependencies

- **Go version compatibility**: `exec.Cmd.WaitDelay` requires Go 1.20+. The module's minimum Go version must be verified before relying on it; an alternative (manual goroutine + timer approach) is needed if the minimum is lower.
- **Output completeness**: Any approach that reads output after a timeout must ensure the full NDJSON output — including the final `{"type":"result",...}` line — is present in the buffer before parsing. If the deadline fires mid-stream, partial output could cause a JSON parse failure rather than a clean recovery. The fix must handle this case and log it distinctly.
- **subprocess isolation**: The culprit processes (`tail -f`, `while/until` polling loops) are monitoring processes with no side effects on the worktree. Killing them is safe. However, the fix must not accidentally kill user-space processes outside the Claude invocation's subprocess tree.

## Related

- #285 — Fixed: non-zero exit code masking completion marker (different root cause)
- #412 — Open: hung subprocess preventing `cmd.Wait()` return (main process still alive — different root cause)
- #419 — `update_issue_body` config migration gap (different subsystem but also a silent failure)

verveguy:
## Summary

The engine worker goroutine gets permanently stuck inside `runClaude` even after the Claude process exits, because `cmd.Wait()` blocks indefinitely. The absence of ANY log entry after `[#N claude] invoking` — including the `[#N stats]` line that only appears after `e.claude.Invoke()` returns — confirms the goroutine never leaves `cmd.Run()`.

**Root cause confirmed:** Claude uses the Bash tool with `run_in_background: true` to start test runs and polling loops. These spawn grandchild processes (notably `tail -f` from the Monitor tool) that inherit Fabrik's stdout pipe file descriptor. When the Claude CLI process exits, these grandchildren are reparented to PID 1 (launchd/init) but keep the pipe's write end open. Go's `exec.Cmd.Wait()` reads stdout until EOF, which never arrives while any process holds the pipe open — so `cmd.Wait()` blocks indefinitely. There is no timeout, so the goroutine stays stuck forever.

This is a variant of #412, distinguished by the fact that the *Claude process itself* has exited (evidenced by `duration_ms` in the session log and the `terminal_reason: completed` field); only the grandchild processes remain.

## Problem

The Fabrik engine intermittently fails to process a completed Claude session, leaving the issue stuck with `fabrik:locked` and `stage:<Name>:in_progress` labels while polling endlessly. No error is logged — the engine simply never acknowledges the completion.

This is distinct from #285 (non-zero exit masking the completion marker) and #412 (hung subprocess preventing `cmd.Wait()` from returning). In this case, the Claude process exits cleanly with `FABRIK_STAGE_COMPLETE`, zero exit code, and `terminal_reason: completed`.

### Approach

Use `exec.Cmd.WaitDelay` (Go 1.20+, available in Go 1.26.1) to bound how long `cmd.Wait()` blocks after the Claude process exits. When grandchild processes hold the stdout pipe open and the deadline fires, Go closes the pipe, returns `exec.ErrWaitDelay`, and the buffered output (complete, since Claude wrote it before exiting) is parsed normally.

Pair this with Unix process group isolation: start Claude in its own process group (`Setpgid: true`), then issue `SIGKILL` to the entire group after `cmd.Run()` returns — cleaning up `tail -f` and polling-loop grandchildren without affecting any other Fabrik-owned processes. The kill is scoped per-invocation (R5) and has no effect on worktree files (R6).

`ErrWaitDelay` is handled immediately after `cmd.Run()` — log the R3 diagnostic, clear `runErr`, and fall through to the existing parse/completion logic unchanged. The existing restart flow already handles R4 (stuck lock clears on engine restart once the goroutine unblocks).

WaitDelay is threaded through as a package-level variable (consistent with `claudePluginDir`, `claudeTUI`) set during engine construction, keeping `runClaude`'s signature unchanged.

The `SysProcAttr.Setpgid` field is Unix-only, so platform abstraction uses build-constrained files (`engine/procattr_unix.go` / `engine/procattr_windows.go`) — cleaner than a runtime `GOOS` check for syscall-level code.

### New/Modified Files

| File | Change |
|------|--------|
| `engine/engine.go` | Add `ClaudeWaitDelay time.Duration` to `Config`; set `claudeWaitDelay` pkg var in engine construction |
| `engine/claude.go` | Add `claudeWaitDelay` pkg var; in `runClaude`: set `cmd.WaitDelay`, call `setCmdProcAttr` / `killProcGroup`, detect and log `exec.ErrWaitDelay` |
| `engine/procattr_unix.go` | New — `//go:build !windows`; `setCmdProcAttr` sets `Setpgid: true`; `killProcGroup` sends `SIGKILL` to pgroup |
| `engine/procattr_windows.go` | New — `//go:build windows`; no-op stubs for both functions |
| `cmd/root.go` | Add `ClaudeWaitDelay int` (seconds) to `cmd.Config`; add `--claude-wait-delay` flag; add `FABRIK_CLAUDE_WAIT_DELAY` env var parsing; add `claudeWaitDelay()` converter; pass to `engine.Config` |
| `engine/invoke_claude_test.go` | Add `TestInvokeClaude_GrandchildHoldsPipe` regression test |
| `README.md` | Add `--claude-wait-delay` row to the CLI flags table |
| `docs/stage-lifecycle.md` | Add subprocess-cleanup note to the Claude invocation section |
| `adrs/005-claude-cli-invocation.md` | Update to document WaitDelay + process group additions |

### Key Decisions

- **`WaitDelay` not a manual goroutine/timer**: Go 1.20 added `WaitDelay` precisely for this scenario. No need for a goroutine-based alternative.
- **Package-level `claudeWaitDelay` var over adding a parameter to `runClaude`**: Consistent with `claudePluginDir`, `claudeTUI` pattern; avoids changing the signature of `InvokeClaude` and `InvokeClaudeForComments`.
- **`ErrWaitDelay` cleared before parse, not caught in the lower error block**: Simplest correct placement — `runErr = nil` after logging, then the existing parse and completion logic runs without modification.
- **`killProcGroup` called unconditionally after `cmd.Run()`**: Safe even on normal exit (SIGKILL on an already-dead/empty pgroup returns ESRCH, ignored). Ensures cleanup regardless of whether WaitDelay fired.
- **Build-constrained files over `runtime.GOOS` check**: Syscall types differ by platform at compile time; `Setpgid` doesn't exist on Windows. Build tags are the correct mechanism.
- **Seconds as unit for `FABRIK_CLAUDE_WAIT_DELAY`**: Default is 30 seconds — minute granularity would force a minimum of 60s, which is workable but unnecessarily coarse.
- **No new ADR**: This is an enhancement of the subprocess lifecycle already described in ADR 005. Update ADR 005 to document WaitDelay and process group management (including the first use of build-constrained files in `engine/`).

### Task Checklist

- [x] **Task 1**: Add `ClaudeWaitDelay time.Duration` to `engine.Config` in `engine/engine.go`; in the engine constructor (where `claudePluginDir` is set, ~line 112), add `claudeWaitDelay = cfg.ClaudeWaitDelay`
- [x] **Task 2**: Add `ClaudeWaitDelay int` to `cmd.Config`; add `--claude-wait-delay` flag (seconds, default 0); add `FABRIK_CLAUDE_WAIT_DELAY` env-var parsing block (same pattern as `FABRIK_CI_WAIT_TIMEOUT`); add `claudeWaitDelay(n int) time.Duration` helper (default 30s when n ≤ 0); pass `ClaudeWaitDelay: claudeWaitDelay(cfg.ClaudeWaitDelay)` in the `engine.New(...)` call in `cmd/root.go`
- [x] **Task 3**: Create `engine/procattr_unix.go` (`//go:build !windows`) with `setCmdProcAttr(cmd *exec.Cmd)` (sets `cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}`) and `killProcGroup(cmd *exec.Cmd)` (if `cmd.Process != nil`, calls `syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)` and ignores ESRCH)
- [x] **Task 4**: Create `engine/procattr_windows.go` (`//go:build windows`) with no-op stubs for `setCmdProcAttr` and `killProcGroup`
- [x] **Task 5**: In `engine/claude.go`: add `var claudeWaitDelay time.Duration` package-level var (with comment matching `claudePluginDir` style); in `runClaude`, add `cmd.WaitDelay = claudeWaitDelay` and `setCmdProcAttr(cmd)` between cmd construction and `cmd.Run()`; immediately after `cmd.Run()`, add `killProcGroup(cmd)` and then the `ErrWaitDelay` detection block: `if errors.Is(runErr, exec.ErrWaitDelay) { claudeLog(issueNumber, "warn", "WaitDelay fired: Claude exited but grandchild processes held stdout pipe open; processing buffered output (%d bytes)
", len(rawOutput)); runErr = nil }`
- [x] **Task 6**: Add regression test `TestInvokeClaude_GrandchildHoldsPipe` in `engine/invoke_claude_test.go`: fake `claude` script emits valid NDJSON with `FABRIK_STAGE_COMPLETE` then execs `sleep 60 &` (background process to hold pipe); set `claudeWaitDelay = 1 * time.Second`; defer restore to `0`; call `InvokeClaude`; assert it returns within 5 seconds with `completed=true` and no error. Mark test `//go:build !windows` or skip on Windows.
- [x] **Task 7**: Run `go build ./...` and `go test -race ./...` and fix any issues
- [x] **Task 8**: Update `README.md` — add `--claude-wait-delay` row to the CLI flags table (after `--max-retries`): document it as seconds, default 30, env var `FABRIK_CLAUDE_WAIT_DELAY`
- [x] **Task 9**: Update `docs/stage-lifecycle.md` — in the Phase 3 / Post-Invocation section, note that after `cmd.Run()` returns: (a) the process group is killed to clean up orphaned grandchild processes; (b) if `WaitDelay` fired, the buffered output is processed normally with a warning logged
- [x] **Task 10**: Update `adrs/005-claude-cli-invocation.md` — add a section documenting the WaitDelay bound (default 30s, `FABRIK_CLAUDE_WAIT_DELAY`), process group isolation via `Setpgid`, and the first use of build-constrained files in `engine/`

### Risks

- **Regression test flakiness on CI with restricted PID namespaces**: The `sleep 60 &` approach may not fork correctly in all CI environments. If this becomes an issue, replace with a small Go-binary helper written to a temp file. Mitigation: mark the test as skippable with `t.Skip` if the shell doesn't support backgrounding, or use `exec.Command("sleep", "60").Start()` directly in the test helper (which avoids shell entirely).

- **`cmd.Process` nil on `Start()` failure**: If `claude` binary is not found, `cmd.Run()` returns an error and `cmd.Process` may be nil. The `killProcGroup` implementation must guard: `if cmd.Process == nil { return }`.

- **WaitDelay and context cancellation ordering**: When the context is cancelled, `exec.CommandContext` kills Claude and `cmd.Run()` may return with both `ctx.Err() != nil` and the pipe still open. In this case the WaitDelay error may be wrapped alongside the context error. The existing `ctx.Err() != nil` check runs first in the error block and is unaffected — the `ErrWaitDelay` check fires before the error block (at the point of `runErr = nil`), so context-cancelled paths still return an error correctly.

---
Used 22/50 turns, 6k input / 11k output tokens.

## Verification

Implemented `cmd.WaitDelay` (30s default, configurable via `--claude-wait-delay` / `FABRIK_CLAUDE_WAIT_DELAY`) on every Claude subprocess to bound how long `cmd.Wait()` blocks after Claude exits — preventing worker goroutines from hanging indefinitely when grandchild processes (Monitor's `tail -f`, `run_in_background` loops) hold the stdout pipe open. Paired with Unix process-group isolation (`Setpgid: true` + `SIGKILL` to the group after `cmd.Run()` returns) via build-constrained platform files (`engine/procattr_unix.go` / `engine/procattr_windows.go`). When `ErrWaitDelay` fires, the buffered output is processed normally with a diagnostic warning logged. A regression test (`engine/grandchild_test.go`) confirms `InvokeClaude` returns in ~1s (not stuck) when a background `sleep 60` holds the pipe open — all 10 plan tasks completed, build and race-detector tests green.

---

Closes #429

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,6 +257,7 @@ GITHUB_TOKEN=ghp_...    # Fallback
 | `--notui` | Disable the interactive TUI dashboard | TUI on by default |
 | `--max-concurrent` | Maximum number of concurrent issue workers | `5` |
 | `--max-retries` | Max failed stage attempts before pausing the issue (0 = unlimited) | `3` |
+| `--claude-wait-delay` | Seconds to wait after Claude exits before recovering buffered output when grandchild processes hold stdout pipe open (env: `FABRIK_CLAUDE_WAIT_DELAY`). Prevents worker goroutines from blocking indefinitely when Claude uses `run_in_background` or the Monitor tool. | `30` |
 | `--review-wait-timeout` | Minutes to wait per reviewer-gate cycle before pausing. Explicitly passing `0` uses the built-in default of `15` and bypasses `FABRIK_REVIEW_WAIT_TIMEOUT`. When absent, `FABRIK_REVIEW_WAIT_TIMEOUT` is consulted first; falls back to `15` if unset. | `0` |
 | `--max-review-cycles` | Max re-invocation cycles per reviewer-gate session. Explicitly passing `0` uses the built-in default of `5` and bypasses `FABRIK_MAX_REVIEW_CYCLES`. When absent, `FABRIK_MAX_REVIEW_CYCLES` is consulted first; falls back to `5` if unset. | `0` |
 | `--debug-output` | Save Claude stage output to `.fabrik/debug/` for debugging | `false` |
--- a/adrs/005-claude-cli-invocation.md
+++ b/adrs/005-claude-cli-invocation.md
@@ -53,6 +53,62 @@ operates on the correct branch and file state.
 - **Claude Code dependency**: Requires `claude` CLI installed and on PATH.
   This is a hard requirement, not optional.
 
+## Subprocess Lifecycle Management (added with issue #429)
+
+### Problem
+
+Claude uses `run_in_background: true` on Bash tool calls for test runs and
+polling loops, and the Monitor tool spawns `tail -f` processes. These
+grandchild processes inherit Fabrik's stdout pipe write-fd. When the Claude
+CLI process exits, grandchildren are reparented to PID 1 but keep the pipe
+open. Go's `cmd.Wait()` blocks until the pipe's read side reaches EOF, which
+never arrives while any process holds the write side — so the worker goroutine
+blocks indefinitely.
+
+### Solution
+
+Two mechanisms, applied in `runClaude` after cmd construction:
+
+**1. `cmd.WaitDelay` (Go 1.20+)**
+
+```go
+cmd.WaitDelay = claudeWaitDelay  // default: 30s, configurable via --claude-wait-delay
+```
+
+After the Claude process exits, Go waits at most `WaitDelay` for I/O copy
+goroutines to finish. If the deadline fires, Go closes the pipe and returns
+`exec.ErrWaitDelay`. The engine detects this, logs a warning, and falls through
+to normal output processing — the buffered output (complete, since Claude wrote
+its final result line before exiting) is parsed normally.
+
+**2. Process Group Isolation + SIGKILL (Unix only)**
+
+```go
+// engine/procattr_unix.go (//go:build !windows)
+cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+// after cmd.Run() returns:
+syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)
+```
+
+Starting Claude with `Setpgid: true` creates a new process group with Claude
+as the process group leader (PGID == Claude's PID). After `cmd.Run()` returns
+(whether via normal exit or WaitDelay), SIGKILL is sent to the entire group,
+cleaning up all grandchildren. This is safe: SIGKILL on an empty/already-dead
+group returns ESRCH, which is ignored. It does not affect worktree file state
+since grandchild monitoring processes have no filesystem side effects.
+
+This is the first use of build-constrained files in `engine/`:
+`engine/procattr_unix.go` (`//go:build !windows`) and
+`engine/procattr_windows.go` (`//go:build windows`) provide Unix-specific
+syscall code while keeping the Windows build clean.
+
+### Configuration
+
+- `--claude-wait-delay <seconds>` (env: `FABRIK_CLAUDE_WAIT_DELAY`): how long
+  to wait after Claude exits before recovering buffered output. Default: 30s.
+  Increase if Claude's final output write is unusually slow; decrease in test
+  environments where fast recovery is preferred.
+
 ## Future Consideration
 
 If Anthropic releases a Go SDK or agent framework, we could switch to
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -43,6 +43,7 @@ type Config struct {
 	CIWaitTimeout     int // minutes; 0 means use default (30)
 	MaxCiFixCycles    int // 0 means use default (5)
 	MaxRebaseCycles   int // 0 means use default (3)
+	ClaudeWaitDelay   int // seconds; 0 means use default (30)
 	DebugOutput       bool
 	PluginDir         string
 }
@@ -115,6 +116,7 @@ func Execute() error {
 	flag.IntVar(&cfg.CIWaitTimeout, "ci-wait-timeout", 0, "Maximum time in minutes to wait for CI in the merge guard before pausing (0 = use default of 30; also FABRIK_CI_WAIT_TIMEOUT)")
 	flag.IntVar(&cfg.MaxCiFixCycles, "max-ci-fix-cycles", 0, "Maximum number of CI-fix cycles per issue before pausing (0 = use default of 5; also FABRIK_MAX_CI_FIX_CYCLES)")
 	flag.IntVar(&cfg.MaxRebaseCycles, "max-rebase-cycles", 0, "Maximum number of rebase-reinvoke cycles per issue before pausing (0 = use default of 3; also FABRIK_MAX_REBASE_CYCLES)")
+	flag.IntVar(&cfg.ClaudeWaitDelay, "claude-wait-delay", 0, "Seconds to wait after Claude exits before recovering buffered output when grandchildren hold stdout pipe open (0 = use default of 30; also FABRIK_CLAUDE_WAIT_DELAY)")
 	flag.BoolVar(&cfg.DebugOutput, "debug-output", false, "Save Claude stage output to .fabrik/debug/ for debugging")
 	flag.StringVar(&cfg.PluginDir, "plugin-dir", "", "Path to Fabrik plugin directory (for development; overrides installed plugin)")
 
@@ -307,6 +309,15 @@ func Execute() error {
 			}
 		}
 	}
+	if !explicitFlags["claude-wait-delay"] {
+		if v := os.Getenv("FABRIK_CLAUDE_WAIT_DELAY"); v != "" {
+			if n, err := strconv.Atoi(v); err == nil && n > 0 {
+				cfg.ClaudeWaitDelay = n
+			} else {
+				fmt.Fprintf(os.Stderr, "[warn] FABRIK_CLAUDE_WAIT_DELAY=%q is invalid (must be a positive integer of seconds); using default 30\n", v)
+			}
+		}
+	}
 	if !cfg.AutoUpgrade {
 		if v := os.Getenv("FABRIK_AUTO_UPGRADE"); v != "" {
 			lv := strings.ToLower(v)
@@ -428,6 +439,7 @@ func Execute() error {
 		CIWaitTimeout:     ciWaitTimeout(cfg.CIWaitTimeout),
 		MaxCiFixCycles:    maxCiFixCycles(cfg.MaxCiFixCycles),
 		MaxRebaseCycles:   maxRebaseCycles(cfg.MaxRebaseCycles),
+		ClaudeWaitDelay:   claudeWaitDelay(cfg.ClaudeWaitDelay),
 		DebugOutput:       cfg.DebugOutput,
 		PluginDir:         cfg.PluginDir,
 		Stages:            stageCfgs,
@@ -491,6 +503,15 @@ func maxRebaseCycles(n int) int {
 		return 3
 	}
 	return n
+}
+
+// claudeWaitDelay converts a ClaudeWaitDelay config value (seconds) to a
+// time.Duration. When seconds is 0 (unset), the default of 30 seconds is used.
+func claudeWaitDelay(seconds int) time.Duration {
+	if seconds <= 0 {
+		return 30 * time.Second
+	}
+	return time.Duration(seconds) * time.Second
 }
 
 // buildProjectInfo assembles the TUI footer metadata from the active config.
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -311,10 +311,15 @@ func Execute() error {
 	}
 	if !explicitFlags["claude-wait-delay"] {
 		if v := os.Getenv("FABRIK_CLAUDE_WAIT_DELAY"); v != "" {
-			if n, err := strconv.Atoi(v); err == nil && n > 0 {
-				cfg.ClaudeWaitDelay = n
+			if n, err := strconv.Atoi(v); err == nil {
+				if n > 0 {
+					cfg.ClaudeWaitDelay = n
+				} else if n < 0 {
+					fmt.Fprintf(os.Stderr, "[warn] FABRIK_CLAUDE_WAIT_DELAY=%q is invalid (must be 0 or a positive integer of seconds); using default 30\n", v)
+				}
+				// n == 0: silently use default (same semantics as --claude-wait-delay 0)
 			} else {
-				fmt.Fprintf(os.Stderr, "[warn] FABRIK_CLAUDE_WAIT_DELAY=%q is invalid (must be a positive integer of seconds); using default 30\n", v)
+				fmt.Fprintf(os.Stderr, "[warn] FABRIK_CLAUDE_WAIT_DELAY=%q is invalid (must be 0 or a positive integer of seconds); using default 30\n", v)
 			}
 		}
 	}
--- a/docs/stage-lifecycle.md
+++ b/docs/stage-lifecycle.md
@@ -168,6 +168,14 @@ Context files are available in .fabrik-context/
 
 Raw Claude output saved to `~/.fabrik/logs/issue-<N>/<stage>-output-<timestamp>.json` after every invocation. Viewable through the TUI's `l` key (piped through `fabrik _stream-filter` for human-readable display).
 
+### Subprocess Cleanup
+
+After `cmd.Run()` returns, two cleanup steps run unconditionally:
+
+1. **Process group kill**: Claude is started in its own process group (`Setpgid: true` on Unix). After `cmd.Run()` returns, `SIGKILL` is sent to the entire process group to clean up grandchild processes (e.g. `tail -f` from the Monitor tool, polling loops spawned via `run_in_background: true`). This is scoped to the specific Claude invocation and does not affect worktree file state — grandchild monitoring processes have no side effects on the filesystem.
+
+2. **WaitDelay bound**: `cmd.WaitDelay` is set to 30s (configurable via `--claude-wait-delay` / `FABRIK_CLAUDE_WAIT_DELAY`). When grandchild processes hold the stdout pipe open after Claude exits, Go's `cmd.Wait()` would otherwise block indefinitely. With `WaitDelay`, Go forcibly closes its end of the pipe after the deadline and returns `exec.ErrWaitDelay`. The engine detects this error, logs a diagnostic warning, clears the error, and processes the buffered output normally — including any `FABRIK_STAGE_COMPLETE` marker. This prevents the worker goroutine from being permanently stuck when Claude uses `run_in_background` or the Monitor tool.
+
 ---
 
 ## Phase 4: Post-Stage Handling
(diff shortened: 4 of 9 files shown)


verveguy:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-429 | commit: a52bea1e | main: 59e0ad7a4af737dd2e7f7762b8d259207df763dd | 2026-04-23 19:13 UTC*

## Review Findings

### Fixed

- **Issue**: `ErrWaitDelay` check cleared `runErr` unconditionally at `engine/claude.go:447`, even when the context was simultaneously cancelled (engine shutdown). When both WaitDelay fires AND the context is cancelled — e.g., Claude exits cleanly, grandchildren hold the pipe, then the engine shuts down — the existing shutdown guard at line 501 (`if ctx.Err() != nil`) was bypassed because `runErr` was already nil. This would cause the cancelled invocation to be treated as a successful stage completion, potentially triggering partial bookkeeping on shutdown.

  **Fix**: Added `ctx.Err() == nil` guard to the WaitDelay recovery block: `if errors.Is(runErr, exec.ErrWaitDelay) && ctx.Err() == nil`. Context-cancelled invocations now preserve `runErr` so the shutdown guard fires correctly. Committed as `a52bea1`.

### Verified

- **All tasks from the plan checklist completed**: `engine/procattr_unix.go` and `engine/procattr_windows.go` (build-constrained platform files) ✅; `claudeWaitDelay` pkg var and `cmd.WaitDelay` set in `runClaude` ✅; `killProcGroup` called unconditionally after `cmd.Run()` ✅; `cmd.Process == nil` guard in `killProcGroup` ✅; `engine.Config.ClaudeWaitDelay` wired through engine constructor ✅; `FABRIK_CLAUDE_WAIT_DELAY` env var parsing in `cmd/root.go` ✅; `claudeWaitDelay()` helper defaults to 30s when ≤ 0 ✅; `--claude-wait-delay` flag ✅; `engine/grandchild_test.go` regression test (`//go:build !windows`, WaitDelay=1s, 10s timeout, verifies `completed=true` when grandchild holds pipe) ✅; README, `docs/stage-lifecycle.md`, `adrs/005-claude-cli-invocation.md` updated ✅.

- **R4 (restart recovery)**: Confirmed handled by the existing `fabrik:locked` restart flow — no additional label cleanup code needed.
- **R5 (per-invocation scope)**: `Setpgid: true` gives Claude its own process group (PGID == Claude's PID); `SIGKILL` targets `-cmd.Process.Pid` which is that exact group and nothing else.
- **R6 (no worktree damage)**: `killProcGroup` kills monitoring processes (`tail -f`, polling loops) which have no filesystem side effects.
- **WaitDelay default fallback**: `runClaude` correctly defaults to 30s when `claudeWaitDelay == 0`, so production behavior is always bounded.

- **Tests**: Full suite green with race detector (148s, all packages). Engine package tests also green post-fix.
- **Build and vet**: Clean.
- **Rebased**: Branch was already current with `origin/main`.

### No Blocking Issues

The implementation is correct, complete, and well-tested. The single bug found and fixed was a context-cancellation edge case in the ErrWaitDelay recovery path.

---

Reviewed the WaitDelay / process-group fix for the stuck-worker bug. Found and fixed one correctness issue: the `ErrWaitDelay` recovery unconditionally cleared `runErr`, bypassing the engine-shutdown guard when both WaitDelay and context cancellation occurred simultaneously. Fixed with a `ctx.Err() == nil` guard. All other aspects of the implementation are correct — platform-constrained files, default fallback, killProcGroup nil-check, regression test, and documentation all verified. Tests pass race-clean.

---
Used 31/50 turns, 10k input / 13k output tokens.

verveguy:
*branch: fabrik/issue-429 | commit: 39d4ab13 | main: 59e0ad7a4af737dd2e7f7762b8d259207df763dd | 2026-04-23 19:18 UTC*

All three Copilot findings addressed, committed, and pushed. All PR review threads resolved.

## Applied Fixes

**Copilot (thread 1) — `engine/procattr_unix.go`**: `killProcGroup` was discarding all `syscall.Kill` errors. Fixed to only suppress `ESRCH` (process group already gone — expected), and log unexpected errors (e.g. `EPERM`) to stderr. Committed as `784194a`.

**Copilot (thread 2) — `engine/engine.go`**: `Engine.New` only assigned `claudeWaitDelay` when `> 0`, which meant a second engine constructed with `ClaudeWaitDelay=0` ("use default") would silently retain the previous engine's non-zero delay. Fixed to assign unconditionally. Committed as `b2a72f7`.

**Copilot (thread 3) — `cmd/root.go`**: `FABRIK_CLAUDE_WAIT_DELAY=0` was printing a noisy warning even though `0` is documented as "use default" (same as the flag's default). Fixed to accept `0` silently; warnings are now emitted only for negative values or non-integer input. Committed as `39d4ab1`.

All tests pass race-clean after the three fixes. PR review threads resolved.

---
**Threads addressed:**
- `engine/procattr_unix.go:25` — resolved
- `engine/engine.go:117` — resolved
- `cmd/root.go:317` — resolved

verveguy:
*branch: fabrik/issue-429 | commit: 39d4ab13 | main: 59e0ad7a4af737dd2e7f7762b8d259207df763dd | 2026-04-23 19:22 UTC*

All validation passes. Here's the full report:

## Validation Report

### Requirements: 7/7 passed

- [x] **R1** — `cmd.WaitDelay = waitDelay` set in `runClaude` (engine/claude.go:441) ensures `cmd.Wait()` returns within a bounded time regardless of grandchild behavior.
- [x] **R2** — `ErrWaitDelay` detection at claude.go:447 (`errors.Is(runErr, exec.ErrWaitDelay) && ctx.Err() == nil`) clears `runErr` and falls through to existing parse/completion logic, processing buffered output and advancing the stage.
- [x] **R3** — Warning logged: `"WaitDelay fired: Claude exited but grandchild processes held stdout pipe open; processing buffered output (%d bytes)"` (claude.go:448).
- [x] **R4** — Existing `fabrik:locked` restart flow handles recovery on engine restart (confirmed by research; no additional code needed).
- [x] **R5** — `Setpgid: true` (procattr_unix.go:15) starts Claude in its own process group; `SIGKILL` targets `-cmd.Process.Pid` — scoped exactly to that invocation's process tree.
- [x] **R6** — `killProcGroup` kills monitoring processes (`tail -f`, polling loops) which have no filesystem side effects; worktree changes are unaffected.
- [x] **R7** — `engine/grandchild_test.go` regression test (`TestInvokeClaude_GrandchildHoldsPipe`) passes with WaitDelay=1s and a 10s timeout; verifies `completed=true` when a grandchild holds the pipe.

### Test Suite: PASSED

- All 8 packages pass with race detector
- `engine` package: 147.7s (includes grandchild regression test with real sleep subprocess)
- Build: clean
- Vet: clean
- Race detector: no races detected

### Plan Checklist: 10/10 tasks completed

All tasks from the plan are implemented and verified in code.

### Documentation: Complete

- README.md: `--claude-wait-delay` row added to CLI flags table (Task 8) ✅
- docs/stage-lifecycle.md: Process group kill + WaitDelay recovery documented in Phase 3 (Task 9) ✅
- adrs/005-claude-cli-invocation.md: WaitDelay and Setpgid sections added (Task 10) ✅

All pre-existing tests pass unchanged. The new build-constrained platform files (`procattr_unix.go`, `procattr_windows.go`) compile correctly on macOS.

### Branch State: Clean

- Rebased: branch is current with `origin/main`
- All commits pushed to `origin/fabrik/issue-429`
- PR #436 is up to date

### Verdict: READY TO MERGE

Validation passed: all 7 requirements verified, full test suite green with race detector (148s, 8 packages), build and vet clean. The `WaitDelay` + process-group fix correctly bounds `cmd.Wait()` and recovers buffered output when grandchild processes hold the stdout pipe open, with a regression test confirming the fix. All 10 plan tasks complete; README, stage-lifecycle, and ADR docs updated.

---
Used 23/50 turns, 12k input / 3k output tokens.

PR #437, opened 2026-04-23
Add per-stage wall-clock timeout to prevent indefinite hangs on stuck Claude processes

### LINKED ISSUE #412: Add per-stage wall-clock timeout to prevent indefinite hangs on stuck Claude processes

verveguy:
## Summary

Add two complementary timeout mechanisms to prevent indefinitely stuck Claude processes. First, an optional per-stage `max_wall_time` field caps total process runtime. Second, a global hardcoded inactivity timeout kills the process if no streamed output has been received in 15 consecutive minutes — indicating a stuck session regardless of elapsed wall time. In both cases, output collected up to the kill point is processed normally using markers detected from the live stream, not only from the final result JSON. Kill signals target the entire process group so hung background children are also reaped. This issue builds the proactive-kill layer on top of issue #429's passive-recovery baseline (`WaitDelay`).

## Problem

Fabrik waits indefinitely for the Claude CLI process to exit via `cmd.Wait()` in `runClaude`. There is no wall-clock timeout. If a child process spawned during a Claude session hangs (stuck test suite, dangling git lock, process waiting on stdin), the Claude CLI stays alive and Fabrik's worker goroutine blocks forever — consuming a semaphore slot and preventing other work.

Observed on verveguy/graphiti#26: the Review stage completed (FABRIK_STAGE_COMPLETE emitted, 65 turns, $2.90), but Claude ran background pytest invocations that hung. The TUI showed the job at 39+ minutes "In Progress" even though the stage was done. The only fix was manual process kill.

`max_turns` bounds the conversation length but NOT wall time. A single turn with a hung Bash background task can block indefinitely.

### Approach

This issue requires three interlocking capabilities: (1) a per-stage `max_wall_time` YAML field with `context.WithTimeout` enforcement, (2) a hardcoded 15-minute inactivity watchdog goroutine, and (3) post-kill NDJSON scanning to detect `FABRIK_STAGE_COMPLETE` from intermediate `assistant` turns when the process is killed before emitting a final `result` line.

**Process kill sequence** (`cmd.Cancel` override): Override `cmd.Cancel` on the `exec.Cmd` to call a new `killProcGroupGraceful` (SIGTERM → 10s sleep → SIGKILL, targeting the process group). Go fires `cmd.Cancel` immediately when `stageCtx` is cancelled, so effective kill latency is ≤10s grace + WaitDelay for I/O drain. Works because `go.mod` requires Go 1.26.1 (well past the 1.20 minimum where `cmd.Cancel` was introduced).

**Inactivity tracking** (`activityWriter`): A minimal `io.Writer` wrapper that stores `time.Now().UnixNano()` atomically on every `Write`. It wraps the existing `stdoutWriter` chain. A watchdog goroutine resets a timer based on elapsed inactivity. When the timer fires, the watchdog calls `killProcGroupGraceful` directly — NOT via context cancellation — to avoid triggering the `ctx.Err() != nil` engine-shutdown guard. A `watchdogCtx` (cancelled in a `defer` after `cmd.Run()`) prevents goroutine leaks.

**R9 marker detection**: Do NOT parse NDJSON in the write path. All bytes streamed up to the kill are already in `stdout bytes.Buffer` (via `activityWriter` forwarding). When `parseClaudeJSON` fails AND `wasTimedOut` is true, call `extractTextFromAssistantTurns(rawOutput)` — a companion to the existing `extractIssueUpdateFromAssistantTurns` — to extract concatenated text from `assistant`-type NDJSON lines. The standard `stageCompleteRE.MatchString(text)` check then handles the marker.

**`wasTimedOut` flag** distinguishes our kills from engine shutdown: `inactivityFired atomic.Bool` (set by watchdog before firing) `||` `(stageCtx.Err() == context.DeadlineExceeded && ctx.Err() == nil)`. When `wasTimedOut`, the existing `if ctx.Err() != nil { return text, false, ... }` guard is gated out, allowing normal marker processing.

**`max_wall_time` field encoding**: Two exported fields on `Stage` — `MaxWallTimeRaw string yaml:"max_wall_time,omitempty"` (YAML-facing) and `MaxWallTime time.Duration yaml:"-"` (programmatic, set by `loadOne` via `time.ParseDuration`). Simpler than a custom `UnmarshalYAML` type and consistent with how `EffortLevel` is a validated string.

**Timeout context location**: `context.WithTimeout` is created inside `runClaude` (not `processItem`) so the clock starts at process spawn, satisfying R2. `stage.MaxWallTime` is threaded as a new `maxWallTime time.Duration` parameter on `runClaude`.

### New/Modified Files

| File | Change |
|------|--------|
| `stages/stages.go` | Add `MaxWallTimeRaw`/`MaxWallTime` fields; parse/validate in `loadOne` |
| `engine/procattr_unix.go` | Add `killProcGroupGraceful(pid, issueNumber int, label string)` |
| `engine/procattr_windows.go` | Add no-op `killProcGroupGraceful` stub |
| `engine/claude.go` | Add `activityWriter`, `extractTextFromAssistantTurns`; refactor `runClaude` with watchdog goroutine, timeout ctx, `cmd.Cancel` override, `wasTimedOut` logic; add `maxWallTime` param; update both `InvokeClaude` and `InvokeClaudeForComments` call sites |
| `engine/grandchild_test.go` | Add three timeout/inactivity tests using the fake-binary pattern |
| `docs/USER_GUIDE.md` | Add `max_wall_time` field entry + inactivity timeout note |
| `docs/state-machine.md` | Add kill-path guard conditions to `itemMayNeedWork` section |
| `adrs/029-proactive-kill-timeout.md` | New ADR (R11) — check highest existing number at implementation time |
| `adrs/005-claude-cli-invocation.md` | Add cross-reference to the new ADR |

### Key Decisions

- **`MaxWallTime` field encoding**: `MaxWallTimeRaw string yaml:"max_wall_time,omitempty"` + `MaxWallTime time.Duration yaml:"-"`, parsed in `loadOne`. Rejected: custom `UnmarshalYAML` type — more complex, no existing precedent in this file.

- **`cmd.Cancel` for wall-time kill**: Overriding `cmd.Cancel` (available since Go 1.20, we're on 1.26) keeps `runClaude` sequential. Go fires `cmd.Cancel` immediately on context cancellation. Rejected: `cmd.Start()` + manual `cmd.Wait()` goroutine — more complex restructuring for identical semantics.

- **Activity tracking via minimal writer**: `activityWriter` only updates an atomic timestamp; no NDJSON parsing in the write path. Rejected: full streaming NDJSON parser in the writer — unnecessary complexity since `rawOutput` is already fully buffered at kill time.

- **R9 via post-kill scan of `rawOutput`**: `extractTextFromAssistantTurns(rawOutput)` in the `!ok` + `wasTimedOut` branch. Rejected: live marker detection in the writer — equivalent result (all streamed bytes are in `rawOutput`) with more implementation complexity and partial-line accumulation hazards.

- **`wasTimedOut` boolean flag**: Set by `inactivityFired` atomic or by stageCtx DeadlineExceeded check. Rejected: new error types — unnecessary abstraction; the two sources are fully covered.

- **`context.WithTimeout` inside `runClaude`**: Most precise (R2) and requires only adding `maxWallTime` to `runClaude`'s signature. Both `InvokeClaude` and `InvokeClaudeForComments` pass `stage.MaxWallTime`.

- **ADR 029 warranted** (R11): The layered passive+proactive kill design is a non-obvious architectural constraint that future contributors need to know without reading the code.

### Task Checklist

- [ ] **Task 1 — Stage struct + loadOne** (`stages/stages.go`): Add `MaxWallTimeRaw string yaml:"max_wall_time,omitempty"` and `MaxWallTime time.Duration yaml:"-"` to `Stage`. In `loadOne`, after the `EffortLevel` validation block, if `s.MaxWallTimeRaw != ""`: call `time.ParseDuration(s.MaxWallTimeRaw)` and reject with an error if it fails or is negative; store the parsed value in `s.MaxWallTime`. Zero/absent = no timeout (backward compatible).

- [ ] **Task 2 — `killProcGroupGraceful`** (`engine/procattr_unix.go`, `engine/procattr_windows.go`): Add `killProcGroupGraceful(pid, issueNumber int, label string)` to the Unix file: `syscall.SIGTERM` to `-pid` (silent `ESRCH`), `time.Sleep(10 * time.Second)`, `syscall.SIGKILL` to `-pid` (silent `ESRCH`). Add no-op stub with same signature to the Windows file.

- [ ] **Task 3 — `activityWriter`** (`engine/claude.go`): Add `activityWriter` struct with `inner io.Writer` and `lastActivity *atomic.Int64`. Its `Write(p []byte) (int, error)` stores `time.Now().UnixNano()` atomically then delegates to `inner`. Add `var claudeInactivityTimeout = 15 * time.Minute` package-level var (settable in tests, same pattern as `claudeWaitDelay`).

- [ ] **Task 4 — `extractTextFromAssistantTurns`** (`engine/claude.go`): Add function that scans NDJSON lines in `rawOutput`, extracts and concatenates text content from all `assistant`-type messages (using the same envelope struct as `extractIssueUpdateFromAssistantTurns`). Returns the full concatenated text — `stageCompleteRE.MatchString` will then detect the marker if present.

- [ ] **Task 5 — `runClaude` setup refactor** (`engine/claude.go`): Add `maxWallTime time.Duration` as the last parameter. After log file setup: (a) create `stageCtx, stageCancel := context.WithTimeout(ctx, maxWallTime)` if `maxWallTime > 0`, else `stageCtx = ctx; stageCancel = func(){}`, always `defer stageCancel()`; (b) initialize `var lastActivity atomic.Int64; lastActivity.Store(time.Now().UnixNano())`; (c) wrap `stdoutWriter` in `activityWriter{inner: stdoutWriter, lastActivity: &lastActivity}` and assign as `cmd.Stdout`; (d) use `stageCtx` in `exec.CommandContext`; (e) after `setCmdProcAttr`, set `cmd.Cancel = func() error { if cmd.Process != nil { logf(issueNumber, "warn", "stage %q exceeded max_wall_time — killing Claude process\n", label); killProcGroupGraceful(cmd.Process.Pid, issueNumber, label) }; return nil }`.

- [ ] **Task 6 — Inactivity watchdog goroutine** (`engine/claude.go`): After configuring `cmd` (before `cmd.Run()`), declare `var inactivityFired atomic.Bool`, then start the watchdog: `watchdogCtx, watchdogCancel := context.WithCancel(context.Background()); go func() { defer watchdogCancel(); timer := time.NewTimer(claudeInactivityTimeout); defer timer.Stop(); for { select { case <-timer.C: since := time.Since(time.Unix(0, lastActivity.Load())); if since >= claudeInactivityTimeout { logf(issueNumber, "warn", "stage %q idle for 15m with no output — killing Claude process\n", label); inactivityFired.Store(true); killProcGroupGraceful(cmd.Process.Pid, issueNumber, label); return }; timer.Reset(claudeInactivityTimeout - since); case <-watchdogCtx.Done(): return } } }()`. After `cmd.Run()` returns, call `watchdogCancel()` immediately (in a `defer` placed before starting the goroutine).

- [ ] **Task 7 — Result processing with `wasTimedOut`** (`engine/claude.go`): After `watchdogCancel()` and `killProcGroup(cmd)`, compute `wasTimedOut := inactivityFired.Load() || (stageCtx.Err() == context.DeadlineExceeded && ctx.Err() == nil)`. In the `parseClaudeJSON` `!ok` branch: if `wasTimedOut`, set `text = extractTextFromAssistantTurns(rawOutput)` and log a warning; otherwise keep the existing error-message path. In the `runErr != nil` block: replace `if ctx.Err() != nil {` with `if ctx.Err() != nil && !wasTimedOut {` so killed-process paths still check for `stageCompleteRE` in `text`.

- [ ] **Task 8 — Thread `maxWallTime` through call chain** (`engine/claude.go`): In `InvokeClaude`, update the `runClaude(...)` call to append `stage.MaxWallTime`. In `InvokeClaudeForComments`, do the same. No changes to `processItem`, `processComments`, or interface signatures.

- [ ] **Task 9 — Tests** (`engine/grandchild_test.go`): Add three `//go:build !windows` tests following the fake-binary pattern:
  - `TestInvokeClaude_MaxWallTimeKillsWithComplete`: fake `claude` prints a valid `assistant` NDJSON line containing `FABRIK_STAGE_COMPLETE` then hangs (`sleep 60`); stage has `MaxWallTime = 500ms`; set `claudeWaitDelay = 1s`; assert `completed=true` and returns within 10s.
  - `TestInvokeClaude_MaxWallTimeKillsWithoutComplete`: fake `claude` hangs immediately (`sleep 60`); `MaxWallTime = 500ms`; assert `completed=false`, returns within 10s.
  - `TestInvokeClaude_InactivityTimeout`: set `claudeInactivityTimeout = 2s`; fake `claude` hangs immediately; assert returns within 15s with `completed=false`.

- [ ] **Task 10 — `docs/USER_GUIDE.md`**: Add `max_wall_time` entry to the Stage YAML Reference section (after `comment_max_turns`). Document: duration string format (e.g. `30m`, `1h`), default (absent = unlimited), suggested `45m` for Implement and Review stages. Add a separate paragraph documenting the hardcoded 15-minute inactivity timeout that applies to every invocation.

- [ ] **Task 11 — `docs/state-machine.md`**: Add two entries to the guard conditions section (invocation kill guards): `max_wall_time exceeded` — wall-clock deadline reached; SIGTERM→SIGKILL sequence terminates the invocation; routes to completion if STAGE_COMPLETE found in stream, otherwise to cooldown/retry. `Inactivity timeout (15m)` — no streamed output for 15 consecutive minutes; same kill and routing behavior.

- [ ] **Task 12 — ADR**: Check `adrs/` for the current highest number at implementation time; create `adrs/NNN-proactive-kill-timeout.md`. Document: the triggering incident (graphiti#26), the layered design (ADR 005 passive WaitDelay + this ADR proactive kill), the `max_wall_time` + 15m inactivity mechanisms, why R9 (stream marker detection) is critical for avoiding re-runs of already-completed stages, and the SIGTERM→10s→SIGKILL sequence reusing #429's process-group plumbing. Add a cross-reference sentence to `adrs/005-claude-cli-invocation.md`.

- [ ] **Task 13 — Build and test**: Run `go build ./...`, `go vet ./...`, and `go test -race ./...` to verify no regressions.

### Risks

1. **Inactivity goroutine leak if `cmd.Run()` panics**: Mitigated by placing `defer watchdogCancel()` before starting the goroutine (defers run even on panic).

2. **Race on `cmd.Process.Pid` in `cmd.Cancel`**: `cmd.Cancel` is only invoked after `cmd.Start()` sets `cmd.Process`, but guard with `if cmd.Process != nil` defensively.

3. **Simultaneous inactivity + wall-time fire**: Both call `killProcGroupGraceful`. The second call gets `ESRCH` (silently ignored). `wasTimedOut` is set by either path. Safe.

4. **`extractTextFromAssistantTurns` edge case — STAGE_COMPLETE split across turns**: If Claude emits the marker string split across two assistant messages, concatenation still assembles it correctly since the whole output is concatenated before the regex check.

5. **WaitDelay interaction with `cmd.Cancel`**: Go calls `cmd.Cancel` on context cancellation, then waits for the process to die. After the process exits, `WaitDelay` drains any remaining I/O (up to 30s). Maximum latency from `max_wall_time` expiry to `cmd.Run()` returning: ≤10s (SIGTERM grace) + 30s (WaitDelay) = ≤40s. Implementer should verify Go 1.26 exec sequencing matches this description (test `TestInvokeClaude_MaxWallTimeKillsWithComplete` will catch divergence).

6. **Fake binary test for inactivity**: The 2s `claudeInactivityTimeout` override must be restored with `defer` to avoid contaminating other tests. Use the same `origDelay` pattern as in `TestInvokeClaude_GrandchildHoldsPipe`.

---
Used 21/50 turns, 8k input / 23k output tokens.

## Verification

(Populated by Implement on completion)

---

Closes #412

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -111,6 +111,11 @@ mark_pr_ready_on_complete: true # Mark PR ready when stage completes
 auto_advance: false             # Override global yolo setting
 read_only: false                # Stash/restore worktree changes (for Specify/Research stages that don't write code)
 cleanup_worktree: false         # Remove worktree when stage completes (for Done/cleanup stages)
+max_wall_time: "45m"            # Optional: wall-clock deadline for a single Claude invocation (e.g. "30m", "1h").
+                                # When exceeded, SIGTERM → 10s → SIGKILL sent to the process group. Output
+                                # collected before the kill is scanned for FABRIK_STAGE_COMPLETE so completed
+                                # stages are not re-run. Absent or zero = no cap. A hardcoded 15-minute
+                                # inactivity timeout (no output received) applies to every invocation regardless.
 disable_adaptive_thinking: true # Disable Claude Code's adaptive (auto-reduced) thinking budget. Default: true.
 effort_level: max               # Claude Code thinking effort: low, medium, high, max. Default: high.
 ```
--- a/adrs/005-claude-cli-invocation.md
+++ b/adrs/005-claude-cli-invocation.md
@@ -109,6 +109,10 @@ syscall code while keeping the Windows build clean.
   Increase if Claude's final output write is unusually slow; decrease in test
   environments where fast recovery is preferred.
 
+### Proactive Kill for Stuck Processes
+
+`WaitDelay` only fires after Claude exits. It provides no bound when Claude itself hangs and never exits. [ADR 029](029-proactive-kill-timeout.md) adds two proactive kill mechanisms on top of this passive layer: a per-stage `max_wall_time` YAML field and a hardcoded 15-minute inactivity watchdog, both using the SIGTERM→10s→SIGKILL sequence via `killProcGroupGraceful`. After either kill, the already-buffered output is scanned for `FABRIK_STAGE_COMPLETE` in intermediate assistant NDJSON turns so completed work is not re-run.
+
 ## Future Consideration
 
 If Anthropic releases a Go SDK or agent framework, we could switch to
--- a/adrs/029-proactive-kill-timeout.md
+++ b/adrs/029-proactive-kill-timeout.md
@@ -1,0 +1,124 @@
+# ADR 029: Proactive Kill Timeouts for Stuck Claude Invocations
+
+## Status
+
+Accepted
+
+## Context
+
+Claude Code invocations can hang indefinitely. Two incident classes have been observed:
+
+1. **Grandchild pipe hold** (ADR 005, issue #429): A background process spawned by Claude (e.g., `tail -f` via Monitor, `sleep` in a test loop) inherits the stdout pipe write-fd and keeps it open after the Claude process exits. ADR 005's `WaitDelay` mechanism bounds this: after Claude exits, `cmd.Wait()` caps the pipe-drain delay at `claudeWaitDelay` (default 30s) and returns `exec.ErrWaitDelay`. The buffered output (complete, since Claude wrote its final result line before exiting) is then parsed normally.
+
+2. **Stuck Claude process** (graphiti#26 and class of related incidents): The Claude CLI process itself hangs — never exiting, producing no output for an extended period. Causes include: model-side infinite loop, stuck network request, deadlocked tool invocation, or a stage whose prompt creates a runaway session. `WaitDelay` is ineffective here because it only fires *after* Claude exits; it provides no bound on a process that never exits.
+
+These are distinct failure modes. ADR 005 addresses the grandchild pipe problem passively (drain-and-discard). This ADR addresses the stuck-process problem proactively (kill with recovery).
+
+## Decision
+
+Add two proactive kill mechanisms to `runClaude()` in `engine/claude.go`:
+
+1. **`max_wall_time` per-stage field**: a Go duration string in stage YAML that caps the total wall-clock runtime of a single Claude invocation via `context.WithTimeout`.
+
+2. **Inactivity timeout (hardcoded 15 minutes)**: a watchdog goroutine that kills the process if no stdout bytes have been received for 15 consecutive minutes, regardless of whether a `max_wall_time` is configured.
+
+Both use the SIGTERM→10s grace→SIGKILL sequence targeting the entire process group.
+
+After either kill, the already-buffered output is scanned for `FABRIK_STAGE_COMPLETE` in intermediate `{"type":"assistant"}` NDJSON lines so completed work is not re-run.
+
+## Rationale
+
+### Why two mechanisms?
+
+`max_wall_time` is opt-in per stage and requires the operator to reason about expected stage durations. It catches the case where a stage runs longer than ever expected. The inactivity timeout is a no-configuration backstop: even a stage with no `max_wall_time` is bounded against processes that produce occasional output but never complete a turn.
+
+### Why proactive kill instead of passive drain?
+
+`WaitDelay` (ADR 005) is passive — it only fires after the Claude process exits. A process that never exits requires a different approach: the engine must actively decide to end the invocation. The `context.WithTimeout` approach reuses Go's existing cancellation machinery; the inactivity watchdog handles the case where the process is still writing (slowly) but will never terminate.
+
+### Why SIGTERM→10s→SIGKILL (not immediate SIGKILL)?
+
+Claude Code may have staged partial file edits, open worktree file handles, or in-progress git operations. SIGTERM gives it a 10-second window to flush and clean up. If it does not exit within 10 seconds, SIGKILL is sent unconditionally. This is the same sequence used by most process supervision systems and is a well-understood contract.
+
+### Why target the process group?
+
+Claude spawns grandchild processes via `run_in_background: true` Bash tool calls (test runners, polling loops) and Monitor tool invocations (`tail -f`, `watch`). Sending the kill signal to the negative PID (`syscall.Kill(-pid, sig)`) targets the entire process group — Claude plus all descendants. This prevents grandchild processes from holding the stdout pipe open after the kill, which would otherwise cause `cmd.Wait()` to block until `WaitDelay` fires (adding 30s to every timeout recovery).
+
+### Why scan the buffer post-kill? (R9: stream marker detection)
+
+A stage that signals completion — then hangs in a post-completion loop — should not be re-run. The Claude CLI emits NDJSON per turn: each assistant turn is a `{"type":"assistant",...}` line with a `content` array. If Claude wrote `FABRIK_STAGE_COMPLETE` in a completed turn before hanging, that text is already in the buffer. `extractTextFromAssistantTurns()` scans the buffer for assistant turns and concatenates their text blocks; if `FABRIK_STAGE_COMPLETE` appears, `completed=true` is returned and the same completion flow runs as for a live marker. Without this scan, a timed-out stage that actually completed would be re-invoked — wasting time and potentially duplicating side effects.
+
+### Why not re-run the stage immediately on timeout?
+
+A timed-out invocation without `FABRIK_STAGE_COMPLETE` follows the same cooldown/retry path as any no-marker run. This gives the operator time to investigate (check the stage comment for partial output, inspect the worktree) before the engine retries. The `wasTimedOut` flag distinguishes our kills from engine-shutdown context cancellation so that only timeout kills enter the retry path, not graceful shutdowns.
+
+### Why hardcode 15 minutes for the inactivity timeout?
+
+15 minutes is chosen to exceed any expected per-turn latency (including reasoning + tool calls + model response) while bounding the worst-case hang to under an hour. It is configurable via the `claudeInactivityTimeout` package variable for tests. If experience shows 15 minutes is too tight, it can be made configurable via an env var; for now, the `max_wall_time` field provides per-stage control for operators who need tighter bounds.
+
+### Why is `cmd.Cancel` used for `max_wall_time` (not direct kill)?
+
+Go 1.20+ allows overriding `cmd.Cancel` — the function called when the context is cancelled. The default `cmd.Cancel` sends SIGKILL to the process (not the process group). Overriding it lets us run the SIGTERM→10s→SIGKILL graceful sequence while still using `context.WithTimeout` for deadline management. This keeps the kill logic in one place (`killProcGroupGraceful`) shared between both mechanisms.
+
+## How It Works
+
+```
+runClaude()
+│
+├── stageCtx = context.WithTimeout(ctx, maxWallTime)   // if maxWallTime > 0
+│
+├── cmd.Cancel = func() {                               // called on stageCtx deadline
+│     killProcGroupGraceful(pid, ...)
+│   }
+│
+├── cmd.Start()
+│   pid := cmd.Process.Pid
+│
+├── watchdog goroutine(pid):
+│   │  timer = 15min
+│   │  on timer fire:
+│   │    if idle >= 15min: inactivityFired=true; killProcGroupGraceful(pid,...)
+│   │    else: reset timer for remaining time
+│   └── exits on watchdogCtx.Done()
+│
+├── cmd.Wait()                         // blocks until process group done
+├── watchdogCancel()
+├── killProcGroup(cmd)                 // cleanup: SIGKILL any remaining grandchildren
+│
+├── wasTimedOut = inactivityFired || (stageCtx.Err()==DeadlineExceeded && ctx.Err()==nil)
+│
+└── output parsing:
+    ├── if parseClaudeJSON ok → normal path
+    ├── else if wasTimedOut:
+    │     text = extractTextFromAssistantTurns(rawOutput)
+    │     completed = strings.Contains(text, "FABRIK_STAGE_COMPLETE")
+    │     return text, completed, usage, nil   // not an error
+    └── else → error path (engine-shutdown or parse failure)
+```
+
+`extractTextFromAssistantTurns(rawOutput []byte)` reads `rawOutput` line by line, JSON-decodes lines with `"type":"assistant"`, and concatenates all `"text"` blocks from the message content array.
+
+## Stage Config
+
+```yaml
+max_wall_time: "45m"   # Optional. Absent or 0 = no wall-clock cap (inactivity backstop still applies).
+```
+
+Recommended values: `"45m"` for Implement and Review stages in typical repos. Research and Plan stages are generally faster but can be given `"20m"` as a conservative cap.
+
+## Trade-offs
+
+- **False positives on `max_wall_time`:** If a stage legitimately needs more time than configured, it will be killed and retried. Operators must tune `max_wall_time` to be conservatively large. The inactivity timeout does not have this problem for active (but slow) processes.
+
+- **10-second grace period adds latency:** Every `max_wall_time` kill takes at least 10 seconds past the deadline (the SIGTERM grace window). This is acceptable — hung processes are the exception, not the rule, and partial cleanup is preferable to corruption.
+
+- **Windows:** `killProcGroupGraceful` is a no-op on Windows (process groups work differently). Both mechanisms still fire and set their flags, but the kill is a best-effort `cmd.Cancel` (the default SIGKILL to the main process). Grandchild cleanup is weaker on Windows.
+
+- **Race on PID access:** `cmd.Process` is written by `cmd.Start()` and must not be read before `Start()` completes. The solution is to capture `pid := cmd.Process.Pid` in the main goroutine immediately after `cmd.Start()` succeeds, then pass `pid` as a parameter to the watchdog goroutine closure. This is race-free because the goroutine is launched after `cmd.Start()` completes and only reads the local `pid` copy.
+
+## References
+
+- [ADR 005: Shell Out to Claude Code CLI](005-claude-cli-invocation.md) — subprocess lifecycle; `WaitDelay` and `killProcGroup`; process group isolation
+- Issue #429 — grandchild pipe hold problem (led to ADR 005 subprocess lifecycle section)
+- Issue #412 — stuck Claude process problem (this ADR)
+- graphiti#26 — production incident that motivated this change
--- a/docs/USER_GUIDE.md
+++ b/docs/USER_GUIDE.md
@@ -458,6 +458,18 @@ max_turns: 50             # Optional. Max conversation turns per main stage invo
 comment_max_turns: 15     # Optional. Max turns when processing user comments. Defaults to
                           #   min(max_turns, 15) when max_turns is set, otherwise 15.
                           #   Keeps comment processing bounded independently of stage budget.
+max_wall_time: "45m"      # Optional. Wall-clock deadline for a single Claude invocation.
+                          #   Accepts Go duration strings (e.g. "30m", "1h", "1h30m").
+                          #   When the deadline is reached, Fabrik sends SIGTERM to the
+                          #   Claude process group, waits 10 s, then sends SIGKILL (reaping
+                          #   any hung background children). Output collected before the kill
+                          #   is processed normally — if FABRIK_STAGE_COMPLETE appears in
+                          #   the streamed output, the stage is marked complete without a
+                          #   retry. When absent or zero, no wall-clock cap is applied.
+                          #   Recommended: "45m" for Implement and Review stages in
+                          #   production use. The hardcoded 15-minute inactivity timeout
+                          #   (see below) acts as a universal backstop even when this field
+                          #   is not set.
 read_only: true           # Optional. Stashes the dirty worktree before Claude runs and
                           #   restores it after. Use for analysis stages that should not
                           #   modify files (e.g., Specify, Research).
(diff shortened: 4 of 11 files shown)


verveguy:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-412 | commit: 3ab0746d | main: f3ecc0f9923dfd779b809695db6aa943b45e39e4 | 2026-04-23 20:25 UTC*

## Review Findings

### Fixed

**[1] Misleading log message in `cmd.Cancel` when `maxWallTime == 0`**

When `maxWallTime == 0`, `stageCtx == ctx`. If the engine shuts down (context cancelled), `cmd.Cancel` fired and logged `"stage %q exceeded max_wall_time — killing Claude process"` — but no wall-time was configured. This would falsely imply a timeout during normal engine shutdown. Additionally, the spec (R3) calls for including the configured duration in the message.

**Fix** (`engine/claude.go`): `cmd.Cancel` now branches on `maxWallTime > 0`. When a wall-time is set it logs `"exceeded max_wall_time (%s)"` with the actual duration. When `maxWallTime == 0` (engine-shutdown path) it logs `"context cancelled — killing Claude process gracefully"`.

**[2] `max_wall_time` missing from CLAUDE.md stage config reference**

The new field was documented in `USER_GUIDE.md`, `docs/state-machine.md`, and the new ADR, but the Stage Config Options block in `CLAUDE.md` — which serves as the in-context reference for all Fabrik stage agents — was not updated.

**Fix** (`CLAUDE.md`): Added `max_wall_time` entry with description of the kill sequence, the FABRIK_STAGE_COMPLETE recovery behavior, and a mention of the hardcoded 15-minute inactivity backstop.

---

### Verified

**Correctness:**
- `wasTimedOut` logic correctly covers all four cases: wall-time kill, inactivity kill, both simultaneously, and engine shutdown (where it must remain `false`).
- When `wasTimedOut=true` and `FABRIK_STAGE_COMPLETE` is found in buffered assistant turns: `completed=true` propagates through `processItem` to `handleStageComplete` correctly even though `runErr != nil` (that error is logged as a warning at line 660 and processing continues).
- When `wasTimedOut=true` and no marker: `claudeRan=true`, `completed=false`, routes to cooldown/retry. Not a hard error.
- `extractTextFromAssistantTurns` correctly handles the NDJSON format and concatenates text across turns, assembling markers even if they happen to appear across turn boundaries.
- `killProcGroupGraceful` correctly silences `ESRCH` (process already gone) on both SIGTERM and SIGKILL calls.
- Inactivity watchdog timer correctly resets by measuring `since` after the timer fires, avoiding false positives from timer drift.
- Watchdog goroutine is started after `cmd.Start()` with a captured `pid` to avoid a data race on `cmd.Process`.

**Testing:**
- `TestInvokeClaude_MaxWallTimeKillsWithComplete` — wall-time kill, FABRIK_STAGE_COMPLETE in stream → `completed=true`.
- `TestInvokeClaude_MaxWallTimeKillsWithoutComplete` — wall-time kill, no marker → `completed=false`.
- `TestInvokeClaude_InactivityTimeout` — inactivity kill → `completed=false`.
- `TestLoadAll_MaxWallTime` — validates all parsing cases: valid minutes, valid compound, absent (zero), invalid format, negative value.
- All tests pass with `-race` flag, 10 packages, including the 170s engine package with the new 10-second grace period tests.

**Documentation:**
- `docs/USER_GUIDE.md`: `max_wall_time` entry and inactivity timeout section correctly placed and accurate.
- `docs/state-machine.md`: Two new state table rows (§7.6) and inline documentation of both kill mechanisms are accurate.
- `adrs/029-proactive-kill-timeout.md`: Complete, covers incident motivation, layered design rationale, mechanism diagram, trade-offs, and Windows caveat.
- `adrs/005-claude-cli-invocation.md`: Cross-reference to ADR 029 added.

**Security:** No new attack surfaces. Process group kill targets the invocation's own pgid (`Setpgid: true` from #429). No command injection or path traversal introduced.

**ADR numbering:** 029 is correct (028 is duplicated; 029 is the next sequential slot).

---

### Rebased

Branch was already current with `origin/main` — no rebase required.

---

Reviewed implementation of per-stage `max_wall_time` and 15-minute inactivity timeout mechanisms (issue #412). Fixed two issues: a misleading "exceeded max_wall_time" log message that fired during engine shutdown when no wall-time was configured (now correctly logs "context cancelled"), and a missing `max_wall_time` entry in the CLAUDE.md stage config reference block. All tests pass with the race detector across 10 packages including the new timeout and inactivity tests. PR is ready for human review.

---
Used 37/50 turns, 13k input / 16k output tokens.

verveguy:
*branch: fabrik/issue-412 | commit: 161654f1 | main: f3ecc0f9923dfd779b809695db6aa943b45e39e4 | 2026-04-23 20:30 UTC*

Done. Here's what was applied for each Copilot thread:

- **Thread 1** (`killProcGroupGraceful` ESRCH): Fixed — returns immediately on ESRCH from SIGTERM (group already gone, no PID-reuse hazard); also probes group with `Kill(-pid, 0)` before sending SIGKILL so a clean exit during the grace window skips the redundant signal.

- **Thread 2** (`cmd.Cancel` misleading log): Already fixed by the Review stage. Thread resolved.

- **Thread 3** (hardcoded "15m" in inactivity log): Fixed — now formats `claudeInactivityTimeout` into the message, so overrides in tests show the correct value.

All three review threads resolved via GraphQL. Tests pass (`ok engine 169s`, `ok stages cached`). Pushed.

---
**Threads addressed:**
- `engine/procattr_unix.go:41` — resolved
- `engine/claude.go:491` — resolved
- `engine/claude.go:520` — resolved

verveguy:
*branch: fabrik/issue-412 | commit: 161654f1 | main: f3ecc0f9923dfd779b809695db6aa943b45e39e4 | 2026-04-23 20:35 UTC*

All validation checks complete. Here is the full report:

---

## Validation Report

### Requirements: 11/11 passed

- [x] **R1** — `MaxWallTimeRaw string yaml:"max_wall_time,omitempty"` and `MaxWallTime time.Duration yaml:"-"` added to `Stage` struct in `stages/stages.go:119-125`. Absent or zero = no timeout (backward compatible). `loadOne` parses/validates the field at lines 202–211, rejecting negative values.

- [x] **R2** — `context.WithTimeout` is created inside `runClaude` at line 454, after log file setup and before `exec.CommandContext`, so the clock starts at process spawn time.

- [x] **R3** — `killProcGroupGraceful` in `engine/procattr_unix.go` implements SIGTERM → 10s sleep → probe-with-SIGKILL-only-if-group-still-alive. Log messages match spec: `"stage %q exceeded max_wall_time (%s) — killing Claude process"` and `"stage %q idle for %s with no output — killing Claude process"`. Windows no-op stub present.

- [x] **R4** — After kill, `wasTimedOut` flag routes to `extractTextFromAssistantTurns(rawOutput)`. If `FABRIK_STAGE_COMPLETE` found in streamed assistant turns, `completed=true` is returned (same path as live marker). Without the marker, routes to cooldown/retry (not a hard error).

- [x] **R5** — `context.WithTimeout` wraps the invocation inside `runClaude` at lines 453–456. Cancellation propagates to `exec.CommandContext`.

- [x] **R6** — Both `InvokeClaude` (line 244) and `InvokeClaudeForComments` (line 273) pass `stage.MaxWallTime` to `runClaude`. Comment processing gets its own fresh timeout per invocation.

- [x] **R7** — `docs/state-machine.md` updated: two state-table rows for the kill paths (§7.6 invocation-level kill paths), plus the `wasTimedOut` flag and guard conditions documented in the invocation lifecycle section.

- [x] **R8** — No default `max_wall_time`. `docs/USER_GUIDE.md` documents the field at line 461 with suggested `45m` for Implement/Review. Hardcoded 15-minute inactivity timeout also documented at line 551–557.

- [x] **R9** — `extractTextFromAssistantTurns` scans the already-buffered NDJSON for `{"type":"assistant"}` lines and concatenates their text content. If `FABRIK_STAGE_COMPLETE` appears in any assistant turn, `completed=true` is returned after a timeout kill.

- [x] **R10** — 15-minute inactivity watchdog goroutine started in `runClaude` (lines 512–530). Resets based on elapsed inactivity after timer fires. Stopped via `watchdogCtx.Done()` in a `defer`, preventing goroutine leaks. PID captured after `cmd.Start()` to avoid data race.

- [x] **R11** — `adrs/029-proactive-kill-timeout.md` created with full documentation: incident motivation (graphiti#26), layered design (ADR 005 passive WaitDelay + ADR 029 proactive kill), both mechanisms, R9 rationale, and SIGTERM→10s→SIGKILL explanation. ADR 005 cross-references ADR 029 at line 114.

### Test Suite: PASSED

- 10 packages tested
- `ok github.com/verveguy/fabrik/engine 169.550s` — includes 3 new timeout tests + 1 new stages parse test
- Race detector: clean
- Vet: clean (`go vet ./...`)

New tests verified:
- `TestInvokeClaude_MaxWallTimeKillsWithComplete` — wall-time kill with FABRIK_STAGE_COMPLETE → `completed=true`
- `TestInvokeClaude_MaxWallTimeKillsWithoutComplete` — wall-time kill without marker → `completed=false`
- `TestInvokeClaude_InactivityTimeout` — inactivity kill → `completed=false`
- `TestLoadAll_MaxWallTime` — YAML parse: valid durations, absent (zero), invalid format, negative rejection

All pre-existing tests pass. Interface signatures unchanged (`InvokeClaude`, `InvokeClaudeForComments` signatures are identical; `maxWallTime` was added only to the internal `runClaude`). The `wasTimedOut` flag correctly scopes the behavior change to timeout-kill paths only, leaving engine-shutdown cancellation behavior intact.

### Verdict: READY TO MERGE

Branch is clean, rebased onto main (already current), all commits pushed. Implementation fully satisfies all 11 requirements with thorough test coverage and updated documentation in USER_GUIDE, state-machine.md, CLAUDE.md, and a new ADR 029.

All 11 requirements verified and passing. The per-stage `max_wall_time` YAML field, 15-minute inactivity watchdog, SIGTERM→10s→SIGKILL graceful kill sequence, and post-kill `FABRIK_STAGE_COMPLETE` recovery from buffered assistant turns are all correctly implemented. Full test suite passes with the race detector (10 packages, including 4 new tests). Documentation updated in USER_GUIDE, state-machine.md, CLAUDE.md, and ADR 029 created. Branch is clean and up to date.

---
Used 36/50 turns, 15k input / 7k output tokens.

PR #18, opened 2026-04-01
Refactor to allow multiple concurrent issues and comments in parallel

verveguy:
## Summary
- Adds a configurable worker pool for concurrent issue and comment processing
- Implements label-based distributed locking (`fabrik:locked:{user}`) to prevent duplicate processing
- Uses a single poll worker that shares query results across goroutines
- Defaults to 5 concurrent workers, configurable via `FABRIK_MAX_CONCURRENT`

## Changes
- `cmd/root.go`: Introduces worker pool initialization and `FABRIK_MAX_CONCURRENT` config
- `engine/engine.go`: Refactors `poll()` to spawn concurrent issue/comment workers with mutex-protected `processedSet`
- `engine/engine_test.go`: Adds test coverage for concurrent processing scenarios

Closes #7

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #455, opened 2026-04-26
Webhook-driven event delivery to reduce GraphQL load and cut event latency

### LINKED ISSUE #452: Webhook-driven event delivery to reduce GraphQL load and cut event latency
## Problem

Fabrik currently polls the GitHub Project board on a fixed cadence (`PollSeconds`, default 30s) via a single GraphQL query that fetches the entire board, every active item's comments, and linked-PR data. Per-item REST calls (`FetchCheckRuns`, `FetchPRMergeable`, `FetchLabelAppliedAt`, etc.) layer on top of that. Two consequences:

1. **GraphQL budget pressure.** Heavy boards consume the 5000-points/hour GraphQL budget quickly. Users running Fabrik alongside other gh-CLI / GraphQL workloads on the same token hit rate-limit caps. The current idle-backoff caps the interval at 5 minutes, which limits the worst case but doesn't help during active periods.
2. **Event latency.** Comments, label changes, CI completions, and review submissions are visible to Fabrik only on the next poll tick — up to 30s of delay even when nothing else is happening. Comment-driven steering ("please look at PR #18") feels sluggish; CI gate clearance is bottlenecked on poll cadence rather than actual GitHub state.

GitHub publishes webhooks for every event Fabrik cares about. Subscribing to them collapses the steady-state polling load to near zero and makes events visible to the engine within seconds of GitHub recording them.

## Summary

Add an optional webhook-driven event ingestion path. When enabled, Fabrik receives event payloads from GitHub, translates them into existing trigger primitives (force-poll-this-issue, process-this-comment, re-evaluate-CI), and dramatically extends idle-backoff during periods when the webhook stream is healthy. Polling continues as a safety net so missed or delayed webhooks don't strand work.

The transport for webhook delivery is the official GitHub CLI command `gh webhook forward`, which uses GitHub's own SSE-based forwarding. No third-party service, no extra account, no public ingress required — Fabrik connects outbound through `gh` and receives events on `localhost`. Authentication piggybacks on the existing `gh auth login` the user already has.

### Transport and ingestion

- **(R1)** Spawn `gh webhook forward` as a managed subprocess when webhook mode is enabled. The subprocess streams events via SSE and POSTs them to a Fabrik-internal HTTP listener on `localhost:<port>` (configurable, default `127.0.0.1:0` so the OS picks a free port and Fabrik discovers it). The listener must bind only to `127.0.0.1` — never accept off-box traffic.
- **(R2)** Verify the HMAC-SHA256 signature on every received payload using the webhook secret. Fabrik generates a random 32-byte secret (from `crypto/rand`) at startup and passes it to `gh webhook forward` via `--secret`. Reject any payload that fails signature verification with a logged warning. Do not act on unverified events under any circumstance.
- **(R3)** Manage the `gh webhook forward` subprocess lifecycle: start on engine start, restart with exponential backoff on unexpected exit (cap at 60s), capture stderr to `.fabrik/fabrik.log` (the same log file as the engine), terminate cleanly on engine SIGINT/SIGTERM. The engine must never deadlock waiting on the subprocess; webhook plumbing failures are non-fatal.
- **(R4)** Track webhook stream health with three distinct states:
  - **`WebhookStreamStartingUp`**: the first `WebhookStartupGrace = 30 seconds` after subprocess start. During this state, the engine treats the stream as healthy for backoff purposes (R8 idle cap applies) but the TUI indicator shows a distinct color (e.g. blue/spinner) to distinguish it from confirmed-healthy. The first received event transitions immediately to `WebhookStreamHealthy` regardless of remaining grace time.
  - **`WebhookStreamHealthy`**: at least one event has been received in the last `WebhookHealthWindow` (default 10 minutes).
  - **`WebhookStreamUnhealthy`**: the health window has elapsed with no events (either after the startup grace expires with no first event, or during normal operation). When unhealthy, fall back to the existing poll cadence and idle-backoff schedule.
  - Log transitions between all three states. On subprocess restart, grace re-applies (`WebhookStreamStartingUp` again).

### Prerequisite check

- **(R5)** At startup, when `--webhooks` is enabled, check `gh --version` and verify it meets the minimum required version (`gh ≥ 2.32.0`, when `gh webhook forward` with `--secret`, `--repo`, and `--events` was introduced). On too-old or missing `gh`: log a clear error explaining the prerequisite and fall back to polling-only mode. Webhook unavailability is always non-fatal.

### Event handling

- **(R6)** Subscribe to these GitHub event types via `gh webhook forward --events`:
  - `issue_comment` (created, edited, deleted)
  - `issues` (opened, edited, labeled, unlabeled, closed, reopened, transferred)
  - `pull_request` (opened, edited, closed, reopened, ready_for_review, synchronize)
  - `pull_request_review` (submitted, edited)
  - `pull_request_review_comment` (created, edited)
  - `check_run` (completed)
  - `check_suite` (completed)
  - `projects_v2_item` (created, edited, archived) — board column changes. Subscribe to it; if `gh webhook forward` rejects this event name (older `gh`, GitHub-side limitation, or removed support), log a clear warning at INFO level and continue without it. The safety-net poll catches board-column changes within `WebhookIdleCap` minutes when this event is unavailable — correctness is preserved, only latency is affected. If Research finds `gh webhook forward` does not support this event, capture the finding in the ADR and document the resulting board-column-latency tradeoff in USER_GUIDE.md.
- **(R7)** Each event type maps to a typed Fabrik trigger. For this release, all triggers enqueue a full board poll — targeted per-issue action without a board re-fetch is a follow-up issue:
  - Comment, label/state, and PR events → enqueue an immediate wake via `wakeCh` (triggers a full board poll; the existing `itemMayNeedWork`/`itemNeedsWork` filter logic naturally focuses work on the changed issue).
  - Check-run completion → enqueue an immediate wake via `wakeCh` if the SHA matches a tracked branch head.
  - `projects_v2_item` → enqueue an immediate wake via `wakeCh`.
- **(R8)** The triggers feed into the existing `wakeCh` mechanism so the poll loop wakes immediately. A webhook event is a wakeup source, not a separate executor — do not invent a parallel dispatch path. The existing board-fetch-and-filter logic handles focusing work on affected items. `wakeCh` is buffered at 1; rapid webhook bursts are correctly deduplicated (only one wake is needed).

### Idle-backoff coupling

- **(R9)** When the webhook stream is in `WebhookStreamHealthy` or `WebhookStreamStartingUp` state, extend the idle-backoff cap from the current 5 minutes to `WebhookIdleCap` (default 60 minutes). The engine still polls at the cap interval as a safety net for missed events, but the steady-state GraphQL load drops by an order of magnitude.
- **(R10)** When the webhook stream is `WebhookStreamUnhealthy` or webhook mode is disabled, the existing 5-minute idle cap remains in force.
- **(R11)** Activity-driven backoff reset behavior is unchanged: any dispatched work or any received webhook event resets `idleStart` and the multiplier back to 1×.

### Signature failure and secret rotation

- **(R12)** Track consecutive HMAC verification failures. Threshold for treating failures as a secret-rotation event: 5 consecutive failures within 2 minutes.
- **(R13)** On threshold breach: log a warning at `[webhook]` level, kill the `gh webhook forward` subprocess, generate a fresh 32-byte random secret, and restart the subprocess with the new secret. Reset the failure counter. The new subprocess re-registers the webhook with GitHub; any events queued during the brief restart window are covered by the safety-net poll.
- **(R14)** If signature failures persist past two consecutive restart cycles (i.e., regeneration did not fix it), give up on webhook mode for this engine session and fall back to polling-only. Log a clear error suggesting the user check `gh auth status` and restart Fabrik. Webhook unavailability is always non-fatal.

### Configuration and discoverability

- **(R15)** New CLI flag `--webhooks` (default off) and env var `FABRIK_WEBHOOKS=true` to enable webhook mode. The feature is opt-in for the first release. Also settable via `.fabrik/config.yaml` as `webhooks: true`, following the same precedence pattern as existing flags (flag > env var > config.yaml).
- **(R16)** New CLI flag `--webhook-port=<port>` and env var `FABRIK_WEBHOOK_PORT` (default 0 = auto-pick) for the local HTTP listener. Also settable via `.fabrik/config.yaml`.
- **(R17)** New CLI flag `--webhook-events=<csv>` for advanced overrides of the default event set; default is the full list in R6.
- **(R18)** `WebhookHealthWindow` (default 10 minutes), `WebhookIdleCap` (default 60 minutes), `WebhookStartupGrace` (default 30 seconds), and the secret-rotation thresholds (5 failures / 2 minutes / 2 cycles) are internal constants for the first release. They may be promoted to user-configurable flags in a follow-up if needed.
- **(R19)** USER_GUIDE.md gets a new section explaining webhook mode: prerequisites (`gh ≥ 2.32.0`, `gh auth login`, repo admin to subscribe to webhooks), how to enable, what events are subscribed to, the three health states and their TUI indicators, expected GraphQL load reduction, and how to verify the stream is healthy (a `[webhook]` tag in `fabrik.log` plus the TUI indicator). If `projects_v2_item` is found to be unsupported, document the board-column-latency tradeoff in this section.

### Observability

- **(R20)** TUI footer or status line gains a webhook health indicator with three visual states:
  - Green dot: `WebhookStreamHealthy`
  - Blue/spinner: `WebhookStreamStartingUp`
  - Yellow dot: `WebhookStreamUnhealthy`
  - No indicator: webhook mode disabled
  - Tooltip / help-panel entry explains each state.
- **(R21)** Per-event log lines under the `webhook` tag in `fabrik.log` (event type, repo, issue/PR number, action). At INFO level. No payload contents in logs (PII / secret leakage risk).
- **(R22)** A counter of webhook events received per type, exposed in the existing TUI stats pane or a new sub-pane, so users can see the order-of-magnitude reduction in polling.

### Multi-repo

- **(R23)** Multi-repo boards (`cfg.Repo == ""`) must subscribe to events from every repo known at startup. Implementation: a single `gh webhook forward` subprocess is sufficient. When a new repo appears on the board during a poll, Fabrik kills the existing subprocess and restarts it with the updated repo list (using the same generated secret). The restart gap is brief; the safety-net poll covers any missed events. Log the restart with the new repo name.
- **(R24)** If a repo cannot be subscribed (admin permission missing, or `gh webhook forward` errors on startup), log a warning and continue with polling-only for that repo — other repos stay on webhook delivery. Per-repo subprocesses (for finer-grained failure isolation) are a documented follow-up if real users encounter problems with the single-subprocess restart approach.
- **(R25)** When org-level webhook subscription is available via `gh webhook forward --org=<org>` and the user's token has org-admin, prefer that over per-repo subscriptions (one subscription, all current and future repos in the org, no restarts needed on new-repo discovery). The implementation should auto-detect and use the org path when permitted.

## Scope

**In:**
- `gh webhook forward` subprocess management (lifecycle, supervision, health tracking, three-state health model)
- Startup `gh` version check (≥ 2.32.0)
- Local HTTP listener + HMAC-SHA256 signature verification
- Signature-failure tracking and secret-rotation logic
- Event-to-trigger translation feeding the existing `wakeCh`
- Idle-backoff coupling (three-state health → backoff cap selection)
- CLI flags, env vars, config.yaml support, and USER_GUIDE documentation
- TUI health indicator (three visual states) and event counter
- Multi-repo / org-webhook subscription logic (single subprocess, restart on new-repo)
- An ADR documenting the chosen transport, the polling-as-safety-net design, and any `projects_v2_item` support findings

**Out:**
- Direct in-engine action on webhook payloads without re-polling (force-poll first; payload-direct action is a follow-up issue).
- Replacing the GraphQL board fetch entirely. Webhooks are augmenting; polling stays.
- Self-hosted webhook receiver (e.g., Cloudflare Tunnel, ngrok). The transport is `gh webhook forward` only.
- Webhook-driven processing of bot-only / Fabrik-emitted events (already filtered by existing rocket-reaction / `🏭 **Fabrik` prefix logic — no new dedup needed).
- Migrating existing rate-limit telemetry / backoff thresholds for non-webhook users.
- Making `WebhookHealthWindow`, `WebhookIdleCap`, `WebhookStartupGrace`, or secret-rotation thresholds user-configurable (first release; follow-up if needed).
- Per-repo subprocesses for finer-grained multi-repo failure isolation (follow-up if single-subprocess restart approach causes real problems).

## Prior Art / Context

- **Smarter conditional polling (ETags / If-Modified-Since):** Only some endpoints support these; would be a partial improvement and doesn't address event latency. Not a substitute for webhooks; could complement them later.
- **Self-hosted reverse tunnel for webhooks (Cloudflare Tunnel, ngrok, SSH):** Functionally equivalent to `gh webhook forward` but adds infrastructure to set up and maintain. Worth supporting as a power-user option later, not as the default.
- **Adaptive per-repo poll intervals:** Already partly implemented; pushing further without webhooks delivers diminishing returns.
- **`gh webhook forward` introduced in gh v2.32.0 (October 2023):** The `--secret`, `--repo`, and `--events` flags have been present since introduction. `projects_v2_item` event support via `gh webhook forward` must be verified during Research.

## Risks / Dependencies

- **`gh webhook forward` experimental status.** The implementation must degrade gracefully if the command is removed or its CLI surface changes; webhook mode failure is always non-fatal and falls back to polling. Floor version is `gh ≥ 2.32.0`; bump in a follow-up if newer versions tighten the surface area.
- **`projects_v2_item` event availability.** If `gh webhook forward` does not support this event type, board-column changes are caught only by the safety-net poll (up to 60 min latency in idle state). This is a correctness-preserving degradation. Research must verify and document the outcome.
- **Event ordering / duplication.** Webhooks are at-least-once. Fabrik's existing dedup primitives (rocket reactions on processed comments, idempotent label mutations, `processedSet`, `inFlight` map) already handle this. The implementer must confirm no new race surfaces — specifically, a webhook wake followed immediately by a scheduled poll could both attempt the same state transition. The existing label-idempotency logic must still hold.
- **Repo admin permission gradient.** Users without admin on every repo on a multi-repo board will have a mixed-mode setup (some repos on webhooks, others on polling). The TUI indicator and logs must make this transparent.
- **Webhook receiver as attack surface.** Even bound to `127.0.0.1`, the receiver still needs HMAC verification before acting on payloads. A malicious local process could otherwise spoof events. The `127.0.0.1` binding plus HMAC is the standard defense; document why both layers exist in USER_GUIDE.md.
- **Long-running subprocess management.** Existing engine code does not currently manage long-running child processes (Claude invocations are bounded). Adding `gh webhook forward` introduces a supervisor concern. Reuse `os/exec` patterns established in `runClaude` where possible (process group, signal handling).
- **Secret rotation edge case.** If `gh auth refresh` invalidates the GitHub-side webhook registration while Fabrik is running, consecutive HMAC failures trigger the regenerate-and-restart path (R12–R14). Two failed restart cycles cause a graceful fallback to polling-only for the session.

verveguy:
## Summary

Add an optional webhook-driven event ingestion path. When enabled, Fabrik receives event payloads from GitHub, translates them into existing trigger primitives (force-poll-this-issue, process-this-comment, re-evaluate-CI), and dramatically extends idle-backoff during periods when the webhook stream is healthy. Polling continues as a safety net so missed or delayed webhooks don't strand work.

The transport for webhook delivery is the official GitHub CLI command `gh webhook forward`, which uses GitHub's own SSE-based forwarding. No third-party service, no extra account, no public ingress required — Fabrik connects outbound through `gh` and receives events on `localhost`. Authentication piggybacks on the existing `gh auth login` the user already has.

## Problem

Fabrik currently polls the GitHub Project board on a fixed cadence (`PollSeconds`, default 30s) via a single GraphQL query that fetches the entire board, every active item's comments, and linked-PR data. Per-item REST calls (`FetchCheckRuns`, `FetchPRMergeable`, `FetchLabelAppliedAt`, etc.) layer on top of that. Two consequences:

1. **GraphQL budget pressure.** Heavy boards consume the 5000-points/hour GraphQL budget quickly. Users running Fabrik alongside other gh-CLI / GraphQL workloads on the same token hit rate-limit caps. The current idle-backoff caps the interval at 5 minutes, which limits the worst case but doesn't help during active periods.
2. **Event latency.** Comments, label changes, CI completions, and review submissions are visible to Fabrik only on the next poll tick — up to 30s of delay even when nothing else is happening. Comment-driven steering ("please look at PR #18") feels sluggish; CI gate clearance is bottlenecked on poll cadence rather than actual GitHub state.

GitHub publishes webhooks for every event Fabrik cares about. Subscribing to them collapses the steady-state polling load to near zero and makes events visible to the engine within seconds of GitHub recording them.

### Approach

**Core design choices (resolved from research):**

1. **`webhookMgr` initialized lazily in `Run()`** (not `New()`). Webhook mode is opt-in; `Run()` has the log file handle and `ctx` the manager needs. Constructing in `New()` would require passing `ctx` through construction.

2. **`idleStart` reset via existing `wakeCh`** — no extra mechanism needed. The `case <-e.wakeCh` branch in `poll.go:348–358` already executes `e.idleStart = time.Time{}` on line 353. Since webhook events always send on `wakeCh`, idle reset (R11) is automatic.

3. **HTTP listener bound once, kept alive across subprocess restarts.** The listener is started before the first subprocess and reused on every restart. The subprocess gets the same `--url http://127.0.0.1:<port>/` on each invocation. This avoids port-churn and simplifies restart logic.

4. **`computeEffectiveInterval` gains a `webhookHealthy bool` parameter** that collapses StartingUp and Healthy into one `true` value for backoff purposes. A new `effectiveIdleCap(webhookHealthy bool)` helper selects between `maxIdleBackoff` (5 min) and `webhookIdleCap` (60 min).

5. **Startup grace applies per subprocess start, not per crash cycle.** If the subprocess crashes repeatedly, the health model transitions to `WebhookStreamUnhealthy` after the grace period expires with no received event. Grace is not re-applied infinitely on each crash — that would permanently suppress the `WebhookStreamUnhealthy` state.

6. **Org-level subscription (R25) is attempted first.** If `gh webhook forward --org=<org>` returns a startup error, fall back silently to per-repo `--repo` subscriptions. The org is inferred from the first repo's owner component.

7. **Single `engine/webhook.go` file** for all webhook plumbing (subprocess, HTTP, HMAC, health, rotation, multi-repo). This keeps a clean boundary from the rest of the engine.

### New/Modified Files

| File | Change |
|------|--------|
| `config/config.go` | Add `Webhooks bool`, `WebhookPort *int`, `WebhookEvents []string` fields to `ProjectConfig` |
| `engine/engine.go` | Add `Webhooks bool`, `WebhookPort int`, `WebhookEvents []string` to `Config`; add `webhookMgr *webhookManager` field to `Engine` |
| `engine/poll.go` | Add `webhookHealthy bool` param to `computeEffectiveInterval()`; update caller in `doPollCycle`; call `webhookMgr.UpdateRepos()` after each poll; start/stop webhookMgr in `Run()` |
| `engine/webhook.go` | **New file**: `webhookManager` struct, subprocess lifecycle, HTTP listener, HMAC verification, health state machine, secret rotation, multi-repo tracking, org-level detection |
| `engine/webhook_test.go` | **New file**: unit tests for HMAC, health state transitions, secret rotation |
| `cmd/root.go` | Add `--webhooks`, `--webhook-port`, `--webhook-events` flags with env var and `config.yaml` fallback precedence |
| `tui/events.go` | Add `WebhookStatusEvent{State WebhookHealthState; EventCounts map[string]int}` |
| `tui/model.go` | Handle `WebhookStatusEvent` in `Update()`; forward to footer |
| `tui/footer.go` | Add `webhookState`, `webhookEventCounts` fields to `FooterComponent`; render three-state indicator |
| `tui/styles.go` | Add blue/spinner style for `WebhookStreamStartingUp` |
| `adrs/032-webhook-event-delivery.md` | **New file**: ADR |
| `docs/state-machine.md` | Update idle-backoff section with three-state health model |
| `docs/USER_GUIDE.md` | Add "Webhook Mode" section |

### Key Decisions

- **`webhookManager` is a stand-alone struct** that holds the listener, subprocess handle, health state, failure counters, and known-repos set — all under its own mutex (`wm.mu`). It has no reference back to `Engine`; instead it receives a `logf func(int, string, string, ...any)`, `wakeCh chan struct{}`, and `emitFn func(tui.Event)` at construction. This prevents circular dependencies and makes unit testing straightforward.

- **`WebhookHealthState` type is defined in `engine/webhook.go`**, not in the `tui` package. The TUI event `WebhookStatusEvent` embeds it. Since `tui/events.go` imports no engine types today, this means `tui/events.go` must define `WebhookHealthState` itself (or accept it as `int`). To avoid an import cycle, use a `int` constant type in `tui/events.go` and an identical constant set in `engine/webhook.go` — or simply use a string type that needs no cross-package constant. **Decision: define `WebhookHealthState` as a string type in `engine/webhook.go` and send it as `string` in `WebhookStatusEvent`.** The TUI uses string comparison for the three-state display.

- **ADR 032 is warranted** — the transport choice, the `gh webhook forward` floor version, the polling-as-safety-net rationale, and the `projects_v2_item` findings all need documentation that isn't obvious from the code. ADR 003 must be explicitly superseded.

- **`seenRepos` comparison happens in `doPollCycle`** after a successful `poll()`. The engine already computes the seen-repos set inside `poll()` but discards it. The simplest approach: return `seenRepos map[string]bool` as a new field on `pollResult`, then call `webhookMgr.UpdateRepos(result.SeenRepos)` in `doPollCycle`. `UpdateRepos` restarts the subprocess only when the repo set grows.

### Task Checklist

- [ ] **Task 1**: Add `Webhooks bool`, `WebhookPort *int`, `WebhookEvents []string` to `ProjectConfig` in `config/config.go`; add corresponding `Webhooks bool`, `WebhookPort int`, `WebhookEvents []string` to `engine.Config`
- [ ] **Task 2**: Add `--webhooks`, `--webhook-port`, `--webhook-events` CLI flags to `cmd/root.go` with `FABRIK_WEBHOOKS`, `FABRIK_WEBHOOK_PORT`, `FABRIK_WEBHOOK_EVENTS` env var fallbacks and `config.yaml` field fallbacks, following the established `explicitFlags` precedence pattern
- [ ] **Task 3**: Create `engine/webhook.go`: define `WebhookHealthState` string type and three constants (`WebhookStreamStartingUp`, `WebhookStreamHealthy`, `WebhookStreamUnhealthy`); define `webhookManager` struct with fields for listener, subprocess PID, health state, secret, known repos, failure counters, event counters, and mutex; define internal constants `webhookIdleCap` (60 min), `webhookHealthWindow` (10 min), `webhookStartupGrace` (30 s), `webhookMaxRestartBackoff` (60 s), `webhookRotationFailures` (5), `webhookRotationWindow` (2 min), `webhookRotationMaxCycles` (2); implement `newWebhookManager(...)` constructor and `generateSecret() (string, error)`
- [ ] **Task 4**: Add HMAC-SHA256 verification to `engine/webhook.go`: `verifySignature(body []byte, signature, secret string) bool` using `crypto/hmac`, `crypto/sha256`, `encoding/hex`, timing-safe `hmac.Equal()`
- [ ] **Task 5**: Add HTTP listener + request handler to `engine/webhook.go`: `startListener(port int) (*net.TCPListener, int, error)` binding `127.0.0.1:<port>`; HTTP handler that reads body, calls `verifySignature`, logs `[webhook]` per-event (type, action, repo, number — no payload body), increments per-type counter, sends non-blocking on `wakeCh`, and transitions health state to `WebhookStreamHealthy` on first verified event; emit `WebhookStatusEvent` after each state change
- [ ] **Task 6**: Add subprocess lifecycle to `engine/webhook.go`: `ghVersionCheck() error` (parse `gh --version` output, compare semver against 2.32.0); `buildGhArgs(repos []string, orgMode bool, port int, secret string, events []string) []string`; `startSubprocess(ctx context.Context)` that runs `gh webhook forward`, sets `setCmdProcAttr`, pipes stderr to engine log, returns the `*exec.Cmd`; supervised restart goroutine (`supervise(ctx context.Context)`) with exponential backoff (start 1s, double, cap at 60s), resets to `WebhookStreamStartingUp` on each restart, transitions to `WebhookStreamUnhealthy` after startup grace expires without events
- [ ] **Task 7**: Add secret rotation to `engine/webhook.go` in the signature-failure path of the HTTP handler: track `consecutiveFailures int` and `firstFailureAt time.Time` (reset on any success); when threshold (5 failures in 2 min) is breached, call `rotateSecret(ctx)` which kills subprocess, generates new secret, increments `rotateCycleCount`; when `rotateCycleCount >= 2` and failures persist, set `disabled = true`, log error suggesting `gh auth status`, emit `WebhookStreamUnhealthy` status, return from supervise goroutine
- [ ] **Task 8**: Add multi-repo tracking + org-level subscription to `engine/webhook.go`: `UpdateRepos(repos map[string]bool)` method that compares against `wm.knownRepos`; restarts subprocess with new repo set when repos grows; `detectOrgMode(repos map[string]bool) (org string, ok bool)` that returns the common org owner when all repos share the same owner; on startup, try `--org=<org>` first, fall back to `--repo <r1> --repo <r2> ...` on stderr error from `gh`
- [ ] **Task 9**: Integrate `webhookMgr` into `engine/engine.go` and `engine/poll.go`: add `webhookMgr *webhookManager` field to `Engine`; in `Run()`, after log file opened and before first poll, construct manager (when `cfg.Webhooks`) and call `webhookMgr.Start(ctx)`; defer `webhookMgr.Stop()`; add `webhookHealthy bool` parameter to `computeEffectiveInterval()` with new `effectiveIdleCap(webhookHealthy bool) time.Duration` helper returning 5 min or 60 min; update `doPollCycle` to pass `webhookMgr.IsHealthyOrStartingUp()`; add `SeenRepos map[string]bool` to `pollResult` returned by `poll()`, call `webhookMgr.UpdateRepos(result.SeenRepos)` in `doPollCycle`; add `gh --version` prerequisite check log at startup when webhooks enabled
- [ ] **Task 10**: Add `WebhookStatusEvent{State string; EventCounts map[string]int}` to `tui/events.go`; handle in `tui/model.go`'s `Update()` forwarding to footer; add `webhookState string`, `webhookCounts map[string]int` fields to `FooterComponent` in `tui/footer.go`; render three-state indicator in `View()` (green dot for Healthy, blue spinner for StartingUp, yellow dot for Unhealthy, no indicator when disabled); add a blue lipgloss style to `tui/styles.go`
- [ ] **Task 11**: Write unit tests in `engine/webhook_test.go`: `TestVerifySignature` (valid HMAC, wrong secret, malformed header, empty body); `TestHealthStateTransitions` (startup grace → healthy on first event, startup grace → unhealthy on expiry, healthy → unhealthy on window expiry); `TestSecretRotationTrigger` (5 failures in 2 min → rotateCycleCount increments); `TestRotationFallback` (2 cycles → disabled=true)
- [ ] **Task 12**: Create `adrs/032-webhook-event-delivery.md`: supersedes ADR 003, documents `gh webhook forward` transport rationale, polling-as-safety-net, floor version choice, `projects_v2_item` graceful degradation, single-subprocess multi-repo restart, org-level subscription auto-detection
- [ ] **Task 13**: Update `docs/state-machine.md` idle-backoff section: document `WebhookStreamStartingUp`/`WebhookStreamHealthy`/`WebhookStreamUnhealthy` states and how each maps to `maxIdleBackoff` (5 min) vs `webhookIdleCap` (60 min)
- [ ] **Task 14**: Add "Webhook Mode" section to `docs/USER_GUIDE.md`: prerequisites (`gh ≥ 2.32.0`, `gh auth login`, `admin:repo_hook` scope, repo admin), enable flags/env/config.yaml, three health states + TUI indicator colors, expected GraphQL reduction, log verification via `[webhook]` tag, `127.0.0.1` binding + HMAC security rationale, `projects_v2_item` latency tradeoff
- [ ] **Task 15**: Run `go build -o fabrik .`, `go vet ./...`, `go test -race ./...`; fix any issues

### ADR Status

ADR 032 (`adrs/032-webhook-event-delivery.md`) is warranted and included as Task 12. No other decisions in this plan rise to ADR threshold — the subprocess pattern, HMAC choice, and config precedence all follow existing established patterns.

### Risks

1. **`projects_v2_item` support unknown at plan time.** The implementer must attempt the subscription at runtime and catch the `gh` stderr error gracefully; the ADR must record whether it worked or not. If unsupported, USER_GUIDE gets the latency-tradeoff note.

2. **`seenRepos` not currently returned from `poll()`.** Task 9 requires adding `SeenRepos map[string]bool` to `pollResult`. This is a minor internal change but must not break the existing `pollResult.Active` / `pollResult.ItemCount` logic — verify `poll()` test coverage covers the new field.

3. **Import cycle: `tui/events.go` must not import `engine`.** `WebhookHealthState` is defined as a string type in `engine/webhook.go`. The `tui.WebhookStatusEvent` carries `State string` (not the engine type), keeping the `tui` package free of engine imports. The engine side converts its typed constant to `string` when emitting.

4. **Org-level auto-detect may be slow on first startup.** If `gh webhook forward --org=<org>` takes time to fail, it could delay the first webhook subscription. Add a timeout (e.g. 5s) to the org probe.

5. **Race detector: webhookManager's state fields accessed from HTTP handler goroutine AND supervise goroutine simultaneously.** All state mutations in `webhookManager` must be guarded by `wm.mu`. The `go test -race ./...` step will catch any lapses.

6. **Note from Research on model complexity**: This feature adds ~500 lines of new concurrent Go across 10+ files. Research suggested adding the `model:opus` label. The plan has resolved all design decisions, so this is primarily an implementation concern — the user should decide whether to add `model:opus` for the Implement stage.

---
Used 18/50 turns, 10k input / 11k output tokens.

## Verification

Implemented webhook-driven event delivery via `gh webhook forward` across all 15 planned tasks. The core is `engine/webhook.go` (~712 lines) implementing: three-state health model (StartingUp/Healthy/Unhealthy), HMAC-SHA256 signature verification, supervised subprocess lifecycle with exponential-backoff restart (1s→60s), secret rotation on 5 consecutive failures within 2 min with 2-cycle fallback to polling, multi-repo tracking with org-level subscription auto-detection, and per-type event counters. The poll loop idle cap extends from 5 min to 60 min when the webhook stream is healthy/starting-up, reducing steady-state GraphQL load ~120×. TUI footer gains a three-state health indicator; CLI adds `--webhooks`, `--webhook-port`, `--webhook-events` flags. ADR 032, state-machine docs, and USER_GUIDE webhook section are complete. All tests pass with the race detector.

---

Closes #452

### DIFF
--- a/adrs/032-webhook-event-delivery.md
+++ b/adrs/032-webhook-event-delivery.md
@@ -1,0 +1,105 @@
+# ADR 032: Webhook-Driven Event Delivery via gh webhook forward
+
+**Status**: Accepted  
+**Date**: 2026-04-26  
+**Supersedes**: ADR 003 (Polling Over Webhooks)
+
+## Context
+
+ADR 003 chose polling because Fabrik runs locally and webhooks require a publicly accessible endpoint. The `gh webhook forward` command (introduced in gh v2.32.0, October 2023) eliminates that constraint: it acts as an SSE client to GitHub and forwards events as HTTP POST requests to a localhost URL. No public endpoint, no ngrok, no infrastructure changes — just the `gh` CLI the user already has.
+
+Two problems motivate adding webhooks alongside polling:
+
+1. **GraphQL budget pressure.** The existing poll loop fetches the entire board every 30s–5min, consuming a significant fraction of the 5,000-points/hour GraphQL budget during active periods. Users running Fabrik alongside other `gh` workloads on the same token hit rate-limit caps.
+
+2. **Event latency.** Comments, label changes, CI completions, and review submissions are visible to Fabrik only at the next poll tick — up to 30s of delay. Comment-driven steering feels sluggish; CI gate clearance is bottlenecked on poll cadence.
+
+## Decision
+
+Add an optional webhook-driven event ingestion path (`--webhooks` flag, disabled by default). When enabled:
+
+- Spawn `gh webhook forward` as a managed subprocess, forwarding events to a Fabrik-internal HTTP listener on `127.0.0.1:<port>` (OS-assigned by default).
+- Verify every incoming payload with HMAC-SHA256 before acting.
+- Translate all events to `wakeCh` signals — the existing poll loop handles all subsequent work.
+- Extend the idle-backoff cap from 5 minutes to 60 minutes when the webhook stream is healthy or starting up (the stream covers the events that would otherwise require frequent polling).
+- Retain the existing poll loop as a safety net for missed/delayed webhook events.
+
+## Transport: gh webhook forward
+
+`gh webhook forward` was chosen over alternatives for these reasons:
+
+- **No public endpoint needed.** The tool handles GitHub's SSE connection outbound; Fabrik only needs to listen on localhost.
+- **No new accounts or third-party services.** Authentication piggybacks on the user's existing `gh auth login`.
+- **Standard GitHub CLI.** Users already install `gh` for the existing `gh api` / `gh pr` usage in Fabrik stages. Floor version `gh ≥ 2.32.0` (October 2023) covers the vast majority of active installations.
+- **Graceful fallback.** If `gh` is missing, too old, or fails to subscribe, Fabrik logs a warning and continues in polling-only mode. Webhook unavailability is always non-fatal.
+
+Alternatives considered:
+- **Self-hosted reverse tunnel (Cloudflare Tunnel, ngrok, SSH):** Functionally equivalent but adds infrastructure the user must provision and maintain. Deferred as a power-user option.
+- **GitHub Apps webhook endpoint:** Requires GitHub App registration, complicates auth, and still needs a public URL or tunnel. Out of scope for a local CLI tool.
+
+## Polling as Safety Net
+
+Webhooks are not a replacement for polling — they are augmentation:
+
+- **At-least-once delivery.** GitHub's webhook semantics are at-least-once. Events can be delayed, duplicated, or dropped (especially during GitHub incidents). The poll loop catches anything the webhook stream misses.
+- **Board-column changes.** The `projects_v2_item` event class may not be supported by all `gh webhook forward` versions or GitHub configurations (see below). Polling at the extended 60-minute cap covers these changes with acceptable latency.
+- **Startup.** On fresh start, the first board state is always fetched by polling; webhooks only deliver delta events.
+- **Dedup.** Fabrik's existing idempotency mechanisms (rocket reactions on processed comments, label mutations, `processedSet`, `inFlight` map) already handle at-least-once delivery. A webhook wake followed immediately by a scheduled poll both converge to the same board state.
+
+## Three-State Health Model
+
+A two-state model (healthy/unhealthy) creates a false-unhealthy condition at startup before the first event arrives. Three states eliminate this:
+
+- **`WebhookStreamStartingUp`**: the first 30 seconds (`WebhookStartupGrace`) after subprocess launch. Treated as healthy for idle-cap purposes (extended 60-min cap applies). TUI shows a blue/spinner indicator.
+- **`WebhookStreamHealthy`**: at least one event received in the last 10 minutes (`WebhookHealthWindow`). TUI shows a green dot.
+- **`WebhookStreamUnhealthy`**: grace expired with no first event, or health window elapsed. Falls back to the 5-minute idle cap. TUI shows a yellow dot.
+
+Grace re-applies on every subprocess restart. This avoids permanently masking an unhealthy state from a crash-looping subprocess — after the grace period expires with no event, the engine transitions to unhealthy and the 5-minute cap resumes.
+
+## Secret Rotation
+
+`gh webhook forward` registers a webhook secret with GitHub at startup. If the secret becomes stale (e.g., `gh auth refresh` invalidated the token), HMAC verification will fail for every subsequent event. To handle this:
+
+1. Track consecutive HMAC failures within a 2-minute window.
+2. On 5 consecutive failures: log a warning, kill the subprocess, generate a fresh 32-byte random secret, restart.
+3. If failures persist after 2 restart cycles: disable webhook mode for the session and fall back to polling. Log a clear error suggesting `gh auth status`.
+
+This is a defense-in-depth mechanism; the primary cause of HMAC failures in practice is GitHub replaying events with the old secret after a token change.
+
+## `projects_v2_item` Event Support
+
+The spec requires attempting to subscribe to `projects_v2_item` for board-column changes. Empirical verification during Research was deferred to runtime detection:
+
+- At subprocess start, Fabrik includes `projects_v2_item` in the `--events` list.
+- If `gh webhook forward` emits an "invalid" or "unknown" error for this event on stderr, Fabrik removes it from the subscription, logs a clear warning, and restarts without it.
+- When `projects_v2_item` is unavailable, board-column changes are caught by the safety-net poll within `WebhookIdleCap` (60 min). Correctness is preserved; only latency is affected.
+
+This behavior is documented in USER_GUIDE.md.
+
+## Multi-Repo Subscription
+
+A single `gh webhook forward` subprocess covers all repos:
+
+- When org-level webhook access is available (`gh webhook forward --org=<org>`), Fabrik uses it. One subscription covers all current and future repos in the org, with no restarts on new-repo discovery.
+- When org-level access fails (detected by the subprocess exiting within 10 seconds of a startup), Fabrik falls back to per-repo `--repo` arguments.
+- When a new repo appears on the board during a poll, the subprocess is killed and restarted with the updated repo list. The restart gap (seconds) is covered by the safety-net poll.
+
+Per-repo subprocesses for finer-grained failure isolation were explicitly deferred. The single-subprocess restart approach is simpler; per-repo subprocesses are a follow-up if real users encounter problems.
+
+## Security
+
+The local HTTP listener binds to `127.0.0.1` only — never accessible off-box. HMAC-SHA256 verification is required on every payload before any action is taken. Both layers are necessary:
+
+- `127.0.0.1` binding prevents off-box attackers.
+- HMAC prevents malicious local processes from spoofing events.
+
+The webhook secret is generated from `crypto/rand` (32 bytes, hex-encoded) at startup and regenerated on rotation. It is never logged or persisted to disk.
+
+## Consequences
+
+- **Steady-state GraphQL reduction:** Active boards that would otherwise poll every 30s now poll at most every 60 minutes when the webhook stream is healthy. For a board with events every few minutes, GraphQL load drops by ~100x.
+- **Event latency:** Comments, label changes, and CI completions become visible within seconds of GitHub recording them, down from up to 30s.
+- **Operational prerequisite:** Users need `gh ≥ 2.32.0` with `admin:repo_hook` scope (or org admin) to use webhook mode. Failure to meet prerequisites falls back gracefully to polling.
+- **Non-fatal failures:** All webhook plumbing failures are non-fatal. The engine never blocks on webhook health.
+- **Subprocess supervision:** A new long-running subprocess (`gh webhook forward`) is added to the engine's lifecycle. It follows the `runClaude` process-group patterns but runs asynchronously for the engine's lifetime.
+- **ADR 003 is superseded** but its core insight (polling as the reliable foundation) remains: polling is now the safety net rather than the only mechanism.
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -377,7 +377,11 @@ func Execute() error {
 				fmt.Fprintf(os.Stderr, "[warn] FABRIK_WEBHOOK_PORT=%q is invalid (must be 0 or a positive integer); using OS-assigned port\n", v)
 			}
 		} else if pc.WebhookPort != nil {
-			cfg.WebhookPort = *pc.WebhookPort
+			if *pc.WebhookPort >= 0 {
+				cfg.WebhookPort = *pc.WebhookPort
+			} else {
+				fmt.Fprintf(os.Stderr, "[warn] webhook_port=%d is invalid (must be 0 or a positive integer); using OS-assigned port\n", *pc.WebhookPort)
+			}
 		}
 	}
 	if cfg.WebhookEvents == "" {
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -361,7 +361,7 @@ func Execute() error {
 			cfg.PluginDir = v
 		}
 	}
-	if !cfg.Webhooks {
+	if !explicitFlags["webhooks"] {
 		if v := os.Getenv("FABRIK_WEBHOOKS"); v != "" {
 			lv := strings.ToLower(v)
 			cfg.Webhooks = lv == "true" || lv == "1" || lv == "yes"
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -46,6 +46,9 @@ type Config struct {
 	ClaudeWaitDelay   int // seconds; 0 means use default (30)
 	DebugOutput       bool
 	PluginDir         string
+	Webhooks          bool
+	WebhookPort       int
+	WebhookEvents     string // comma-separated; empty means default event set
 }
 
 func Execute() error {
@@ -119,6 +122,9 @@ func Execute() error {
 	flag.IntVar(&cfg.ClaudeWaitDelay, "claude-wait-delay", 0, "Seconds to wait after Claude exits before recovering buffered output when grandchildren hold stdout pipe open (0 = use default of 30; also FABRIK_CLAUDE_WAIT_DELAY)")
 	flag.BoolVar(&cfg.DebugOutput, "debug-output", false, "Save Claude stage output to .fabrik/debug/ for debugging")
 	flag.StringVar(&cfg.PluginDir, "plugin-dir", "", "Path to Fabrik plugin directory (for development; overrides installed plugin)")
+	flag.BoolVar(&cfg.Webhooks, "webhooks", false, "Enable webhook-driven event delivery via gh webhook forward (requires gh ≥ 2.32.0; also FABRIK_WEBHOOKS)")
+	flag.IntVar(&cfg.WebhookPort, "webhook-port", 0, "Local port for the webhook HTTP listener (0 = OS-assigned; also FABRIK_WEBHOOK_PORT)")
+	flag.StringVar(&cfg.WebhookEvents, "webhook-events", "", "Comma-separated list of GitHub event types to subscribe to (default: all supported events; also FABRIK_WEBHOOK_EVENTS)")
 
 	if err := flag.CommandLine.Parse(os.Args[1:]); err != nil {
 		return err
@@ -355,6 +361,30 @@ func Execute() error {
 			cfg.PluginDir = v
 		}
 	}
+	if !cfg.Webhooks {
+		if v := os.Getenv("FABRIK_WEBHOOKS"); v != "" {
+			lv := strings.ToLower(v)
+			cfg.Webhooks = lv == "true" || lv == "1" || lv == "yes"
+		} else if pc.Webhooks {
+			cfg.Webhooks = true
+		}
+	}
+	if !explicitFlags["webhook-port"] {
+		if v := os.Getenv("FABRIK_WEBHOOK_PORT"); v != "" {
+			if n, err := strconv.Atoi(v); err == nil && n >= 0 {
+				cfg.WebhookPort = n
+			} else {
+				fmt.Fprintf(os.Stderr, "[warn] FABRIK_WEBHOOK_PORT=%q is invalid (must be 0 or a positive integer); using OS-assigned port\n", v)
+			}
+		} else if pc.WebhookPort != nil {
+			cfg.WebhookPort = *pc.WebhookPort
+		}
+	}
+	if cfg.WebhookEvents == "" {
+		if v := os.Getenv("FABRIK_WEBHOOK_EVENTS"); v != "" {
+			cfg.WebhookEvents = v
+		}
+	}
 
 	if cfg.Owner == "" || cfg.ProjectNum == 0 {
 		flag.Usage()
@@ -425,6 +455,19 @@ func Execute() error {
 		fmt.Fprintf(os.Stderr, "[upgrade] warning: plugin skill check failed: %v\n", err)
 	}
 
+	// Parse webhook events from comma-separated string.
+	var webhookEvents []string
+	if cfg.WebhookEvents != "" {
+		for _, ev := range strings.Split(cfg.WebhookEvents, ",") {
+			ev = strings.TrimSpace(ev)
+			if ev != "" {
+				webhookEvents = append(webhookEvents, ev)
+			}
+		}
+	} else if len(pc.WebhookEvents) > 0 {
+		webhookEvents = pc.WebhookEvents
+	}
+
 	eng, err := engine.New(engine.Config{
 		Owner:             cfg.Owner,
 		Repo:              cfg.Repo,
@@ -448,6 +491,9 @@ func Execute() error {
 		DebugOutput:       cfg.DebugOutput,
 		PluginDir:         cfg.PluginDir,
 		Stages:            stageCfgs,
+		Webhooks:          cfg.Webhooks,
+		WebhookPort:       cfg.WebhookPort,
+		WebhookEvents:     webhookEvents,
 		ReadyCh:           testReadyCh,
 	})
 	if err != nil {
--- a/config/config.go
+++ b/config/config.go
@@ -25,9 +25,12 @@ type ProjectConfig struct {
 	Yolo          bool   `yaml:"yolo"`
 	AutoUpgrade   bool   `yaml:"auto_upgrade"`
 	GitSSH        bool   `yaml:"git_ssh"`
-	TUI           *bool  `yaml:"tui"`
-	DebugOutput   bool   `yaml:"debug_output"`
-	Version       string `yaml:"version"`
+	TUI           *bool    `yaml:"tui"`
+	DebugOutput   bool     `yaml:"debug_output"`
+	Version       string   `yaml:"version"`
+	Webhooks      bool     `yaml:"webhooks"`
+	WebhookPort   *int     `yaml:"webhook_port"`
+	WebhookEvents []string `yaml:"webhook_events"`
 }
 
 // LoadProjectConfig reads .fabrik/config.yaml from CWD.
--- a/docs/USER_GUIDE.md
+++ b/docs/USER_GUIDE.md
@@ -1614,6 +1614,8 @@ On multi-repo boards, Fabrik uses a single `gh webhook forward` subprocess:
 1. If your token has org-admin access, Fabrik uses `--org=<org>` to subscribe to all repos in the org with one subscription — no restarts needed when new repos appear.
 2. Otherwise, Fabrik subscribes per-repo. When a new repo appears on the board, the subprocess is briefly restarted with the updated repo list (the safety-net poll covers any events missed during the restart).
 
+> **Startup delay on multi-repo boards**: Webhook subscriptions are established after the first board poll completes (up to one poll interval, typically 30 seconds). During this window the TUI shows the blue "starting up" indicator and the safety-net poll is in effect. This is expected — the webhook manager waits until the repo list is known before subscribing.
+
 ### Troubleshooting Webhook Mode
 
 **"prerequisite check failed"** — `gh` is missing or too old. Install or upgrade: <https://cli.github.com>
--- a/docs/USER_GUIDE.md
+++ b/docs/USER_GUIDE.md
@@ -25,7 +25,8 @@ For the authoritative engine state-machine spec (label semantics, review gate tr
 7. [Permissions](#7-permissions)
 8. [TUI Dashboard](#8-tui-dashboard)
 9. [Observability](#9-observability)
-10. [Troubleshooting](#10-troubleshooting)
+10. [Webhook Mode](#10-webhook-mode)
+11. [Troubleshooting](#11-troubleshooting)
 
 ---
 
@@ -133,7 +134,7 @@ exactly (case-sensitive). The default pipeline uses:
 > ```bash
 > rm -rf ~/.claude/plugins/cache/claude-plugins-official/superpowers
 > ```
-> See [§10 Troubleshooting → Duplicate Comments on Issues](#10-troubleshooting) for full details.
+> See [§11 Troubleshooting → Duplicate Comments on Issues](#11-troubleshooting) for full details.
 
 ### First Run
 
@@ -299,13 +300,13 @@ experience as release binaries.
 
 On startup, Fabrik attempts to fetch the project board and status field so it can compare stage names in your YAML configs against the column names on the board. When that metadata is fetched successfully, any non-cleanup stage missing from the board causes Fabrik to exit with a detailed error listing the mismatched names — catching config drift before any work begins. Extra board columns without a matching stage produce a warning but do not block startup. If Fabrik cannot fetch the board or status field, it logs a warning and continues without blocking startup.
 
-See [§10 Troubleshooting → Startup Board Validation Failure](#startup-board-validation-failure) if validation succeeds and reports a mismatch.
+See [§11 Troubleshooting → Startup Board Validation Failure](#startup-board-validation-failure) if validation succeeds and reports a mismatch.
 
 ### Instance Lock
 
 > **Note:** When Fabrik starts, it creates a PID lock file at `.fabrik/fabrik.lock`. If a second instance attempts to start in the same directory, it reads the lock file, logs an error identifying the running process, and exits immediately. The lock is automatically released when the process exits — including on crash or SIGKILL — so there is no need to manually delete the file after an unclean shutdown.
 >
-> See [§10 Troubleshooting → Multiple Fabrik Instances](#10-troubleshooting) if you encounter a stale lock or need to run multiple instances against different projects.
+> See [§11 Troubleshooting → Multiple Fabrik Instances](#11-troubleshooting) if you encounter a stale lock or need to run multiple instances against different projects.
 
 ---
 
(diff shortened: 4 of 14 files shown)


verveguy:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-452 | commit: 516bfe59 | main: dde43e60e6903e2b1772ce077d664a4ccc570a92 | 2026-04-26 23:16 UTC*

## Review Findings

### Fixed

- **Issue**: Data race in `webhookManager.supervise` — `detectOrgMode(wm.repos)` was called after releasing `wm.mu`, allowing a concurrent `UpdateRepos()` call to write `wm.repos` simultaneously. This is a real race that `go test -race` would catch under concurrent load. **Fix**: Moved `detectOrgMode(wm.repos)` inside the mutex-locked section, where the `repos` slice is already being built.

- **Issue**: `wm.logFn` called while holding `wm.mu` in two places — `handleWebhook` (on `StartingUp`→`Healthy` transition) and `checkHealthTransitions` (on any state transition). `logFn` is `e.logf` which acquires `e.logMu`; calling injected functions inside a lock violates lock discipline and creates latent deadlock risk if the lock ordering ever changes. **Fix**: State writes moved inside the lock, logging moved to after `Unlock()`.

- **Issue**: `--webhooks` flag used `!cfg.Webhooks` instead of `!explicitFlags["webhooks"]` to guard env var / config.yaml lookup. Since `false` is the zero value, this is equivalent to the default — but it means `--webhooks=false` on the command line cannot override `FABRIK_WEBHOOKS=true` in the environment (the env var would still win). Every other bool flag uses `!explicitFlags[...]`. **Fix**: Changed to `!explicitFlags["webhooks"]`, matching the established pattern.

### Verified

- Build clean (`go build ./...`, `go vet ./...`) — no issues before or after fixes.
- Tests pass with race detector (`go test -race -timeout 5m ./...`) — all 10 packages green, including the 401-line `engine/webhook_test.go` covering HMAC verification, semver parsing, org detection, health state transitions, secret rotation, event counting, and arg construction.
- Rebased onto latest main — branch is current, no conflicts.
- All 15 planned tasks implemented: CLI flags, config.yaml fields, webhook.go (~712 lines), TUI three-state indicator, ADR 032, state-machine docs update, and USER_GUIDE webhook section.
- Spec compliance spot-checked: three-state health model (R4), HMAC verification (R2), subprocess lifecycle with exponential backoff 1s→60s (R3), `projects_v2_item` graceful degradation (R6), wakeCh integration (R7/R8), idle cap extension 5 min→60 min (R9/R10), secret rotation logic 5 failures/2 min/2 cycles (R12–R14), org-level auto-detection (R25).

Reviewed webhook-driven event delivery implementation (15 tasks, ~1500 lines). Fixed 3 issues: a data race in `supervise` where `wm.repos` was read outside the mutex, `logFn` calls inside `wm.mu` in two places (lock discipline violation), and `--webhooks` flag using wrong guard condition preventing `--webhooks=false` from overriding env vars. All tests pass with the race detector.

---
Used 35/50 turns, 15k input / 17k output tokens.

verveguy:
*branch: fabrik/issue-452 | commit: 4ebb1902 | main: dde43e60e6903e2b1772ce077d664a4ccc570a92 | 2026-04-26 23:28 UTC*

All changes pushed. Here is the full review comment output:

---

## Copilot Review — Applied Fixes and Decisions

### Applied [8 of 9 threads]

**Thread sltY** — Race in `supervise`: `detectOrgMode(wm.repos)` called without holding `wm.mu`. **Already fixed by Review stage** (moved inside the lock). Resolved.

**Thread sltd** — Webhook manager constructed with `repos=nil`: the first `gh webhook forward` invocation launched with no `--repo` or `--org` args, causing it to fail or subscribe to the wrong target until `UpdateRepos` was called after the first poll. **Fixed**: seeded with `cfg.Owner/cfg.Repo` when both are set. Multi-repo boards still discover additional repos via `UpdateRepos`.

**Thread sltf** — On `disabled=true`, state was not set to `WebhookStreamUnhealthy` and the subprocess was not killed. TUI would display a stale health state and the supervisor goroutine would stay blocked in `cmd.Wait()`. **Fixed**: now atomically sets `state = WebhookStreamUnhealthy` and kills the current subprocess when disabling (combined into the slts+sltu fix below).

**Thread slth** — `webhook_port` loaded from `config.yaml` was assigned without validation; a negative value would fail silently at listener bind time. **Fixed**: added non-negative check with `[warn]` log line, matching the existing env-var validation.

**Thread sltl** — `state-machine.md` rate-limit bullet said "raised to `max(PollSeconds, 60s)`" but the actual code doubles the configured interval up to `rateLimitMaxBackoffMultiplier×`. **Fixed**: corrected the description.

**Thread slto** — `state-machine.md` table said StartingUp is "within startup grace (30s)" implying it only lasts 30s, but the code correctly keeps it until `startupGrace (30s) + healthWindow (10 min)` with no event. **Fixed**: updated table and transition entries to describe the actual timing.

**Thread slts** — HMAC failure window never reset: `firstFailureAt` was only set when `consecutiveFailures == 0`, so if an isolated failure occurred in the past, subsequent bursts anchored to that stale `firstFailureAt` and never satisfied `now.Sub(firstAt) <= webhookRotationWindow`. **Fixed**: reset the window when `now.Sub(firstFailureAt) > webhookRotationWindow` (combined with sltu).

**Thread sltu** — Concurrent `handleWebhook` goroutines read `failures`/`rotateCycleCount`/`disabled` snapshots under the lock then acted after unlocking, allowing multiple goroutines to simultaneously decide to rotate or disable. **Fixed**: the entire threshold check, `shouldDisable`/`shouldRotate` decision, and state mutations are now performed atomically within one `wm.mu.Lock()` section. Added `TestFailureWindowReset` and extended `TestRotationFallback` to cover the new behaviors.

### Rejected [1 thread, left open for human review]

**Thread sltp** — Copilot suggested changing `> webhookStartupGrace+webhookHealthWindow` to `> webhookStartupGrace` so that `StartingUp → Unhealthy` fires after 30 seconds instead of 10m 30s.

**Reason for rejection**: The spec (Q5 in the issue comments) explicitly says: *"After grace expires with no event received, the standard health window (R4, default 10 min) takes over from that point. If no event arrives in the next 10 min, the stream transitions to WebhookStreamUnhealthy."* Transitioning at just 30s would be far too aggressive and would defeat the purpose of keeping the 60-min idle cap for the first ~10 minutes after subprocess start. The code is intentionally correct; the doc confusion was resolved in thread slto. Leaving thread sltp open for the human reviewer to confirm.

### Verified

- Build, vet, and `go test -race -timeout 5m ./...` — all 10 packages pass after all fixes.
- 8 of 9 Copilot review threads resolved; thread sltp left open per above.

---
**Threads addressed:**
- `engine/webhook.go:390` — resolved
- `engine/poll.go:267` — resolved
- `engine/webhook.go:632` — resolved
- `cmd/root.go:380` — resolved
- `docs/state-machine.md:891` — resolved
- `docs/state-machine.md:916` — resolved
- `engine/webhook.go:538` — resolved
- `engine/webhook.go:614` — resolved
- `engine/webhook.go:636` — resolved

Resolved 9 review thread(s) across 9 comment(s).

verveguy:
*branch: fabrik/issue-452 | commit: 4ebb1902 | main: dde43e60e6903e2b1772ce077d664a4ccc570a92 | 2026-04-26 23:35 UTC*

All checks pass. Let me compile the validation report.

---

## Validation Report

### Requirements: 25/25 passed

- [x] **R1** — HTTP listener bound to `127.0.0.1:<port>` in `startListener()` (webhook.go:243–249); `127.0.0.1` hardcoded, never 0.0.0.0.
- [x] **R2** — HMAC-SHA256 verification via `verifySignature()` (webhook.go:140–151) using stdlib `crypto/hmac`, timing-safe `hmac.Equal()`, `X-Hub-Signature-256` header. Secret generated with `crypto/rand` (webhook.go:130–136).
- [x] **R3** — Supervised subprocess in `supervise()` goroutine (webhook.go:375–500); exponential backoff 1s→60s; stderr piped to engine log; `killProcGroupGraceful` on stop; non-fatal to engine.
- [x] **R4** — Three-state model: `WebhookStreamStartingUp/Healthy/Unhealthy` constants (webhook.go:27–29); `checkHealthTransitions()` (webhook.go:530–555) with correct 30s grace + 10m window timing; first event → Healthy; grace expiry → Unhealthy after `startupGrace + healthWindow`; state transitions logged.
- [x] **R5** — `ghVersionCheck()` (webhook.go:154–177) with `semverAtLeast()` comparing against 2.32.0; clear error message; falls back to polling-only.
- [x] **R6** — All 8 event types in `defaultWebhookEvents` (webhook.go:47–55); `projects_v2_item` graceful degradation via stderr detection (webhook.go:479–494).
- [x] **R7/R8** — All events send non-blocking to `wakeCh` (webhook.go:696). No parallel dispatch path invented.
- [x] **R9/R10** — `effectiveIdleCap(webhookHealthy bool)` (poll.go:73–77) returns 60min when healthy/starting-up, 5min otherwise. `IsHealthyOrStartingUp()` (webhook.go:325) collapses both healthy states for backoff.
- [x] **R11** — Idle reset is automatic: webhook events send on `wakeCh` which clears `idleStart` in the existing poll.go select handler (poll.go:353).
- [x] **R12/R13** — Consecutive failure tracking with window reset logic (webhook.go:607–652); 5 failures in 2 min triggers `rotateSecret()`.
- [x] **R14** — After 2 rotation cycles with continuing failures: `disabled=true`, state→Unhealthy, subprocess killed, clear log message (webhook.go:626–650).
- [x] **R15** — `--webhooks` flag with `FABRIK_WEBHOOKS` env var and `config.yaml` `webhooks: bool` field; `explicitFlags["webhooks"]` guard for correct precedence (cmd/root.go:364).
- [x] **R16** — `--webhook-port` / `FABRIK_WEBHOOK_PORT` / `webhook_port` in config.yaml; default 0; validation with `[warn]` log.
- [x] **R17** — `--webhook-events=<csv>` / `FABRIK_WEBHOOK_EVENTS` / `webhook_events` in config.yaml.
- [x] **R18** — All thresholds are internal constants (webhook.go:34–44).
- [x] **R19** — `docs/USER_GUIDE.md` section 10 covers all required topics (prerequisites, enable flags, three health states, GraphQL reduction, log verification, security rationale, projects_v2_item tradeoff).
- [x] **R20** — TUI footer three-state indicator: green (`●`) for Healthy, blue (`○`) for StartingUp, yellow (`◌`) for Unhealthy; lipgloss colors 42/39/214 (footer.go:51–57, styles.go:14–19).
- [x] **R21** — Per-event log at `[webhook]` level: `event: type=%s action=%s repo=%s num=%d` (webhook.go:676), no payload body.
- [x] **R22** — `eventCounts map[string]int` incremented per event (webhook.go:681); counts displayed in TUI footer indicator (footer.go:48–50).
- [x] **R23** — `UpdateRepos()` (webhook.go:328) called from `doPollCycle` after each poll (poll.go:337–338); subprocess restarted on new-repo discovery.
- [x] **R24** — Subscription failures logged and continue with polling-only; `webhookMgr.Start()` is non-fatal.
- [x] **R25** — `detectOrgMode()` (webhook.go:211); org-level probe attempted first with 10s timeout; falls back to per-repo on failure (webhook.go:385–435).
- [x] **ADR 032** — `adrs/032-webhook-event-delivery.md` created; explicitly supersedes ADR 003; covers transport rationale, safety-net polling, projects_v2_item findings, org-level detection.
- [x] **state-machine.md** — Idle-backoff section updated with three-state webhook health model and correct transition timing.
- [x] **All 15 plan tasks** — Checked off in the plan; confirmed implemented.

### Test Suite: PASSED

- 10 packages (9 with tests, 1 no test files)
- Race detector: clean
- Build: clean
- Vet: clean
- `engine/webhook_test.go`: 432 lines covering HMAC verification, semver parsing, org detection, health state transitions, secret rotation, failure window reset, event counting, arg construction

- `engine/poll_test.go` updated for the new `computeEffectiveInterval()` signature — all poll tests pass.
- All cached package results confirm no regressions in `cmd`, `config`, `github`, `stages`, `tui`, `watch`, `streamfilter`.

### Issues Found: None

Branch is clean, synced with remote (`origin/fabrik/issue-452` up to date), and rebased on main. PR #455 is open (not draft). No TODO/FIXME/debug code in new files.

### Verdict: READY TO MERGE

Validation passed: all 25 requirements verified, build/vet/tests clean with race detector across 10 packages. The webhook implementation covers the full spec — three-state health model, HMAC-SHA256 verification, supervised subprocess lifecycle, secret rotation, org-level auto-detection, multi-repo tracking, idle-backoff coupling (5min→60min), TUI three-state indicator, ADR 032, and USER_GUIDE documentation. No issues found; branch is synced with remote and rebased on main.

---
Used 35/50 turns, 18k input / 7k output tokens.

verveguy:
Validation review — 6 follow-up items found. None are ship-blockers; flagging for Fabrik to address before merge.

## 1. `_ = ctx` in `startSubprocessInternal` — replace with `exec.CommandContext`

`engine/webhook.go` (in `startSubprocessInternal`):

```go
_ = ctx // kept for future use (CommandContext replacement)
cmd := exec.Command("gh", args...)
```

The TODO-style hold-over is a code smell. Switch to `exec.CommandContext(ctx, \"gh\", args...)`. That propagates engine-context cancellation directly to the subprocess and removes the dependence on `Stop() → killProcGroup` being the only termination path. The `Stop()` path stays for cooperative shutdown, but ctx cancellation now also works.

## 2. Org-mode probe is purely time-based — add a stderr signal

`supervise()` flips `wm.orgModeFailed = true` whenever an org-mode subprocess exits within `orgModeProbeTimeout` (10s). Any other early crash within that window — transient network blip, malformed token, gh self-update, etc. — is mis-diagnosed as a permission failure and disables org mode for the rest of the engine session.

Tighten the heuristic by combining the time-based signal with a stderr content check. Look for `403`, `forbidden`, `permission denied`, `requires admin`, or `not allowed` (case-insensitive) in the captured stderr buffer for that subprocess generation. Only flip `orgModeFailed` when both conditions hold: short-lived AND auth-shaped error string.

If neither holds (just a fast exit with no auth signal), treat it as a normal crash → exponential backoff → retry org mode again. The current behavior loses org mode permanently for the session on a single transient hiccup, which is the wrong default.

## 3. `projects_v2_item` rejection detection by stderr substring is brittle

The current detection inside the stderr drainer:

```go
if strings.Contains(line, "projects_v2_item") &&
    (strings.Contains(strings.ToLower(line), "invalid") || strings.Contains(strings.ToLower(line), "unknown")) {
```

Future `gh` versions could phrase the rejection differently (`unsupported`, `not recognized`, `415`, etc.) and the detection silently misses. The fallback in that case is the supervise loop crash-restarting forever with the bad event still in the list, eventually hitting the 60s backoff cap.

Tighten with two things:
- **Broaden the stderr matcher** to include `not recognized`, `unsupported`, and `bad request` alongside `invalid` / `unknown`. Keep the `projects_v2_item` keyword as the gating term so other rejections don't accidentally match.
- **Add a second signal:** if a subprocess generation exits within `orgModeProbeTimeout` AND the event list contains `projects_v2_item`, drop that event for the next generation regardless of what the stderr said. Convergent fallback: even if stderr matching fails, the time signal will still strip the offending event after one fast crash.

Document the matcher set in the code comment so future contributors know to update it when the gh stderr format changes.

## 4. `UpdateRepos` not unit-tested

`engine/webhook_test.go` has 10 tests covering HMAC, health, rotation, handler, semver, args building, org detection — but no test for the multi-repo discovery → subprocess restart path. Add at least:

- **`TestUpdateRepos_NewRepoTriggersRestart`**: seed `wm` with `{owner/a}`, call `UpdateRepos({owner/a, owner/b})`, assert `wm.repos` now has both AND that a kill signal was sent to the current cmd (use a stub `*exec.Cmd` or a counter on a fake `killProcGroup`).
- **`TestUpdateRepos_NoNewRepos_NoRestart`**: same starting set, call `UpdateRepos` with the same set, assert no kill signal was emitted (zero-cost no-op path).
- **`TestUpdateRepos_DisabledManager_NoOp`**: with `wm.disabled = true`, calling `UpdateRepos` should not mutate state or kill anything.

If `killProcGroup` is hard to stub from a test, refactor to a method-on-wm function pointer that defaults to the real implementation but can be overridden in tests.

## 5. Multi-repo startup window — document or harden

When `cfg.Repo == \"\"` (multi-repo discovery mode), `initialRepos` is constructed empty in `Run()`:

```go
var initialRepos map[string]bool
if e.cfg.Owner != \"\" && e.cfg.Repo != \"\" {
    initialRepos = map[string]bool{e.cfg.Owner + \"/\" + e.cfg.Repo: true}
}
```

So the first `gh webhook forward` invocation has no `--repo` arguments and no `--org` (since `detectOrgMode` returns `(\"\", false)` for an empty repo set). The subprocess fails its arg validation and crashes; supervise backs off and retries. Only after the first board poll fires `UpdateRepos` does the subprocess get a real subscription.

Two options to fix:

- **Defer subprocess start until first repo set is known.** In `Start`, bind the listener and start the supervisor goroutine, but have `supervise()` block on a `repoReadyCh` (or check `len(wm.repos) == 0`) before launching the first subprocess. `UpdateRepos` (or a new method) closes the channel / signals readiness when the first non-empty repo set arrives.
- **Skip subprocess launch when no repos are subscribed.** `supervise()` checks `len(wm.repos) == 0 && !wm.orgModeFailed` at the top of each loop iteration and waits on a condition variable / channel until `UpdateRepos` adds repos.

Either approach prevents the crash-restart-backoff cycle during multi-repo startup. Add a USER_GUIDE.md sentence noting that multi-repo users see the first event 30–60s after startup (one full poll cycle delay) — that's expected, not a bug.

## 6. `detectOrgMode` semantics — clarify in code comment

`detectOrgMode` returns the common owner whenever all repos share an owner, which is true for any single-repo set as well. That's fine because the 10s probe falls back to `--repo` mode if the user isn't org-admin, but a future contributor reading the code may think it's a single-repo edge case bug.

Add a comment above the function explaining: \"Single-repo sets also satisfy this (the one repo's owner becomes the candidate org). The supervise loop's org-mode probe handles the case where the user isn't org-admin — falling back to per-repo subscription within `orgModeProbeTimeout`.\"

---

All six items are minor. Items 1, 4, and 6 are clear refactors. Items 2, 3, and 5 are correctness improvements that prevent edge-case misbehavior on rare inputs. Address them in this PR or in a small follow-up issue, validator's call.

PR #466, opened 2026-04-27
Detect drift between local .fabrik/stages and embedded defaults at startup

### LINKED ISSUE #464: Detect drift between local .fabrik/stages and embedded defaults at startup
## Problem

When stage YAML configs in the embedded plugin source gain new fields (e.g. v0.0.49 added `wait_for_ci: true` and `wait_for_reviews: true` to validate.yaml), users with existing `.fabrik/stages/` directories don't pick those fields up automatically. `fabrik init` extracts the defaults on first run, and `fabrik upgrade` refreshes `.fabrik/plugin/`, but the stage YAMLs are intentionally NOT auto-overwritten because users may have customized them.

This silently downgrades behavior. Real-world: verveguy/liminis was running v0.0.49 with a v0.0.48-era validate.yaml missing `wait_for_ci: true`. The catch-up loop fell through to the legacy `attemptMergeOnValidate` path instead of the conjunctive CI gate.

## Summary

At startup, after loading stages from `.fabrik/stages/*.yaml`, compare each loaded stage against the embedded default for the same name. If the default YAML contains keys that the user's YAML does not, emit a warning to stderr naming the missing keys and the affected stage file. Do not auto-overwrite — too risky for users with customizations. The warning makes the drift visible so users know to act.

## Requirements

- After `stages.LoadAll()` succeeds in the daemon startup path, compare each loaded user stage against the embedded default with the same `name` field.
- Comparison is **YAML-key presence only**: warn when the embedded default YAML contains a key that is absent from the user's YAML. Value differences (e.g., user has `wait_for_ci: false` explicitly set) do not trigger a warning.
- User stages with no matching embedded default (custom stages) are silently skipped — no warning, no error.
- Warning format per affected stage:
  ```
  [startup] warning: .fabrik/stages/<filename>.yaml is missing fields present in v<version> defaults: <key1>, <key2>. Run `fabrik upgrade` to see what changed.
  ```
  Where `<version>` is the current binary version (the `Version` package-level variable already used in root.go), and `<key1>`, `<key2>` are the missing top-level YAML keys.
- Warning is printed to stderr (consistent with existing `[warn]` output in root.go).
- The check is a no-op if `.fabrik/stages/` does not exist (no `fabrik init` done yet) — this is already handled by the startup flow, but the drift check should not fail if stage loading produced zero results.
- Test: Load a stage YAML missing a field present in the embedded default → warning fires naming the field. Load a stage YAML with all default fields present → no warning. Load a stage YAML where the missing field is explicitly set to a non-default value → no warning.
- The `fabrik refresh-stages` subcommand is out of scope; this issue covers warning only.

## Scope

**In scope:**
- Startup drift detection (key-presence comparison) after `stages.LoadAll()`
- Warning output to stderr naming missing keys and the affected `.yaml` file
- Unit tests for the comparison function

**Out of scope:**
- Auto-overwriting user stage YAMLs
- `fabrik refresh-stages` subcommand (follow-up issue)
- Value drift (user has a key present but with a different value than the default)
- Nested-key comparison (only compare top-level YAML keys; nested objects like `completion.type` are not individually flagged)

## Open Questions

None — all questions resolved.

## Prior Art / Context

The existing `checkPluginSkills` / `diffingPluginFiles` pattern in `cmd/upgrade.go` performs a similar drift check for plugin skill files using SHA-256 comparison. The stage drift check is analogous but operates at the YAML-key level rather than full-file level, because users are expected to customize stage values (especially `prompt`, `max_turns`, etc.) while still needing new behavioral keys.

The embedded defaults live in `stages/examples/*.yaml` and are exposed via `stages.DefaultStages` (`embed.FS`). Each file has a `name:` field; this is how user stages are matched to defaults.

The `*bool` pointer pattern used for `WaitForCI`, `WaitForReviews`, `AutoAdvance`, and `DisableAdaptiveThinking` already supports `nil`-vs-set distinction at the struct level, but raw YAML-key parsing is cleaner and handles all field types uniformly.

## Risks / Dependencies

- Low risk: warning-only, no mutations to user files.
- The embedded `stages/examples/` files must stay named consistently with the `name:` fields they contain for the match to work.
- The filename in the warning (e.g., `validate.yaml`) must be derived from the user's loaded file path, not inferred from the stage name — users may have renamed their stage files.

verveguy:
## Summary

At startup, after loading stages from `.fabrik/stages/*.yaml`, compare each loaded stage against the embedded default for the same name. If the default YAML contains keys that the user's YAML does not, emit a warning to stderr naming the missing keys and the affected stage file. Do not auto-overwrite — too risky for users with customizations. The warning makes the drift visible so users know to act.

## Problem

When stage YAML configs in the embedded plugin source gain new fields (e.g. v0.0.49 added `wait_for_ci: true` and `wait_for_reviews: true` to validate.yaml), users with existing `.fabrik/stages/` directories don't pick those fields up automatically. `fabrik init` extracts the defaults on first run, and `fabrik upgrade` refreshes `.fabrik/plugin/`, but the stage YAMLs are intentionally NOT auto-overwritten because users may have customized them.

This silently downgrades behavior. Real-world: verveguy/liminis was running v0.0.49 with a v0.0.48-era validate.yaml missing `wait_for_ci: true`. The catch-up loop fell through to the legacy `attemptMergeOnValidate` path instead of the conjunctive CI gate.

### Approach

Add a startup drift check in `stages/drift.go` that compares each user stage YAML against the embedded default with the same `name:` field at the **top-level key** level. A new `WarnStageDrift` function accepts an `io.Writer` for testability (following the `checkPluginSkillsWithReader` pattern in `cmd/upgrade.go`). The `Stage` struct gets a `FilePath` field so the function can re-read the user's file as raw YAML for key extraction. The check is wired into the daemon startup path in `cmd/root.go` immediately after the `len(stageCfgs) == 0` guard.

**Resolved questions from research:**

- **Warning prefix**: Use `[startup] warning:` exactly as specified. This deliberately diverges from the generic `[warn]` prefix to signal that this warning fires at startup, not during engine operation, and matches the format the spec's warning message prescribes.
- **io.Writer**: Accept `io.Writer` parameter (not hardcode `os.Stderr`) — enables table-driven tests without TTY hacks. `cmd/root.go` passes `os.Stderr`.
- **`completion:` key**: No special-casing. The spec says "key presence only" — if the embedded default has `completion:` and the user's file doesn't, the warning fires. This is correct: it tells users their stage file predates an important structural field. The runtime `loadOne` defaulting does not suppress the warning.
- **File organisation**: New `stages/drift.go` (not extending `stages.go`) keeps drift logic separated from loading/validation.
- **Key extraction**: Re-read user YAML from disk via `Stage.FilePath` as raw bytes → unmarshal to `map[string]interface{}`. One extra `os.ReadFile` per stage at startup is negligible. Avoids making `Stage` heavier.
- **Key sort order in warning**: Sort alphabetically for deterministic output and test assertions.

### New/Modified Files

| File | Change |
|------|--------|
| `stages/stages.go` | Add `FilePath string \`yaml:"-"\`` to `Stage` struct; set it in `loadOne` |
| `stages/drift.go` | New file: `WarnStageDrift` function + internal `missingKeys` helper |
| `stages/drift_test.go` | New file: unit tests for drift detection |
| `cmd/root.go` | Call `stages.WarnStageDrift(stageCfgs, Version, os.Stderr)` after `len(stageCfgs) == 0` guard |
| `CLAUDE.md` | Add entry to "Common Issues" explaining the drift warning |

### Key Decisions

- **`[startup] warning:` prefix**: Follows the spec exactly. Deviates from `[warn]` to distinguish startup checks from runtime engine warnings.
- **`io.Writer` over `os.Stderr` direct write**: Matches the `checkPluginSkillsWithReader` test pattern already established in `cmd/upgrade.go`. The public call site passes `os.Stderr`.
- **No `completion:` special-case**: The spec says top-level key presence comparison, no exceptions. If `completion:` is absent from a user stage, the warning is informative and correct — the user should know.
- **New file `stages/drift.go`**: Keeps drift logic out of `stages.go` (which owns loading/validation). Easier to find and modify independently.
- **FilePath in `Stage` struct**: The simplest way to propagate the source path from `loadOne` to the drift function without changing `LoadAll`'s signature. The `yaml:"-"` tag means it is never parsed from or written to YAML.
- **No ADR warranted**: The design (top-level-key warning, no auto-overwrite, stderr-only) is straightforward and already documented in the spec. No surprising constraint on future contributors that code comments and the spec don't already explain.

### Task Checklist

- [ ] **Task 1**: Add `FilePath string \`yaml:"-"\`` to the `Stage` struct in `stages/stages.go`; assign `s.FilePath = path` in `loadOne` before returning
- [ ] **Task 2**: Create `stages/drift.go` with:
  - `WarnStageDrift(userStages []*Stage, version string, w io.Writer)` — iterates over `DefaultStages` embedded files, parses each as `map[string]interface{}` to extract `name` and key set, matches against user stages by `name`, reads user stage from `Stage.FilePath`, diffs top-level keys, writes `[startup] warning: .fabrik/stages/<filename>.yaml is missing fields present in v<version> defaults: <k1>, <k2>. Run \`fabrik upgrade\` to see what changed.\n` to `w` for any stage with missing keys; custom stages (no matching embedded default) are silently skipped; empty `userStages` is a no-op
  - Internal helper `missingTopLevelKeys(userPath string, defaultKeys map[string]bool) ([]string, error)` — reads user file, unmarshals to map, returns sorted slice of keys present in `defaultKeys` but absent from user map
- [ ] **Task 3**: Create `stages/drift_test.go` with table-driven tests (using `writeStageFile` helper pattern):
  - Missing key present in embedded default → warning fires naming the key (use a synthetic embedded map, not the real `DefaultStages`, to avoid coupling to default YAML churn)
  - Stage has all keys from embedded default → no warning output
  - Stage has the key explicitly set to a value → no warning (key is present)
  - Custom stage (name not in any embedded default) → no warning
  - Empty `userStages` → no warning, no error
- [ ] **Task 4**: In `cmd/root.go`, insert `stages.WarnStageDrift(stageCfgs, Version, os.Stderr)` immediately after the `len(stageCfgs) == 0` guard (line ~384), before the TUI setup block
- [ ] **Task 5**: Add a "Stage drift warning at startup" entry to the "Common Issues" section of `CLAUDE.md` (in the project root) explaining that the warning fires when `.fabrik/stages/*.yaml` is missing fields added in a newer binary, and directing users to run `fabrik upgrade` then manually merge new fields into their customized stage files

### Risks

- **Spurious `completion:` warnings for older users**: Users whose stage files predate the `completion:` key will see a warning. This is correct behavior per spec — it informs them their file is behind. Not a code risk, but worth noting in the CLAUDE.md common-issues entry so users aren't confused.
- **Test coupling to real embedded defaults**: If tests use `DefaultStages` directly, they'll need updating every time a new field is added to the embedded defaults. **Mitigation**: tests in `stages/drift_test.go` should construct synthetic embedded key sets (pass a `map[string]struct{}` or similar) rather than calling `WarnStageDrift` end-to-end against `DefaultStages`. The end-to-end wiring in `cmd/root.go` is simple enough to not need a separate integration test.
- **Key extraction approach depends on `FilePath`**: If `Stage.FilePath` is ever empty (e.g., stages created in-memory for tests), the re-read will fail. The drift function should silently skip stages with empty `FilePath`. This is not a production risk (only `LoadAll` creates stages, and it always sets `FilePath`), but defensive handling avoids panics in test scenarios.

### Documentation Impact

- `CLAUDE.md` "Common Issues" section: Add entry for drift warning (Task 5).
- `docs/state-machine.md` and `docs/stage-lifecycle.md`: No changes needed — the drift check runs before the engine starts and doesn't affect any state-machine or stage-lifecycle behavior.

---
Used 14/50 turns, 4k input / 5k output tokens.

## Verification

Implemented startup stage drift detection: added `FilePath` to `Stage` struct, created `stages/drift.go` with `WarnStageDrift` (using `DefaultStages` FS) and `missingTopLevelKeys` helpers, wrote 9 passing tests in `stages/drift_test.go` using `testing/fstest.MapFS` for synthetic FS injection, and wired `stages.WarnStageDrift(stageCfgs, Version, os.Stderr)` into the daemon startup path in `cmd/root.go`. Updated `CLAUDE.md` and `docs/USER_GUIDE.md` with a "Stage YAML Drift Warning" section explaining the warning and how to resolve it.

---

Closes #464


verveguy:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-464 | commit: 9aec0a3f | main: 86c492b83c017e54b6ddfcf55902d24050dda0a2 | 2026-04-27 22:40 UTC*

All 9 drift tests and all 3 `missingTopLevelKeys` tests pass. Full test suite (all 9 packages) passes with the race detector. No issues found.

## Review Findings

### Verified

- **All tasks completed per plan**: `FilePath` added to `Stage` struct with `yaml:"-"` tag, `stages/drift.go` created with `WarnStageDrift` + `missingTopLevelKeys`, `stages/drift_test.go` with 9 tests covering all required cases, `cmd/root.go` wired after `len(stageCfgs) == 0` guard, `CLAUDE.md` and `docs/USER_GUIDE.md` updated.

- **Spec compliance**: Warning format matches exactly — `[startup] warning: .fabrik/stages/<filename>.yaml is missing fields present in <version> defaults: <k1>, <k2>. Run \`fabrik upgrade\` to see what changed.` — with alphabetically sorted keys for deterministic output.

- **Key-presence-only semantics**: `missingTopLevelKeys` correctly compares map presence, not values. A key set to `false` explicitly is not flagged.

- **Testability architecture**: `warnDriftFrom` accepts `fs.FS` for injection; tests use `testing/fstest.MapFS` synthetic FS — decoupled from `DefaultStages` churn.

- **Edge cases handled**: empty `userStages` is a no-op; stages with no matching default are silently skipped; empty `FilePath` (in-memory stages) is silently skipped; unreadable embedded files are skipped (best-effort walk).

- **Go conventions**: No `testify`, stdlib only, errors wrapped with context, `yaml:"-"` tag prevents YAML round-trip pollution.

- **All tests pass**: 9 packages, race detector enabled, 144s engine tests included.

- **Build and vet clean**: `go build ./...` and `go vet ./...` both succeed with no warnings.

- **Branch is current**: Already rebased onto `origin/main` (no gap).

### No Fixes Required

The implementation is clean, correct, and complete. No issues to fix.

Reviewed startup stage drift detection implementation (PR 466). All 9 test cases pass (including race detector), build and vet are clean. Implementation correctly performs key-presence-only comparison between user stage YAMLs and embedded defaults, with sorted alphabetical output, proper `io.Writer` injection for testability, and all edge cases handled. No fixes required — PR is ready for human review.

---
Used 18/50 turns, 7k input / 6k output tokens.

verveguy:
*branch: fabrik/issue-464 | commit: 018a0b02 | main: 86c492b83c017e54b6ddfcf55902d24050dda0a2 | 2026-04-27 22:42 UTC*

All four threads resolved. Here's a summary of what was applied:

## Copilot Review Findings — Applied

**Copilot (thread PRRT_kwDOR2bN_c59-Uw7)** — `stages/stages.go` line 127
**Fix applied**: Updated `FilePath` comment from "absolute path" to "path as provided to loadOne" — accurately reflects that `loadOne` receives a `filepath.Join` output which may be relative.

**Copilot (thread PRRT_kwDOR2bN_c59-Uxd)** — `stages/drift.go` line 91
**Fix applied**: Warning now emits `userStage.FilePath` directly instead of constructing `.fabrik/stages/<basename>`. Eliminates incorrect path when `--stages` / `FABRIK_STAGES` points to a non-default location.

**Copilot (thread PRRT_kwDOR2bN_c59-Uxu)** — `stages/drift.go` line 39
**Fix applied**: Walk callback now returns `nil` (instead of `err`) on entry errors, so an unreadable directory entry skips only that entry rather than aborting the entire walk. The outer `_ = err` binding was removed (no longer needed).

**Copilot (thread PRRT_kwDOR2bN_c59-UyB)** — `docs/USER_GUIDE.md` line 324
**Fix applied**: `completion:` bullet updated to say "for normal Claude-invoked stages" — accurately scopes the defaulting behaviour (which only applies when `cleanup_worktree` is false).

All 9 drift tests still pass. Build and vet clean. Changes pushed; all four review threads resolved.

---
**Threads addressed:**
- `stages/stages.go:127` — resolved
- `stages/drift.go:91` — resolved
- `stages/drift.go:39` — resolved
- `docs/USER_GUIDE.md:324` — resolved

verveguy:
*branch: fabrik/issue-464 | commit: 018a0b02 | main: 86c492b83c017e54b6ddfcf55902d24050dda0a2 | 2026-04-27 22:50 UTC*

All checks pass. Here is the validation report:

## Validation Report

### Requirements: 7/7 passed

- [x] **Drift check after `stages.LoadAll()`**: `stages.WarnStageDrift(stageCfgs, Version, os.Stderr)` is called at `cmd/root.go:386`, immediately after the `len(stageCfgs) == 0` guard, before TUI setup.
- [x] **YAML-key presence only (not value)**: `missingTopLevelKeys` compares map key presence only. `TestWarnStageDrift_KeyPresentWithNonDefaultValue` verifies `wait_for_ci: false` (key present, non-default value) produces no warning.
- [x] **Custom stages silently skipped**: The `byName` lookup skips stages with no matching embedded default. `TestWarnStageDrift_CustomStageSkipped` confirms.
- [x] **Warning format correct**: `[startup] warning: <FilePath> is missing fields present in <version> defaults: <k1>, <k2>. Run \`fabrik upgrade\` to see what changed.` — uses `userStage.FilePath` directly (Copilot fix applied), keys sorted alphabetically.
- [x] **Warning to stderr**: `os.Stderr` passed at call site.
- [x] **No-op when no stages loaded**: `warnDriftFrom` returns immediately when `len(userStages) == 0`.
- [x] **Unit tests**: 9 tests covering all spec-required cases — missing key, all keys present, key with non-default value, custom stage, empty input, multiple missing keys sorted, and 3 `missingTopLevelKeys` unit tests.

### Test Suite: PASSED
- 9 packages with tests (1 no test files: plugin)
- All pass including race detector (`go test -race -timeout 5m ./...`)
- Build: clean (`go build ./...` — no output)
- Vet: clean (`go vet ./...` — no output)

All pre-existing tests pass. Changes are additive: new `FilePath` field in `Stage` has `yaml:"-"` tag (no round-trip side effects), new `stages/drift.go` file, new `stages/drift_test.go`, and a single call wired into the startup path.

### Branch State
- Branch `fabrik/issue-464` is rebased onto and up to date with `origin/main`
- All commits pushed: `018a0b0` is HEAD on both local and remote
- Changed files: `CLAUDE.md`, `cmd/root.go`, `docs/USER_GUIDE.md`, `stages/drift.go`, `stages/drift_test.go`, `stages/stages.go`

### Verdict: READY TO MERGE

All 7 requirements verified: startup drift check is wired correctly in `cmd/root.go`, key-presence-only comparison is implemented in `stages/drift.go`, warning format matches spec exactly, custom stages are silently skipped, and all 9 new unit tests pass. Full test suite (10 packages, race detector) is clean; build and vet are clean. Branch is up to date with main.

---
Used 22/50 turns, 9k input / 4k output tokens.

PR #492, opened 2026-05-03
Rate-limit backoff doesn't escalate — flat 2× multiplier defeats the documented 10× cap

### LINKED ISSUE #487: Rate-limit backoff doesn't escalate — flat 2× multiplier defeats the documented 10× cap
## Problem

When the GraphQL rate-limit ratio drops below 20%, the engine logs `[poll] rate-limit backoff active — effective poll interval: <interval>` but the interval **never escalates beyond 2× the configured poll interval**, regardless of how depleted the budget gets. The documented "cap at 10× configured" is dead code.

## Observed Behaviour (Liminis instance, fabrik v0.0.53, poll: 15s)

```
[poll] rate limit GraphQL: 160/5000 remaining
[poll] rate-limit backoff active — effective poll interval: 30s
...
[poll] rate limit GraphQL: 94/5000 remaining
[poll] rate-limit backoff active — effective poll interval: 30s
...
[poll] rate limit GraphQL: 23/5000 remaining
[poll] rate-limit backoff active — effective poll interval: 30s
```

Budget cliff-dives 160 → 94 → 23 over three polls but the effective interval stays glued at 30s (= 2 × 15s configured). It should escalate toward the 150s (= 10 × 15s) ceiling as the budget gets dire.

## Root Cause (Confirmed)

`engine/poll.go` — `computeEffectiveInterval` (lines ~99–106):

```go
rateLimitInterval := configuredInterval
if rateLimitLow {
    rateLimitInterval = configuredInterval * 2          // ← fixed 2× constant
    maxRL := configuredInterval * time.Duration(rateLimitMaxBackoffMultiplier)
    if rateLimitInterval > maxRL {                      // ← always false: 2 < 10
        rateLimitInterval = maxRL
    }
}
```

The function signature takes a `rateLimitLow bool`, so it has no information about _how depleted_ the budget is — only whether the threshold was crossed. The 10× cap (`rateLimitMaxBackoffMultiplier`) is dead code because `2 < 10` is always true. The function always returns exactly 2× the configured interval when backoff is active.

The docstring (lines 82–85) and the constant comment (lines 33–35) both imply progressive escalation, confirming this is an incomplete implementation rather than deliberate behaviour.

## Test Coverage Encodes the Bug

Every test case with `rateLimitLow: true` asserts the result is exactly `2 × base`. The function signature itself (`bool`) makes it impossible to express "more depleted = larger multiplier." `TestComputeEffectiveInterval_RateLimitExceeds5Min` checks 6 min (2× of 3-min base) but never anything past 2×.

Any correct fix will require:
1. Changing `computeEffectiveInterval`'s signature to accept the actual remaining ratio (e.g. `rateLimitRatio float64`) instead of a bool.
2. Rewriting the affected test cases — they will appear as regressions against any correct implementation.

## Relationship to ADR 028

ADR 028 (`adrs/028-rate-limit-backoff-hysteresis.md`) introduced two-threshold hysteresis to prevent premature backoff clearing on busy boards. Its "Alternatives Considered" section says "the existing 2× rate-limit interval is sufficient for the problem" — but that statement addressed _thrashing/premature clearing_, not depth-of-depletion escalation. The two concerns are orthogonal. This fix adds escalation; the hysteresis behaviour (20%/50% thresholds, sticky state) is correct and unchanged.

## Summary

Fix `computeEffectiveInterval` to accept the actual remaining rate-limit ratio and implement a proper escalation curve that scales the poll multiplier from 2× (at ~20% budget) up to 10× (at or near 0% budget), capping at `rateLimitMaxBackoffMultiplier × configuredInterval`. Update the tests and `docs/USER_GUIDE.md` to match.

## Requirements

- When `rateLimitLow` is true, the effective rate-limit interval scales above 2× configured as the remaining ratio decreases, capping at `rateLimitMaxBackoffMultiplier × configured` (currently 10×).
- At or near budget = 0%, the effective interval equals the 10× cap.
- The escalation curve choice is a Research/Plan decision; three options are presented in the investigation scope below.
- Hysteresis behaviour is unchanged: backoff activates below 20%, clears only above 50%.
- `computeEffectiveInterval` signature changes to accept the actual ratio value rather than a bool, so the escalation can be computed.
- `engine/poll_test.go` is updated to cover the escalation curve; existing flat-2× assertions are rewritten to reflect the new behaviour.
- `docs/USER_GUIDE.md` Rate-limit backoff section is updated to describe the actual implemented curve.

## Scope

**In scope:**
- `engine/poll.go` — `computeEffectiveInterval` and its callers.
- `engine/poll_test.go` — existing rate-limit backoff tests and new escalation coverage.
- `docs/USER_GUIDE.md` — Rate-limit backoff section.

**Out of scope:**
- The separate "deep-fetch loop wastes GraphQL on quiescent items" issue. That's a distinct mechanism contributing to fast budget depletion.
- Idle backoff (different code path, working correctly).
- Tuning the 20%/50% hysteresis thresholds.

## Open Questions

None. All design decisions deferred to Research/Plan are called out below.

## Investigation Scope for Research

- **Confirm the bug** by reading `engine/poll.go:74–113` (`computeEffectiveInterval`) and the existing tests (`engine/poll_test.go:967–1038`).
- **Choose the escalation curve.** Three options to evaluate:
  - **Linear by depletion**: multiplier scales linearly from 2× (at 20% remaining) to 10× (at 0% remaining). Formula: `multiplier = 2 + (1 - ratio/0.20) * 8`, clamped to [2, 10].
  - **Stepwise by threshold**: 2× under 20%, 4× under 10%, 6× under 5%, 10× under 1%. Mirrors the shape of idle backoff; predictable for operators.
  - **Time-since-low**: stay at 2× for the first N minutes after activation, then escalate. Simpler signature change but couples backoff magnitude to wall time rather than quota state.
  - The idle backoff precedent (stepwise) and operator predictability are relevant selection criteria.
- **Check for related TODOs or ADRs** beyond ADR 028 — none are expected, but worth confirming.
- **Cross-check with `docs/USER_GUIDE.md` §1** to confirm what the doc claims vs. what should be implemented.

## Prior Art / Context

- The idle backoff uses a stepwise schedule (5-min buckets → 1×/2×/4×/max). A stepwise rate-limit curve would be architecturally consistent.
- Linear curves are simpler to reason about in tests and have no threshold boundaries to argue over, but operators can't predict exact intervals without a formula.
- ADR 028 confirmed the hysteresis design (correct and unchanged by this fix).
- `rateLimitMaxBackoffMultiplier = 10` is already present; no new constants are likely needed.

## Risks / Dependencies

- Changing `computeEffectiveInterval`'s signature requires updating all call sites; grep for callers before implementing.
- Test rewrite is required — the existing tests encode the bug as expected values. A naive approach of "run tests, see failures" will produce misleading output.
- `docs/USER_GUIDE.md` must be updated atomically with the code change — the doc already claims 10× cap behaviour, so the fix brings code and docs into alignment.

## Pointers

- `engine/poll.go:74–113` — `computeEffectiveInterval` (the bug).
- `engine/poll.go:23–35` — backoff constants and docstrings.
- `engine/poll.go:56–67` — hysteresis logic (correct, unchanged).
- `engine/poll_test.go:967–1038` — existing tests that encode the bug as expected values.
- `docs/USER_GUIDE.md` §1 "Rate-limit backoff" — documented behaviour.
- `adrs/028-rate-limit-backoff-hysteresis.md` — hysteresis design rationale.

verveguy:
## Summary

Fix `computeEffectiveInterval` to accept the actual remaining rate-limit ratio and implement a proper escalation curve that scales the poll multiplier from 2× (at ~20% budget) up to 10× (at or near 0% budget), capping at `rateLimitMaxBackoffMultiplier × configuredInterval`. Update the tests and `docs/USER_GUIDE.md` to match.

## Problem

When the GraphQL rate-limit ratio drops below 20%, the engine logs `[poll] rate-limit backoff active — effective poll interval: <interval>` but the interval **never escalates beyond 2× the configured poll interval**, regardless of how depleted the budget gets. The documented "cap at 10× configured" is dead code.

### Approach

Fix `computeEffectiveInterval` in `engine/poll.go` by replacing its `rateLimitLow bool` parameter with `rateLimitRatio float64`, then implement the stepwise escalation curve recommended by Research. The call site in `doPollCycle` must be updated to track and pass the actual ratio. Tests are rewritten to cover all four escalation tiers. Two docs files are updated to reflect the actual behavior.

**Escalation curve chosen: Stepwise** (Research recommendation, confirmed here)

Four tiers mirroring `idleBackoffMultiplier`'s switch-statement pattern:

| Remaining quota | Multiplier | At 30s base | At 60s base |
|---|---|---|---|
| 10%–20% | 2× | 60s | 120s |
| 5%–10% | 4× | 120s | 240s |
| 1%–5% | 6× | 180s | 360s |
| < 1% | 10× | 300s | 600s |

**Sentinel design for "not in backoff"**: `rateLimitRatio = 1.0`. The activation guard inside `computeEffectiveInterval` becomes `if rateLimitRatio < 1.0`. This avoids coupling the activation check to `rateLimitBackoffThreshold` and correctly handles the sticky zone (ratio 20%–50%: caller passes actual ratio when `rateLimitLow == true`, which lands in the default 2× tier since no escalation tiers cover that range).

**Caller design**: Add `rateLimitRatio float64 = 1.0` as a closure variable alongside `rateLimitLow` in `doPollCycle`. Inside `if graphqlStats.Limit > 0`, update `rateLimitRatio = ratio` when `rateLimitLow` is true after the hysteresis update, or reset to `1.0` when `rateLimitLow` is false. Pass `rateLimitRatio` instead of `rateLimitLow` to `computeEffectiveInterval`.

**No ADR needed**: This is an implementation bug fix completing originally-intended behavior. ADR 028 covers hysteresis (unchanged). The escalation curve choice (stepwise vs. linear) does not constrain future contributors in a non-obvious way.

### New/Modified Files

| File | Change |
|------|--------|
| `engine/poll.go` | Change `computeEffectiveInterval` signature + stepwise rate-limit logic; add `rateLimitRatio` closure var + update call site in `doPollCycle` |
| `engine/poll_test.go` | Rewrite 4 existing rate-limit test functions; add new escalation tier coverage |
| `docs/USER_GUIDE.md` | Replace rate-limit backoff paragraph with a tiered-table description (§1, lines ~169–171) |
| `docs/state-machine.md` | Update §7.7 line 936: "doubles the configured interval" → stepwise escalation description |

### Key Decisions

- **Stepwise over linear**: Architecturally consistent with `idleBackoffMultiplier`, produces exact integer expected values in tests, and operators can read the table without knowing a formula. Linear requires floating-point tolerance in assertions and the `2 + 8*(1-ratio/0.20)` formula is harder to audit under sticky-zone edge cases.

- **Sentinel `1.0` not `rateLimitBackoffThreshold`**: `computeEffectiveInterval` should not know about the hysteresis thresholds — that's `nextRateLimitLow`'s concern. Using `1.0` as the "not in backoff" sentinel keeps the function's activation check independent of the threshold constants, so the sticky-zone case (ratio 20%–50%) correctly defaults to 2× without any special handling.

- **`rateLimitRatio` as closure variable**: Mirrors the existing `rateLimitLow` closure variable. No architectural change to `doPollCycle` — just a second piece of backoff state that persists across poll cycles. The alternative (re-reading stats inside the function) would require an additional dependency injection and is out of scope.

- **`TestComputeEffectiveInterval_RateLimitExceeds5Min` rewritten but not deleted**: The test's purpose (confirming rate-limit backoff is NOT clamped by the 5-min idle cap) is still valid. The assertion changes from 6min (2× of 3min base) to a ratio-specific expected value. A ratio of `0.15` still produces 2×, so the expected value stays 6min — only the parameter type changes.

- **`TestComputeEffectiveInterval_MaxIdleRateLimit` similarly rewritten**: Passes `0.15` (2× tier) to match the original assertion (5min idle wins). An additional case should be added at `0.005` (10× tier) to confirm that `max(5min, 10×30s=300s)` = 5min still holds.

### Task Checklist

- [ ] **Task 1**: In `engine/poll.go`, rewrite `computeEffectiveInterval` — change signature from `rateLimitLow bool` to `rateLimitRatio float64`, replace the fixed-2× block (lines 99–106) with a stepwise switch, update the function docstring (lines 81–85) to document the new signature and four-tier escalation table.

- [ ] **Task 2**: In `engine/poll.go`, update `doPollCycle` — add `rateLimitRatio := 1.0` alongside `rateLimitLow` on line 303, update it inside the `if graphqlStats.Limit > 0` block after the hysteresis update (set to `ratio` when `rateLimitLow` is true, reset to `1.0` otherwise), and update the `computeEffectiveInterval` call on line 344 to pass `rateLimitRatio` instead of `rateLimitLow`.

- [ ] **Task 3**: In `engine/poll_test.go`, rewrite `TestComputeEffectiveInterval` (lines 967–994) — change the table's `rateLimitLow bool` field to `rateLimitRatio float64`, update the function call on line 988, choose representative ratio values for each case (e.g., `0.15` for 2× tier, `0.0` for "not applicable" mapped to `1.0`), and update expected values where they change.

- [ ] **Task 4**: In `engine/poll_test.go`, rewrite `TestComputeEffectiveInterval_MaxIdleRateLimit` (lines 1019–1027) — replace `true` with `0.15`; add a second assertion at `0.005` (10× tier, rate-limit = 300s = 5min, idle = 5min, result = 5min).

- [ ] **Task 5**: In `engine/poll_test.go`, rewrite `TestComputeEffectiveInterval_RateLimitExceeds5Min` (lines 1029–1039) — replace `true` with `0.15` (2× tier, 3min base → 6min, assertion unchanged); add assertions for ratio `0.07` (4× tier → 12min), ratio `0.03` (6× tier → 18min), and ratio `0.005` (10× tier → 30min).

- [ ] **Task 6**: In `engine/poll_test.go`, add a new `TestComputeEffectiveInterval_RateLimitStepwise` function covering: `1.0` (not in backoff → 1×), `0.35` (sticky zone → 2×), `0.15` (10%–20% → 2×), `0.099` (just below 10% → 4×), `0.049` (just below 5% → 6×), `0.009` (just below 1% → 10×), `0.0` (zero remaining → 10×). Use a 30s base.

- [ ] **Task 7**: In `docs/USER_GUIDE.md` (lines ~169–171), replace the current prose description of rate-limit backoff (which mentions only the 10× cap, not escalation steps) with a paragraph plus a four-row table matching the idle backoff table format above it, documenting the four quota tiers and example intervals at 30s and 60s bases.

- [ ] **Task 8**: In `docs/state-machine.md` (§7.7, line 936), update the bullet "Rate-limit adjustment: when … `computeEffectiveInterval` doubles the configured interval for rate-limit backoff, capped at `rateLimitMaxBackoffMultiplier×` the configured interval" to describe the stepwise escalation (reference the four tiers; match the language used in USER_GUIDE.md).

- [ ] **Task 9**: Run `go build ./...` and `go test -race ./...` — all tests pass, build clean.

### Risks

- **Sticky-zone test coverage gap**: The sticky zone (ratio 20%–50%, `rateLimitLow = true`) is the trickiest edge case — callers pass a ratio > 20% but the function must still apply the minimum 2× backoff. Task 6 explicitly includes `0.35` as a sticky-zone test case to guard against this.

- **`TestComputeEffectiveInterval_MaxIdleRateLimit` misleading at high escalation**: At 10× tier (300s rate-limit, 300s idle), the test returns 5min regardless of which wins. Adding the `0.005` case in Task 4 confirms the correct value, but reviewers may not notice it tests a tie. The assertion comment should clarify.

- **Docstring and docs drift**: The function docstring at lines 81–85 currently says nothing about the escalation curve. Task 1 must update it, or `docs/state-machine.md` and the code will diverge again within the same PR. The task checklist explicitly includes the docstring update in Task 1 so it is not skipped.

- **No call-site count surprise**: Research confirmed `computeEffectiveInterval` is called in exactly one production location (line 344). If a new call site was added between the research commit and implementation, the build will catch the signature mismatch at compile time. No manual grep is needed.

---
Used 10/50 turns, 5k input / 7k output tokens.

## Verification

Fixed `computeEffectiveInterval` in `engine/poll.go` to replace the hardcoded `rateLimitLow bool` parameter with `rateLimitRatio float64`, implementing a four-tier stepwise escalation (2×/4×/6×/10×) that scales as GraphQL quota depletes — bringing the 10× cap from dead code to active behavior. Updated tests, `docs/USER_GUIDE.md` (added escalation table), and `docs/state-machine.md` §7.7 to match.

---

Closes #487

### DIFF
--- a/docs/USER_GUIDE.md
+++ b/docs/USER_GUIDE.md
@@ -166,9 +166,16 @@ or `poll:` in `config.yaml`).
 | 10–20 min | 4× | 120s |
 | 20+ min | max (5 minutes) | 300s |
 
-**Rate-limit backoff**: When Fabrik detects GraphQL API quota pressure, the same interval infrastructure adds a separate rate-limit component. This component is capped at `10 × configured poll interval` (300s at the default 30s base; 600s at 60s base) — there is no separate 5-minute ceiling for the rate-limit contributor. The effective poll interval is `max(idle interval, rate-limit interval)`, so whichever is larger governs at any given moment.
+**Rate-limit backoff**: When Fabrik detects GraphQL API quota pressure, the same interval infrastructure adds a separate rate-limit component that escalates as quota depletes. There is no separate 5-minute ceiling for the rate-limit contributor — it can exceed 5 minutes when the configured base interval is large. The effective poll interval is `max(idle interval, rate-limit interval)`, so whichever is larger governs at any given moment.
 
-Rate-limit backoff uses two-threshold hysteresis to prevent thrashing: it **activates** when GraphQL remaining quota drops below **20%** of the hourly limit, and **clears only when quota rises above 50%** of the limit. Activity detection (items deep-fetched or dispatched) resets idle backoff but does NOT reset rate-limit backoff — the two concerns are independent. When rate-limit backoff is active, Fabrik logs the effective poll interval each cycle so operators can observe the actual cadence.
+| Remaining GraphQL quota | Multiplier | At 30s base | At 60s base |
+|---|---|---|---|
+| 10%–20% (backoff active) | 2× | 60s | 120s |
+| 5%–10% | 4× | 120s | 240s |
+| 1%–5% | 6× | 180s | 360s |
+| < 1% | 10× (cap) | 300s | 600s |
+
+Rate-limit backoff uses two-threshold hysteresis to prevent thrashing: it **activates** when GraphQL remaining quota drops below **20%** of the hourly limit, and **clears only when quota rises above 50%** of the limit. While quota is recovering (20%–50%, sticky zone), the 2× tier applies. Activity detection (items deep-fetched or dispatched) resets idle backoff but does NOT reset rate-limit backoff — the two concerns are independent. When rate-limit backoff is active, Fabrik logs the effective poll interval each cycle so operators can observe the actual cadence.
 
 **Board fetch resilience**: `FetchProjectBoard` automatically retries up to 3 times (with 1s/2s backoff) when GitHub returns an empty response (zero items), which occurs during transient Projects v2 indexer degradation that intermittently returns empty boards for non-empty projects (both returned node count and `totalCount` are zero in degraded responses). The log line:
 
--- a/docs/USER_GUIDE.md
+++ b/docs/USER_GUIDE.md
@@ -170,10 +170,10 @@ or `poll:` in `config.yaml`).
 
 | Remaining GraphQL quota | Multiplier | At 30s base | At 60s base |
 |---|---|---|---|
-| 10%–20% (backoff active) | 2× | 60s | 120s |
-| 5%–10% | 4× | 120s | 240s |
-| 1%–5% | 6× | 180s | 360s |
-| < 1% | 10× (cap) | 300s | 600s |
+| >=10% (incl. sticky zone 20%–50%) | 2× | 60s | 120s |
+| >=5% and <10% | 4× | 120s | 240s |
+| >=1% and <5% | 6× | 180s | 360s |
+| <1% | 10× (cap) | 300s | 600s |
 
 Rate-limit backoff uses two-threshold hysteresis to prevent thrashing: it **activates** when GraphQL remaining quota drops below **20%** of the hourly limit, and **clears only when quota rises above 50%** of the limit. While quota is recovering (20%–50%, sticky zone), the 2× tier applies. Activity detection (items deep-fetched or dispatched) resets idle backoff but does NOT reset rate-limit backoff — the two concerns are independent. When rate-limit backoff is active, Fabrik logs the effective poll interval each cycle so operators can observe the actual cadence.
 
--- a/docs/state-machine.md
+++ b/docs/state-machine.md
@@ -933,7 +933,7 @@ The poll loop's effective interval grows when there is nothing to do (idle backo
 - Base interval: `PollSeconds` (default 30s).
 - Activity multiplier: doubles each poll cycle where no work was dispatched, up to the idle cap.
 - Activity reset: any dispatched work OR any received webhook event resets `idleStart` and the multiplier back to 1×.
-- Rate-limit adjustment: when the GitHub API signals rate limit pressure (`rateLimitLow`), `computeEffectiveInterval` doubles the configured interval for rate-limit backoff, capped at `rateLimitMaxBackoffMultiplier×` the configured interval.
+- Rate-limit adjustment: when `rateLimitRatio < 1.0` (backoff active), `computeEffectiveInterval` applies a stepwise escalation — 2× at 10%–100% remaining (including the 20%–50% sticky hysteresis zone), 4× at 5%–10%, 6× at 1%–5%, and 10× (`rateLimitMaxBackoffMultiplier`) below 1%. The rate-limit component has no 5-minute ceiling; only the caller-supplied `rateLimitMaxBackoffMultiplier` cap applies.
 
 **Idle cap selection** (`effectiveIdleCap` in `engine/poll.go`):
 
--- a/docs/state-machine.md
+++ b/docs/state-machine.md
@@ -933,7 +933,12 @@ The poll loop's effective interval grows when there is nothing to do (idle backo
 - Base interval: `PollSeconds` (default 30s).
 - Activity multiplier: doubles each poll cycle where no work was dispatched, up to the idle cap.
 - Activity reset: any dispatched work OR any received webhook event resets `idleStart` and the multiplier back to 1×.
-- Rate-limit adjustment: when `rateLimitRatio < 1.0` (backoff active), `computeEffectiveInterval` applies a stepwise escalation — 2× at >=10% remaining (including the 20%–50% sticky hysteresis zone), 4× at >=5% and <10%, 6× at >=1% and <5%, and 10× (`rateLimitMaxBackoffMultiplier`) below 1%. The rate-limit component has no 5-minute ceiling; only the caller-supplied `rateLimitMaxBackoffMultiplier` cap applies.
+- Rate-limit adjustment (`nextRateLimitLow` + `computeEffectiveInterval` in `engine/poll.go`):
+  - **Activation**: rate-limit backoff engages when remaining GraphQL quota drops below 20% (`rateLimitBackoffThreshold`). The actual remaining fraction is passed to `computeEffectiveInterval` as `rateLimitRatio`.
+  - **Clearance**: backoff clears only when quota rises above 50% (`rateLimitHealthyThreshold`). Between 20% and 50% the state is sticky — backoff remains active to prevent thrashing on boards where quota fluctuates near the activation point.
+  - **Stepwise escalation**: the multiplier scales with depletion depth — 2× at >=10% remaining (includes the 20%–50% sticky zone), 4× at >=5% and <10%, 6× at >=1% and <5%, 10× (`rateLimitMaxBackoffMultiplier`) below 1%.
+  - **No idle cap**: the rate-limit component has no 5-minute ceiling; it is capped only at `rateLimitMaxBackoffMultiplier × configuredInterval`.
+  - **Independence from idle backoff**: activity detection (items dispatched or webhooks received) resets idle backoff but does NOT reset rate-limit backoff.
 
 **Idle cap selection** (`effectiveIdleCap` in `engine/poll.go`):
 
--- a/docs/state-machine.md
+++ b/docs/state-machine.md
@@ -933,7 +933,7 @@ The poll loop's effective interval grows when there is nothing to do (idle backo
 - Base interval: `PollSeconds` (default 30s).
 - Activity multiplier: doubles each poll cycle where no work was dispatched, up to the idle cap.
 - Activity reset: any dispatched work OR any received webhook event resets `idleStart` and the multiplier back to 1×.
-- Rate-limit adjustment: when `rateLimitRatio < 1.0` (backoff active), `computeEffectiveInterval` applies a stepwise escalation — 2× at 10%–100% remaining (including the 20%–50% sticky hysteresis zone), 4× at 5%–10%, 6× at 1%–5%, and 10× (`rateLimitMaxBackoffMultiplier`) below 1%. The rate-limit component has no 5-minute ceiling; only the caller-supplied `rateLimitMaxBackoffMultiplier` cap applies.
+- Rate-limit adjustment: when `rateLimitRatio < 1.0` (backoff active), `computeEffectiveInterval` applies a stepwise escalation — 2× at >=10% remaining (including the 20%–50% sticky hysteresis zone), 4× at >=5% and <10%, 6× at >=1% and <5%, and 10× (`rateLimitMaxBackoffMultiplier`) below 1%. The rate-limit component has no 5-minute ceiling; only the caller-supplied `rateLimitMaxBackoffMultiplier` cap applies.
 
 **Idle cap selection** (`effectiveIdleCap` in `engine/poll.go`):
 
--- a/engine/poll.go
+++ b/engine/poll.go
@@ -82,7 +82,16 @@ func effectiveIdleCap(webhookHealthy bool) time.Duration {
 // idle backoff and rate-limit backoff. The result is max(idle, rateLimit).
 // The idle component is capped at effectiveIdleCap(webhookHealthy); the rate-limit
 // component uses its own cap (rateLimitMaxBackoffMultiplier × configured).
-func computeEffectiveInterval(configuredInterval time.Duration, idleDuration time.Duration, rateLimitLow bool, webhookHealthy bool) time.Duration {
+//
+// rateLimitRatio is the remaining-to-total GraphQL quota fraction. Pass 1.0 when
+// no rate-limit backoff is active; pass the actual fraction when backoff is active.
+// The stepwise escalation schedule (activates when ratio < 1.0):
+//
+//	10%–100% remaining: 2× configured  (includes sticky hysteresis zone 20%–50%)
+//	 5%–10% remaining:  4× configured
+//	 1%– 5% remaining:  6× configured
+//	  < 1% remaining:  10× configured  (rateLimitMaxBackoffMultiplier)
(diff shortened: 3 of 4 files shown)


verveguy:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-487 | commit: c3650049 | main: 5f03922a808e9b34f584e1976d0b994ca8e5c3ac | 2026-05-03 03:16 UTC*

## Review Findings

### Fixed

**Copilot finding [1] — Test: missing exact boundary cases**
The stepwise test covered values just *below* each threshold (0.099, 0.049, 0.009) but not the boundary values themselves. A comparator change from `>=` to `>` would silently shift tiers without failing any test.
**Fix**: Added `{"boundary exactly 10% (0.10)", 0.10, 60 * time.Second}`, `{"boundary exactly 5% (0.05)", 0.05, 2 * time.Minute}`, and `{"boundary exactly 1% (0.01)", 0.01, 3 * time.Minute}` to `TestComputeEffectiveInterval_RateLimitStepwise`.

**Copilot finding [2] — Docstring: ambiguous boundary notation**
The docstring used dash-ranges ("5%–10%", "1%–5%") that don't clarify whether the upper value belongs to that tier or the one above. The implementation uses `>=` comparators, so 10% maps to the 2× tier, not the 4× tier.
**Fix**: Rewrote the schedule table to use explicit inequality notation (`>=10%`, `>=5% and <10%`, `>=1% and <5%`, `<1%`).

**Copilot finding [3] — USER_GUIDE.md: same boundary ambiguity**
The table had `"10%–20% (backoff active)"` / `"5%–10%"` / `"1%–5%"` — ambiguous at each boundary, and the first row incorrectly implied the sticky zone (20%–50%) doesn't get 2× backoff.
**Fix**: Updated table labels to `">=10% (incl. sticky zone 20%–50%)"`, `">=5% and <10%"`, `">=1% and <5%"`, `"<1%"`.

**Copilot finding [4] — state-machine.md: same boundary ambiguity**
The §7.7 description used "4× at 5%–10%", "6× at 1%–5%" with the same ambiguity.
**Fix**: Updated to "4× at >=5% and <10%", "6× at >=1% and <5%".

All four Copilot review threads have been resolved.

### Verified

- Build: `go build ./...` clean
- Tests: `go test -race -timeout 5m ./...` passes (10 packages, no race conditions detected)
- Vet: `go vet ./...` clean
- Branch rebased onto latest main (`5f03922`)
- Implementation correctness: sticky-zone behavior is correct — when `rateLimitLow == true` and ratio is 20%–50%, the actual ratio (e.g., 0.35) is passed to `computeEffectiveInterval`, which maps to 2× (the `>= 0.10` case). The sentinel design (`1.0` = not in backoff) correctly decouples activation from the `rateLimitBackoffThreshold` constant.
- The cap guard (`rateLimitInterval > maxRL`) is technically redundant but defensive; leaving it is the right call.

Reviewed implementation of stepwise rate-limit backoff escalation (issue 487). Fixed four boundary-ambiguity findings from the Copilot review: added exact threshold test cases (0.10, 0.05, 0.01) to pin tier assignments, and clarified dash-range notation to explicit `>=`/`<` labels in the function docstring, USER_GUIDE.md table, and state-machine.md. All Copilot threads resolved. Build, vet, and race-detector test suite pass.

---
Used 44/50 turns, 9k input / 17k output tokens.

verveguy:
*branch: fabrik/issue-487 | commit: 27a75d3b | main: 5f03922a808e9b34f584e1976d0b994ca8e5c3ac | 2026-05-03 03:26 UTC*

All checks pass. Branch is clean, up to date, and pushed. Here is the full validation report.

---

## Validation Report

### Requirements: 7/7 passed

- [x] **Effective rate-limit interval escalates above 2× as quota depletes**: Implemented via a 4-tier stepwise switch in `computeEffectiveInterval` (`engine/poll.go:109–125`). Verified: tiers 2×/4×/6×/10× match the plan table.
- [x] **At or near 0% budget, interval equals 10× cap**: `default: rlMult = rateLimitMaxBackoffMultiplier` (= 10). Verified by `TestComputeEffectiveInterval_RateLimitStepwise` (`ratio=0.0`, `ratio=0.009`) and `TestComputeEffectiveInterval_RateLimitExceeds5Min` (`ratio=0.005`).
- [x] **Hysteresis behaviour unchanged (activate <20%, clear >50%)**: `nextRateLimitLow` is unmodified. Sticky zone (20–50%) correctly maps to 2× tier via the `>= 0.10` case. Verified by test case `{"sticky zone (0.35)", 0.35, 60 * time.Second}`.
- [x] **`computeEffectiveInterval` signature accepts ratio float64 not bool**: Signature is `computeEffectiveInterval(configuredInterval, idleDuration time.Duration, rateLimitRatio float64, webhookHealthy bool)`. Call site at `poll.go:370` passes `rateLimitRatio` closure variable.
- [x] **`rateLimitRatio` closure variable tracks actual ratio**: Added alongside `rateLimitLow` at line 324; updated when `rateLimitLow` is true (passes actual ratio), reset to `1.0` otherwise. Sentinel `1.0` means "not in backoff".
- [x] **`engine/poll_test.go` updated**: All 4 original test functions rewritten; new `TestComputeEffectiveInterval_RateLimitStepwise` added covering all 4 tiers plus boundary values (0.10, 0.05, 0.01) and zero-remaining. Previous flat-2× assertions are gone.
- [x] **`docs/USER_GUIDE.md` rate-limit backoff section updated**: 4-row escalation table with explicit `>=`/`<` boundary notation and example intervals at 30s and 60s base. Hysteresis activation/clearance fully documented in prose below the table.

Additional (required by CLAUDE.md as-built spec maintenance):
- [x] **`docs/state-machine.md` §7.7 updated**: Expanded from "doubles the configured interval" to a structured sub-bullet list covering activation threshold, clearance threshold, sticky zone semantics, stepwise escalation tiers, no idle cap, and independence from idle backoff.

### Test Suite: PASSED

- 10 packages; race detector clean
- `engine` package: 125s (integration tests with real git)
- Vet: clean (`go vet ./...`)

No files deleted that were on main. No shared interfaces or types broken — the only function signature change was to an internal function with a single production call site.

### Verdict: READY TO MERGE

The stepwise rate-limit backoff escalation is implemented and verified. `computeEffectiveInterval` now accepts `rateLimitRatio float64` instead of a bool and scales the multiplier through four tiers (2×/4×/6×/10×) as quota depletes, bringing the 10× cap from dead code to active behaviour. All tests pass (race detector clean), docs in `USER_GUIDE.md` and `state-machine.md` are updated to reflect the actual escalation curve and hysteresis rules, and the branch is clean and pushed.

---
Used 20/50 turns, 10k input / 4k output tokens.

Issue #544, opened 2026-05-05
Self-advance after worker exit: race + cooldown amplifier delays next-stage dispatch up to 2.5 min


verveguy:
## Summary

When a worker completes one stage and the engine advances the item to the next stage on the same project board, the next-stage dispatch is racy:

- If the previous worker's goroutine has exited `e.inFlight` before the next dispatch loop runs → dispatch fires immediately.
- If the previous worker is still in flight (post-completion cleanup: posting comments, flipping labels, pushing the branch, network jitter) → dispatch is skipped, the item gets stamped with `CooldownAt["periodic-re-eval"]`, and the next dispatch waits ~2.5 minutes (`PollSeconds * 10 = 150s`).

The user-visible symptom is wildly inconsistent: same code path, same kind of stage transition, different latency by 2.5 minutes depending on a microseconds-wide race window.

## Live evidence

Same fabrik run, issue #540, two consecutive transitions:

**Research → Plan (lost the race, 2.5 min delay):**
```
13:38:30 [#540 claude] completed in 40 turns
13:38:36 [#540 advance] moving to stage "Plan"
13:38:36 [poll] deep-fetching details for #540
13:38:36 [poll] workers active                 ← dispatch skipped, worker still in inFlight
13:38:36 [poll] workers active                 ← second poll, same skip
13:38:51 [poll] nothing to do                  ← worker exited, but cooldown stamped
... (10 more idle polls) ...
13:41:08 [poll] deep-fetching details for #540 ← cooldown expired
13:41:14 [#540 claude] invoking (Plan)
```

**Plan → Implement (won the race, instant):**
```
13:44:44 [#540 claude] completed in 19 turns
13:44:48 [#540 advance] moving to stage "Implement"
13:44:49 [poll] deep-fetching details for #540
13:44:49 [#540 process] ... — stage: Implement ← dispatch succeeded immediately
13:44:55 [#540 claude] invoking (Implement)
```

The only material difference is timing: in the first case the dispatch loop ran before the Research worker's deferred `e.inFlight.Delete(iKey)` fired; in the second case the Plan worker's cleanup was fast enough that `inFlight` was clear by the time dispatch evaluated.

**Additional evidence — Review → Validate catch-up advance (issue #539):**

Same bug pattern, observed live on #539 (a yolo'd issue running the `wait_for_reviews` path):

```
15:33:45  [#539 done] comment processing complete       ← review-reinvoke worker exits
[...]

PR #570, opened 2026-05-06
Detect external project column moves within one poll cycle via cheap project.updatedAt gate

### LINKED ISSUE #565: Detect external project column moves within one poll cycle via cheap project.updatedAt gate
## Problem

External status changes — a user manually dragging an issue to a different column in the GitHub project board UI, another Fabrik instance advancing an issue, or any automation moving items — are currently invisible to Fabrik for up to **10 minutes**. This is the current Layer 2 sweep cadence (`ProjectStatusPollSeconds`, default 600 s). Fabrik's own mutations write through immediately (Layer 0, added in #544), but external moves wait for the next full status-batch poll.

The UX gap is pronounced: a user drags an issue to "Specify" and watches Fabrik sit idle for several minutes before it notices.

## Summary

Add a one-field `project.updatedAt` GraphQL gate that Fabrik checks at the top of every poll cycle (~15 s). When the timestamp hasn't changed, nothing else happens (cost: ~1 GraphQL point). When it has, Fabrik runs the existing `FetchProjectItemStatusBatch` + `ApplyStatusBatch` diff path to identify and apply the changed items. This brings external-move detection latency from ≤10 min down to ≤15 s at negligible API cost.

The implementation uses the **"fold into main poll loop" variant**: the `project.updatedAt` check is added at the top of `Engine.poll()` rather than keeping `runReconciliationLoop` as a separate goroutine. This eliminates moving parts and matches the existing 15 s poll cadence without adding a second timer.

## Requirements

- **New client method** `FetchProjectUpdatedAt(projectID string) (time.Time, error)` added to:
  - `github/project.go` (implementation: one-field GraphQL query `node(id:$id){ ...on ProjectV2 { updatedAt } }`)
  - `engine/interfaces.go` (`GitHubClient` interface)
  - `engine/mocks_test.go` (`mockGitHubClient` stub)
  - `cmd/cmd_extra_test.go` (`testGitHubUpgradeClient` stub)

- **Gate logic in `Engine.poll()`**: at the top of each poll cycle, call `FetchProjectUpdatedAt`. Store the last-seen timestamp in a new `lastProjectUpdatedAt time.Time` field on the Engine struct. Skip `FetchProjectItemStatusBatch` + `ApplyStatusBatch` when the timestamp is unchanged. Run them when it advances.
  - Gate is skipped (batch runs unconditionally) when `cache.ProjectID() == ""` (bootstrap not yet complete).
  - Gate is only active when `cacheImpl != nil` (webhooks + in-memory cache mode).

- **`runReconciliationLoop` removal or bypass**: the separate goroutine is replaced by the in-poll-loop check. The ticker/goroutine pattern is dropped entirely for the periodic status path.

- **`ProjectStatusPollSeconds` default updated** to 15 in `cmd/root.go`'s `statusPollSeconds()` (from 600). The config field and `--status-poll` / `FABRIK_STATUS_POLL` flags remain for user override.

- **`FetchProjectItemStatusBatch` unchanged**: no modifications to the existing batch query function.

- **Tests**:
  - Gate receives same timestamp twice → `FetchProjectItemStatusBatch` is NOT called the second time.
  - Gate receives new timestamp → `FetchProjectItemStatusBatch` IS called and `ApplyStatusBatch` applies drifted Status; unchanged items produce no observer notifications.
  - Both tests in `engine/poll_test.go` using the mock client pattern.

- **`go test -race ./...` passes**.

- **`docs/state-machine.md` updates** (in the same PR):
  - §D.4 (Periodic status-only sweep / Layer 2): update the cadence note; replace the `runReconciliationLoop` goroutine description with the in-poll-loop gate pattern; explain the two-step flow (gate answers "did anything change?", batch answers "which items changed?"); note the GitHub API limitation (no `since` filter / no `orderBy: UPDATED_AT` on `ProjectV2.items`, so the full batch is unavoidable).
  - §D.7 (Status field reconciliation in user mode): update the Layer 2 row in the table (cadence 10 min → 15 s) and the Residual latency note.

## Scope

**In scope:**
- `FetchProjectUpdatedAt` client method + interface + mocks
- `lastProjectUpdatedAt` field on Engine and gate logic in `poll()`
- Removal of `runReconciliationLoop` goroutine
- `ProjectStatusPollSeconds` default change from 600 to 15
- Tests for gate-hit and gate-miss paths
- `docs/state-machine.md` updates for §D.4 and §D.7

**Out of scope:**
- Per-item `updatedAt` filtering inside `FetchProjectItemStatusBatch` (mentioned in the original issue as an optional cleanliness improvement — not needed; the Store's existing diff already de-dupes unchanged-Status mutations)
- Restructuring webhook event handling or org-level `projects_v2_item` event delivery
- Any changes to Layer 0, Layer 1, or the stream-recovery reconciliation path (`reconcileCache`)

## Risks / Dependencies

- **Interface change**: adding `FetchProjectUpdatedAt` to `GitHubClient` requires updating all mock implementations. Failing to do so will cause compile errors — a safe failure mode.
- **No `since` filter on GitHub API**: the full batch (`FetchProjectItemStatusBatch`) is the only way to identify which items changed once the gate fires. This is a permanent API limitation unless GitHub adds a `since` param to `ProjectV2.items`. Cost is ~1-2 GraphQL points per active cycle.
- **`ProjectStatusPollSeconds` override**: users with custom `--status-poll` values (e.g. lower than 15 or much higher) are unaffected by the default change; the gate makes higher values safe and lower values cheaper.
- **First poll cycle**: on engine startup, `lastProjectUpdatedAt` is zero; the gate will fire the batch on the first cycle (timestamp advanced from zero to real value). This is correct and expected — it catches any drift that occurred before engine start.
- **Config field renamed**: the issue mentions `config/config.go` as a possible change location, but the default is actually in `cmd/root.go`'s `statusPollSeconds()`. No `config/config.go` change is needed unless the YAML config field name or parsing changes (it does not).

## Files to Change

- `github/project.go` — new `FetchProjectUpdatedAt` method
- `engine/interfaces.go` — add `FetchProjectUpdatedAt` to `GitHubClient` interface
- `engine/mocks_test.go` — add stub for `FetchProjectUpdatedAt` on `mockGitHubClient`
- `cmd/cmd_extra_test.go` — add stub for `FetchProjectUpdatedAt` on `testGitHubUpgradeClient`
- `engine/poll.go` — add `lastProjectUpdatedAt` field to Engine; add gate at top of `poll()`; remove `runReconciliationLoop` goroutine
- `cmd/root.go` — change `statusPollSeconds()` default from 600 to 15
- `engine/poll_test.go` — new gate-hit and gate-miss test cases
- `docs/state-machine.md` — update §D.4 and §D.7

Estimated scope: ~80–120 lines of code + tests + docs.

## Architectural Connection

This closes the last UX gap in per-repo-webhook mode. With the gate, Fabrik in per-repo mode achieves the same external-move detection latency as org-mode (~15 s) for a negligible API cost increase (~240 idle pts/hr vs ~24 pts/hr today, well within the 5000/hr budget).

verveguy:
## Summary

Add a one-field `project.updatedAt` GraphQL gate that Fabrik checks at the top of every poll cycle (~15 s). When the timestamp hasn't changed, nothing else happens (cost: ~1 GraphQL point). When it has, Fabrik runs the existing `FetchProjectItemStatusBatch` + `ApplyStatusBatch` diff path to identify and apply the changed items. This brings external-move detection latency from ≤10 min down to ≤15 s at negligible API cost.

The implementation uses the **"fold into main poll loop" variant**: the `project.updatedAt` check is added at the top of `Engine.poll()` rather than keeping `runReconciliationLoop` as a separate goroutine. This eliminates moving parts and matches the existing 15 s poll cadence without adding a second timer.

## Problem

External status changes — a user manually dragging an issue to a different column in the GitHub project board UI, another Fabrik instance advancing an issue, or any automation moving items — are currently invisible to Fabrik for up to **10 minutes**. This is the current Layer 2 sweep cadence (`ProjectStatusPollSeconds`, default 600 s). Fabrik's own mutations write through immediately (Layer 0, added in #544), but external moves wait for the next full status-batch poll.

The UX gap is pronounced: a user drags an issue to "Specify" and watches Fabrik sit idle for several minutes before it notices.

### Approach

Fold the Layer 2 status sweep directly into the main `poll()` loop using a one-field `project.updatedAt` GraphQL gate. At the top of each `poll()` call, when in cache mode with a bootstrapped project, the engine checks whether the project's `updatedAt` timestamp has advanced since the last check. If it hasn't, the batch is skipped (cost: ~1 GraphQL point). If it has, `FetchProjectItemStatusBatch` + `ApplyStatusBatch` run to identify and apply changed statuses.

This eliminates the `runReconciliationLoop` goroutine entirely — one fewer timer, one fewer goroutine, and no concurrent writer for the status batch path. The gate runs on every poll cycle (every ~15 s), so `ProjectStatusPollSeconds` effectively becomes an unused knob post-implementation (see Key Decisions).

Call order within the modified `poll()`:

```
poll() called
  → [Gate] if cacheImpl != nil && cacheImpl.ProjectID() != "":
      → e.client.FetchProjectUpdatedAt(projectID)
      → if timestamp advanced from lastProjectUpdatedAt:
          → e.client.FetchProjectItemStatusBatch(projectID)
          → cacheImpl.ApplyStatusBatch(updates)
          → e.lastProjectUpdatedAt = newTimestamp
  → e.readClient.FetchProjectBoard(...)    [cache now holds fresh Status]
  → ... rest of poll unchanged
```

### New/Modified Files

| File | Change |
|------|--------|
| `github/project.go` | Add `FetchProjectUpdatedAt(projectID string) (time.Time, error)` method |
| `engine/interfaces.go` | Append `FetchProjectUpdatedAt` to `GitHubClient` interface |
| `engine/mocks_test.go` | Add `fetchProjectUpdatedAtFn` field + method + `fetchProjectItemStatusBatchCalls` counter to `mockGitHubClient` |
| `cmd/cmd_extra_test.go` | Add trivial stub `FetchProjectUpdatedAt` to `testGitHubUpgradeClient` |
| `engine/engine.go` | Add `lastProjectUpdatedAt time.Time` field to `Engine` struct |
| `engine/poll.go` | Add gate block at top of `poll()`; remove `runReconciliationLoop` goroutine launch from `Run()`; remove `runReconciliationLoop` function |
| `cmd/root.go` | Change `statusPollSeconds()` default from 600 → 15; update inline comment and flag help text |
| `engine/poll_test.go` | Add `TestPollGateMiss` and `TestPollGateFire` |
| `docs/state-machine.md` | Rewrite §D.4 Layer 2 description; update §D.7 table row and residual latency note |
| `adrs/035-four-layer-status-reconciliation.md` | Add implementation-evolution note |

### Key Decisions

- **`lastProjectUpdatedAt` on Engine struct, not poll-local**: It must survive across poll cycles. Precedent: `idleStart time.Time` already lives on the Engine struct for the same reason. No mutex needed — `poll()` is single-goroutine, and removing `runReconciliationLoop` eliminates the only other writer.

- **Gate uses `e.client`, not `e.readClient`**: `FetchProjectUpdatedAt` is not part of `boardcache.ReadClient` (and should not be — it's a real-API call, not a cache read). Consistent with how `runReconciliationLoop` called `e.client.FetchProjectItemStatusBatch`.

- **Gate fires unconditionally on first poll cycle**: `lastProjectUpdatedAt` starts as `time.Time{}` (zero). Any real `updatedAt` from GitHub is after the zero time, so the batch runs on cycle 1. This is correct — it catches drift that occurred before engine start.

- **`ProjectStatusPollSeconds` becomes an unused knob**: After removing `runReconciliationLoop`, the `--status-poll` flag and `cfg.ProjectStatusPollSeconds` field no longer control anything at runtime (the gate fires every poll cycle regardless). The spec says to keep the field and flag, and change the default to 15 — this is done. The field value is simply not consumed in the new code path. No deprecation notice needed; the spec accepted this ambiguity.

- **No ADR needed for this change**: The four-layer architecture is preserved (ADR 035's decision stands). The mechanism change (goroutine → in-poll gate) is an implementation evolution of ADR 035's Layer 2, not a new architectural decision. A short evolution note is added to ADR 035 instead.

- **Call tracker for `FetchProjectItemStatusBatch` in mock**: The gate tests assert whether the batch was or was not called. Rather than adding a generic counter field for every method, add a specific `fetchProjectItemStatusBatchCalls int` counter (accessed under `mu`) to `mockGitHubClient`. Same pattern as `lookupIssueProjectItemCalls`.

### Task Checklist

- [ ] Task 1: Add `FetchProjectUpdatedAt` to `github/project.go` — one-field GraphQL query `node(id:$id){ ...on ProjectV2 { updatedAt } }` using the existing `graphqlRequest` + `parseTime` pattern from `FetchProjectItemStatus`
- [ ] Task 2: Add `FetchProjectUpdatedAt(projectID string) (time.Time, error)` to `GitHubClient` interface in `engine/interfaces.go`
- [ ] Task 3: Add stub to `engine/mocks_test.go` — `fetchProjectUpdatedAtFn func(projectID string) (time.Time, error)` field on `mockGitHubClient`, corresponding method body (mutex + fn call pattern), and `fetchProjectItemStatusBatchCalls int` counter incremented in the existing `FetchProjectItemStatusBatch` method body
- [ ] Task 4: Add trivial stub to `cmd/cmd_extra_test.go` — `func (m *testGitHubUpgradeClient) FetchProjectUpdatedAt(projectID string) (time.Time, error) { return time.Time{}, nil }`
- [ ] Task 5: Add `lastProjectUpdatedAt time.Time` field to `Engine` struct in `engine/engine.go`, adjacent to `idleStart time.Time`
- [ ] Task 6: Modify `engine/poll.go` — (a) add gate block at top of `poll()` before `FetchProjectBoard` call; (b) remove the `runReconciliationLoop` goroutine launch block (lines 397–403 in `Run()`); (c) delete the `runReconciliationLoop` function body
- [ ] Task 7: Update `cmd/root.go` — change `statusPollSeconds()` default return from `600` to `15`; update the function comment; update flag help text at line ~135 from "600" to "15"
- [ ] Task 8: Add `TestPollGateMiss` and `TestPollGateFire` to `engine/poll_test.go` using `testEngineWithCache` — gate-miss asserts `fetchProjectItemStatusBatchCalls == 1` after two polls with identical timestamps; gate-fire asserts `fetchProjectItemStatusBatchCalls == 2` after two polls with advancing timestamps
- [ ] Task 9: Update `docs/state-machine.md` — rewrite §D.4 "Periodic status-only sweep (Layer 2)" to describe the in-poll gate pattern; update §D.7 table Layer 2 row (mechanism, cadence 10 min → 15 s); update §D.7 residual latency note
- [ ] Task 10: Add evolution note to `adrs/035-four-layer-status-reconciliation.md` — brief paragraph noting that the Layer 2 mechanism was refactored from `runReconciliationLoop` goroutine to an in-poll `updatedAt` gate in issue #565, and that `state-machine.md` §D.4 is the authoritative as-built description
- [ ] Task 11: Run `go test -race ./...` and `go vet ./...`; fix any failures

### Risks

- **Interface-change compile cascade**: Adding `FetchProjectUpdatedAt` to `GitHubClient` breaks compilation until all three implementors (`*github.Client`, `mockGitHubClient`, `testGitHubUpgradeClient`) are updated. Tasks 1–4 must all complete before any `go build` or `go test` attempt; if implementing incrementally, add a dummy stub early to stay in a compilable state.

- **Gate test setup complexity**: `poll()` does many things beyond the gate — board fetch, rate limit logging, item dispatch. Gate tests need a mock client that returns benign non-error responses for all other calls (empty board, no items). `testEngineWithCache` already provides reasonable defaults; the implementer should verify the helper returns a valid `ProjectBoard` with `ProjectID: "PVT_1"` set so the gate's `cacheImpl.ProjectID() != ""` guard is satisfied.

- **Gate placement ordering**: The gate block must come before `e.readClient.FetchProjectBoard(...)` so that `ApplyStatusBatch` runs before the board is read in this cycle. If placed after, the fresh status won't be visible to item processing in the same cycle. Verify placement is before line 689 of the current `poll.go`.

- **`ProjectStatusPollSeconds` field residual confusion**: After the change, the Config field exists but has no runtime effect. Future contributors reading the config struct may wonder why it exists. The doc updates in §D.4/§D.7 will note the gate runs on every poll cycle; this is sufficient without adding a code comment to the struct.

---
Used 15/50 turns, 5k input / 7k output tokens.

## Verification

Implemented the `project.updatedAt` gate folded into `Engine.poll()`, replacing the `runReconciliationLoop` goroutine. Added `FetchProjectUpdatedAt` to `github/project.go`, `GitHubClient` interface, and all mock stubs; added `lastProjectUpdatedAt` to the Engine struct; wired the two-step gate (FetchProjectUpdatedAt → FetchProjectItemStatusBatch when timestamp advances) at the top of `poll()` before the board read. Changed `statusPollSeconds()` default from 600 to 15, replaced old reconciliation loop tests with `TestPollGateMiss` and `TestPollGateFire`, and updated `docs/state-machine.md` §D.4/§D.7 and ADR 035. All tests pass with `go test -race ./...`.

---

Closes #565

### DIFF
--- a/adrs/035-four-layer-status-reconciliation.md
+++ b/adrs/035-four-layer-status-reconciliation.md
@@ -96,3 +96,7 @@ The `ReadClient` interface (used by `GitHubAdapter` and `boardcache` tests) is n
 **Include PR events in Layer 1**: Rejected. Mapping a PR number to a cache key requires an O(N) scan of all cached items under read lock. The marginal coverage improvement (catching column moves that coincide only with PR events, not issue events) doesn't justify the complexity and latency. Layer 2's 10-minute sweep covers the gap.
 
 **Export `ItemKey` vs. inline format string**: Exporting chosen. The format `repo + "#" + strconv.Itoa(number)` is simple but must be consistent across packages. An exported helper prevents a silent divergence if the format ever changes and signals to callers that the format is stable and intentional.
+
+## Implementation Evolution
+
+**Issue #565 (2026-05-05)**: The Layer 2 mechanism was refactored from the `runReconciliationLoop` goroutine (a separate ticker at `ProjectStatusPollSeconds` cadence) to an in-poll `updatedAt` gate at the top of `Engine.poll()`. The gate calls `FetchProjectUpdatedAt` (~1 GraphQL point) every poll cycle (~15 s) and only fires `FetchProjectItemStatusBatch` + `ApplyStatusBatch` when the project's `updatedAt` timestamp has advanced. This eliminates the separate goroutine and reduces external-move detection latency from ≤10 min to ≤15 s. The four-layer architecture and the decision rationale above remain valid; only the Layer 2 mechanism changed. See `docs/state-machine.md §D.4` for the authoritative as-built description.
--- a/cmd/cmd_extra_test.go
+++ b/cmd/cmd_extra_test.go
@@ -127,6 +127,9 @@ func (m *testGitHubUpgradeClient) FetchPRsForSHA(owner, repo, sha string) ([]int
 func (m *testGitHubUpgradeClient) FetchProjectItem(owner, repo string, issueNumber int) (*gh.ProjectItem, error) {
 	return nil, nil
 }
+func (m *testGitHubUpgradeClient) FetchProjectUpdatedAt(projectID string) (time.Time, error) {
+	return time.Time{}, nil
+}
 func (m *testGitHubUpgradeClient) LookupIssueProjectItem(projectID, repo string, issueNumber int) (string, string, error) {
 	return "", "", nil
 }
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -50,7 +50,7 @@ type Config struct {
 	WebhookPort       int
 	WebhookEvents     string // comma-separated; empty means default event set
 	BoardCacheMode       string // "in-memory" or "none"; empty = auto (in-memory when webhooks enabled)
-	StatusPollSeconds    int    // Layer 2 status-only sweep cadence in seconds; 0 = use default (600)
+	StatusPollSeconds    int    // Layer 2 status-only sweep cadence in seconds; 0 = use default (15)
 }
 
 func Execute() error {
@@ -407,11 +407,11 @@ func Execute() error {
 			if n, err := strconv.Atoi(v); err == nil && n > 0 {
 				cfg.StatusPollSeconds = n
 			} else {
-				fmt.Fprintf(os.Stderr, "[warn] FABRIK_STATUS_POLL=%q is invalid (must be a positive integer); using default 600\n", v)
+				fmt.Fprintf(os.Stderr, "[warn] FABRIK_STATUS_POLL=%q is invalid (must be a positive integer); using default 15\n", v)
 			}
 		} else if pc.StatusPoll != nil {
 			if *pc.StatusPoll <= 0 {
-				fmt.Fprintf(os.Stderr, "[warn] config.yaml status_poll=%d is invalid (must be a positive integer); using default 600\n", *pc.StatusPoll)
+				fmt.Fprintf(os.Stderr, "[warn] config.yaml status_poll=%d is invalid (must be a positive integer); using default 15\n", *pc.StatusPoll)
 			} else {
 				cfg.StatusPollSeconds = *pc.StatusPoll
 			}
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -132,7 +132,7 @@ func Execute() error {
 	flag.IntVar(&cfg.WebhookPort, "webhook-port", 0, "Local port for the webhook HTTP listener (0 = OS-assigned; also FABRIK_WEBHOOK_PORT)")
 	flag.StringVar(&cfg.WebhookEvents, "webhook-events", "", "Comma-separated list of GitHub event types to subscribe to (default: all supported events; also FABRIK_WEBHOOK_EVENTS)")
 	flag.StringVar(&cfg.BoardCacheMode, "board-cache", "", `Board cache mode: "in-memory" (cache board state; requires --webhooks) or "none" (always fetch from GitHub). Default: "in-memory" when --webhooks is enabled, "none" otherwise. Also FABRIK_BOARD_CACHE.`)
-	flag.IntVar(&cfg.StatusPollSeconds, "status-poll", 0, "Cadence in seconds for the periodic status-only board sweep (Layer 2; 0 = use default of 600; also FABRIK_STATUS_POLL)")
+	flag.IntVar(&cfg.StatusPollSeconds, "status-poll", 0, "Cadence in seconds for the periodic status-only board sweep (Layer 2; 0 = use default of 15; also FABRIK_STATUS_POLL)")
 
 	if err := flag.CommandLine.Parse(os.Args[1:]); err != nil {
 		return err
@@ -619,10 +619,11 @@ func claudeWaitDelay(seconds int) time.Duration {
 }
 
 // statusPollSeconds returns the configured ProjectStatusPollSeconds, defaulting
-// to 600 (10 min) when n is 0 (unset).
+// to 15 s when n is 0 (unset). The gate now runs every poll cycle so the default
+// matches the main poll cadence rather than the former goroutine ticker period.
 func statusPollSeconds(n int) int {
 	if n <= 0 {
-		return 600
+		return 15
 	}
 	return n
 }
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -132,7 +132,7 @@ func Execute() error {
 	flag.IntVar(&cfg.WebhookPort, "webhook-port", 0, "Local port for the webhook HTTP listener (0 = OS-assigned; also FABRIK_WEBHOOK_PORT)")
 	flag.StringVar(&cfg.WebhookEvents, "webhook-events", "", "Comma-separated list of GitHub event types to subscribe to (default: all supported events; also FABRIK_WEBHOOK_EVENTS)")
 	flag.StringVar(&cfg.BoardCacheMode, "board-cache", "", `Board cache mode: "in-memory" (cache board state; requires --webhooks) or "none" (always fetch from GitHub). Default: "in-memory" when --webhooks is enabled, "none" otherwise. Also FABRIK_BOARD_CACHE.`)
-	flag.IntVar(&cfg.StatusPollSeconds, "status-poll", 0, "Cadence in seconds for the periodic status-only board sweep (Layer 2; 0 = use default of 15; also FABRIK_STATUS_POLL)")
+	flag.IntVar(&cfg.StatusPollSeconds, "status-poll", 0, "Retained for config compatibility; the Layer 2 updatedAt gate now runs every poll cycle (~15 s) regardless of this value. Also FABRIK_STATUS_POLL.")
 
 	if err := flag.CommandLine.Parse(os.Args[1:]); err != nil {
 		return err
--- a/docs/state-machine.md
+++ b/docs/state-machine.md
@@ -1686,16 +1686,26 @@ Reconciliation operates at two levels:
 
 **Periodic status-only sweep (Layer 2)**
 
-A `runReconciliationLoop` goroutine ticks at `ProjectStatusPollSeconds` (default 600 s, configurable via `--status-poll` / `FABRIK_STATUS_POLL` / `config.yaml status_poll`). Each tick calls `e.client.FetchProjectItemStatusBatch(projectID)` — a lightweight GraphQL query returning only `itemNodeID → statusName` for every board item — and passes the result to `cacheImpl.ApplyStatusBatch(updates)`.
+At the top of every `poll()` cycle (~15 s), when the in-memory cache is bootstrapped (`cacheImpl != nil` and `cacheImpl.ProjectID() != ""`), the engine runs a two-step gate:
+
+1. **Gate step** — `e.client.FetchProjectUpdatedAt(projectID)`: fetches the single `updatedAt` field for the project node (`node(id:$id){ ...on ProjectV2 { updatedAt } }`). Cost: ~1 GraphQL point per poll cycle. If the returned timestamp equals `e.lastProjectUpdatedAt` (no project-level change since last cycle), the batch is skipped entirely.
+
+2. **Batch step** — fired only when the timestamp has advanced: `e.client.FetchProjectItemStatusBatch(projectID)` returns `itemNodeID → statusName` for every board item (paginated in pages of 100, no nested fields). The result is passed to `cacheImpl.ApplyStatusBatch(updates)`.
 
 `ApplyStatusBatch` holds the write lock for one pass:
 - For each `(itemID, newStatus)` pair, look up the cache key via `itemIDToKey`.
-- If the status differs from the cached value, update `Status` and `UpdatedAt`. Unchanged items are skipped.
+- If the status differs from the cached value, update `Status` and `UpdatedAt`. Unchanged items are skipped (no observer notifications).
 - Unknown item IDs (not yet bootstrapped) are silently skipped.
 
-This sweep is far cheaper than a full board fetch (no nested fields, no comments, no labels) and can run at 10-minute cadence without meaningful GraphQL budget pressure.
+After a successful batch step, `e.lastProjectUpdatedAt` is updated to the new timestamp so the next poll cycle can compare correctly.
 
-The loop skips a tick and logs a warning if `cacheImpl.ProjectID()` is empty, which means the bootstrap has not yet completed.
+**First poll cycle**: `e.lastProjectUpdatedAt` starts at the zero value (`time.Time{}`). The first real `updatedAt` returned by GitHub is always later than zero, so the batch fires on the first cycle. This catches any Status drift that occurred before engine start.
+
+**GitHub API limitation**: GitHub's `ProjectV2.items` does not support a `since` filter or `orderBy: UPDATED_AT`. The full batch (`FetchProjectItemStatusBatch`) is the only way to identify which specific items changed once the gate fires. The Store's existing diff in `ApplyStatusBatch` de-dupes unchanged-Status items so no spurious observer notifications are emitted.
+
+**Gate placement**: The gate runs before `e.readClient.FetchProjectBoard(...)` in `poll()`, so the cache holds the latest Status values when the board is read and items are dispatched in the same cycle.
+
+**Gate inactive when**: `e.readClient` is not a `*boardcache.CacheImpl` (non-cache mode), or `cacheImpl.ProjectID()` is empty (bootstrap not yet complete).
 
 **Full reconcile — stream-recovery only**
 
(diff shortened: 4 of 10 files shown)


verveguy:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-565 | commit: 832bb3d6 | main: 8ac2538727f13da94f10b55c1e64135e92ece1fb | 2026-05-06 01:41 UTC*

## Review Findings

### Fixed

- **Issue [1]**: Four stale `600` references survived the default change from 600→15: two user-visible warning messages (`FABRIK_STATUS_POLL=%q is invalid … using default 600` and `config.yaml status_poll=%d … using default 600`) plus two struct field comments (`StatusPollSeconds` in `cmd/root.go` and `ProjectStatusPollSeconds` in `engine/engine.go`). **Fix**: Updated all four to reference `15`.

- **Issue [2]**: Double blank line left behind in `engine/poll.go` at the deletion site of `runReconciliationLoop`. **Fix**: Removed extra blank line.

### Verified

- `go build ./...` and `go vet ./...` clean after all changes.
- `go test -race -timeout 5m ./...` — all 11 packages pass (11/11 `ok`).
- Branch is already rebased onto `origin/main`.
- No PR comments from review bots to address.

### Implementation review

- **Gate logic**: `updatedAt.After(lastProjectUpdatedAt)` with strict ordering is correct — zero value on startup always triggers first cycle; same timestamp correctly skips batch.
- **Error handling**: `FetchProjectUpdatedAt` failure is logged and skipped (batch doesn't fire), matching the engine's non-fatal logging pattern.
- **Concurrency**: `lastProjectUpdatedAt` lives on the Engine struct; `poll()` is single-goroutine and `runReconciliationLoop` was the only other writer — now removed. No mutex needed. Race detector confirms no races.
- **Gate placement**: Runs before `FetchProjectBoard`, so cache holds the freshest Status when items are dispatched in the same cycle.
- **Interface coverage**: All three `GitHubClient` implementors updated (`*github.Client`, `mockGitHubClient`, `testGitHubUpgradeClient`).
- **Tests**: `TestPollGateMiss` (1 batch call across 2 polls when timestamp is static) and `TestPollGateFire` (2 batch calls when timestamp advances) correctly exercise the gate invariants.
- **Docs**: `state-machine.md` §D.4 rewrites the goroutine description to the in-poll gate pattern; §D.7 table and residual latency note updated from 10 min to ~15 s. ADR 035 has an evolution note.

### Blocking

None.

---

Reviewed the `updatedAt` gate implementation for Layer 2 status sweep. Fixed 4 stale "600 s" references in warning messages and struct comments that weren't updated when the default changed to 15 s, and removed a double blank line from poll.go. All 11 packages pass `go test -race ./...`. PR is ready for human review.

---
Used 43/50 turns, 9k input / 13k output tokens.

verveguy:
*branch: fabrik/issue-565 | commit: 475e440d | main: 8ac2538727f13da94f10b55c1e64135e92ece1fb | 2026-05-06 01:50 UTC*

All four threads resolved. Here's a summary of actions taken:

## Copilot Review Findings — All Fixed

**Thread [1] — `engine/poll.go`: Gate error silently disabling Layer 2**
Fixed. When `FetchProjectUpdatedAt` fails, the gate now falls back to running `FetchProjectItemStatusBatch` unconditionally, logging the error and noting the fallback. `lastProjectUpdatedAt` is only updated when the gate call succeeds (has a real timestamp), so the next cycle retries the gate cleanly.

**Thread [2] — `github/project.go`: Non-ProjectV2 node gives misleading parse error**
Fixed. Added an explicit check for `UpdatedAt == ""` before calling `parseTime`. When the `ProjectV2` fragment doesn't match (node is a different type), the field is absent and returns empty string. Now returns a clear `"node is not a ProjectV2 (updatedAt field absent)"` error instead of a confusing time-parse failure.

**Thread [3] — `cmd/root.go`: `--status-poll` flag help describes a now-unused cadence**
Fixed. Updated the flag help text to: `"Retained for config compatibility; the Layer 2 updatedAt gate now runs every poll cycle (~15 s) regardless of this value."` — accurately reflects that the value is no longer consumed at runtime.

**Thread [4] — `engine/poll_test.go`: Tests only check call counts, not behavioral contract**
Fixed. Added `TestPollGateFire_AppliesStatusDrift` which seeds the cache with one item in `"Research"`, configures `FetchProjectItemStatusBatch` to return `{"PVTI_001": "Implement"}`, calls `poll()` once, and asserts the cached board reflects `"Implement"` — verifying that `ApplyStatusBatch` actually updates cached Status when the gate fires.

All three engine tests pass (`TestPollGateMiss`, `TestPollGateFire`, `TestPollGateFire_AppliesStatusDrift`). Full suite passes. All four Copilot threads resolved.

---
**Threads addressed:**
- `engine/poll.go:697` — resolved
- `github/project.go:957` — resolved
- `cmd/root.go:135` — resolved
- `engine/poll_test.go:1474` — resolved

verveguy:
*branch: fabrik/issue-565 | commit: 475e440d | main: 8ac2538727f13da94f10b55c1e64135e92ece1fb | 2026-05-06 01:54 UTC*

All checks pass. Here's the complete validation report:

---

## Validation Report

### Requirements: 8/8 passed

- [x] **`FetchProjectUpdatedAt` in `github/project.go`**: Implemented at line 929 using the `node(id:$id){ ...on ProjectV2 { updatedAt } }` GraphQL pattern with explicit empty-string guard.
- [x] **`GitHubClient` interface updated**: `engine/interfaces.go:49` — method added.
- [x] **Mock stubs updated**: `engine/mocks_test.go:536` and `cmd/cmd_extra_test.go:130` — both present.
- [x] **`lastProjectUpdatedAt` on Engine struct**: `engine/engine.go:78` — adjacent to `idleStart`, with comment.
- [x] **Gate logic in `Engine.poll()`**: `engine/poll.go:688-699` — gate checks `cacheImpl != nil` + non-empty `ProjectID()`, falls back to batch on error, updates timestamp only on success.
- [x] **`runReconciliationLoop` removed**: No occurrences in `poll.go` — function and goroutine launch fully deleted.
- [x] **`statusPollSeconds()` default → 15**: `cmd/root.go:624` — returns 15 when `n <= 0`; flag help updated to reflect gate semantics.
- [x] **`docs/state-machine.md` §D.4 and §D.7 updated**: §D.4 describes the in-poll `updatedAt` gate at lines 1689+; §D.7 table row at line 1752 shows "In-poll `updatedAt` gate… ~15 s cadence"; residual latency note at line 1755 updated.

### Tests: PASSED

- 3 gate tests: `TestPollGateMiss`, `TestPollGateFire`, `TestPollGateFire_AppliesStatusDrift` — all in `engine/poll_test.go`
- Full suite: 12 packages, all pass
- Race detector: clean
- Vet: clean (`go vet ./...`)

All 12 packages pass with race detector enabled. The interface addition compiled cleanly across all three implementors.

### Verdict: READY TO MERGE

The implementation fully meets the spec: the `updatedAt` gate is folded into `poll()`, `runReconciliationLoop` is gone, the default status-poll cadence is 15 s, all mocks are updated, and the docs reflect the new behavior.

All requirements verified: `FetchProjectUpdatedAt` method implemented in `github/project.go`, wired into the `GitHubClient` interface and all mocks, `lastProjectUpdatedAt` field added to the Engine struct, gate logic active at the top of `poll()` with error fallback, `runReconciliationLoop` fully removed, `statusPollSeconds()` default changed to 15, and `docs/state-machine.md` §D.4/§D.7 updated. Full test suite (12 packages) passes clean with race detector.

---
Used 21/50 turns, 12k input / 3k output tokens.

PR #606, opened 2026-05-06
test: pipeline smoke test #4 — Issue H (blockedBy Issue G)

verveguy:
## Summary

Issue H is the "tail" of a two-issue dependency chain (G → H). It sits in `fabrik:blocked` state until Issue G (#602) closes. Once unblocked, it appends a single timestamped line to `tests/pipeline-runs.md`, then lets `fabrik:yolo` auto-merge the resulting PR. The only observable output is that one appended line; everything else is engine-behavior verification via the acceptance criteria.

## Problem

Fabrik's `fabrik:blocked` / dependency-resolution path went through several bug-fix cycles (#576, #578, #580, #587, #588, #589, #598, #599). This smoke test exercises the full end-to-end blocked→unblocked→complete chain with all those fixes deployed, specifically verifying:
- No TUI ghost during the blocked period
- No repeated `[blocked] waiting for …` log spam
- Push-unblock fires within seconds of the blocking issue closing
- The `fabrik:awaiting-review` gate clears in <30s after Copilot reviews (key #598 verification)

### Approach

Append the single required line verbatim to the end of `tests/pipeline-runs.md`. No design decisions are needed — the spec prescribes the exact content and the only file to touch.

### New/Modified Files

| File | Change |
|------|--------|
| `tests/pipeline-runs.md` | Append one line: `- 2026-05-06 Run H — smoke test #4, blockedBy chain tail (post-#598/#599 verification)` |

### Key Decisions

- **Single file append only**: The spec and acceptance criteria are explicit — no other files are modified. No documentation, engine code, or test infrastructure changes.
- **Exact line content**: The line is specified verbatim in the requirements; it must be reproduced character-for-character including the em dash (`—`) and version references.

No ADR is warranted — this is a smoke test file append with no architectural significance.

No documentation updates are needed — Research explicitly noted no documentation impact.

### Task Checklist

- [ ] Task 1: Append `- 2026-05-06 Run H — smoke test #4, blockedBy chain tail (post-#598/#599 verification)` to `tests/pipeline-runs.md`
- [ ] Task 2: Commit the change with message `test: record pipeline smoke test #4 Run H`
- [ ] Task 3: Push the branch

### Risks

- None. Single-line append to a markdown log; CI has no tests that would fail.

---
Used 6/50 turns, 2k input / 1k output tokens.

## Verification

[...]

PR #604, opened 2026-05-06
test: pipeline smoke test #4 — Issue G (no blocker)

verveguy:
## Summary

Append a single timestamped entry to `tests/pipeline-runs.md` to record smoke test #4, Issue G (the blocker in a blockedBy chain). Issue H is the companion that depends on this issue, validating the full blockedBy chain traversal.

## Problem

Pipeline smoke tests #1–#3 measured baseline behavior (awaiting-review stall latency, markPRReady failure recovery) before fixes #598 and #599 landed. This run captures a post-fix timing measurement to verify the improvements are working as expected in production.

### Approach

Single-file append. The spec provides the exact line to add verbatim; research confirmed the file exists and follows a simple bullet-list format. No design decisions required.

### New/Modified Files

| File | Change |
|------|--------|
| `tests/pipeline-runs.md` | Append one line: `- 2026-05-06 Run G — smoke test #4, blockedBy chain head (post-#598/#599 verification)` |

### Key Decisions

- **Append only**: The spec explicitly prohibits modifying any other files. No reformatting, no reordering of existing entries.
- **No ADR needed**: This is a test-record append with no architectural implications.
- **No documentation updates**: `tests/pipeline-runs.md` is a test record, not user-facing or as-built documentation. No other docs require changes.

### Task Checklist

- [ ] Task 1: Append `- 2026-05-06 Run G — smoke test #4, blockedBy chain head (post-#598/#599 verification)` to `tests/pipeline-runs.md`

### Risks

None. The file exists, the format is established, and the content is specified verbatim.

---
Used 5/50 turns, 1k input / 0k output tokens.

## Verification

Appended `- 2026-05-06 Run G — smoke test #4, blockedBy chain head (post-#598/#599 verification)` to `tests/pipeline-runs.md` as the only change. Committed and pushed to `fabrik/issue-602`.

---

Closes #602

Issue #603, opened 2026-05-06
test: pipeline smoke test #4 — Issue H (blockedBy Issue G)


verveguy:
## Problem

Fabrik's `fabrik:blocked` / dependency-resolution path went through several bug-fix cycles (#576, #578, #580, #587, #588, #589, #598, #599). This smoke test exercises the full end-to-end blocked→unblocked→complete chain with all those fixes deployed, specifically verifying:
- No TUI ghost during the blocked period
- No repeated `[blocked] waiting for …` log spam
- Push-unblock fires within seconds of the blocking issue closing
- The `fabrik:awaiting-review` gate clears in <30s after Copilot reviews (key #598 verification)

## Summary

Issue H is the "tail" of a two-issue dependency chain (G → H). It sits in `fabrik:blocked` state until Issue G (#602) closes. Once unblocked, it appends a single timestamped line to `tests/pipeline-runs.md`, then lets `fabrik:yolo` auto-merge the resulting PR. The only observable output is that one appended line; everything else is engine-behavior verification via the acceptance criteria.

## Requirements

- Do NOT proceed until Issue G (#602) has closed and the engine has cleared `fabrik:blocked` automatically.
- Do NOT remove `fabrik:blocked` manually.
- Append exactly this line to `tests/pipeline-runs.md`:
  ```
  - 2026-05-06 Run H — smoke test #4, blockedBy chain tail (post-#598/#599 verification)
  ```
- No other files are modified.
- CI passes; the linked PR auto-merges via `fabrik:yolo`.

## Scope

**In scope:**
- Waiting for Issue G to close
- Appending the single `Run H` line to `tests/pipeline-runs.md`
- Validating the acceptance criteria observations listed below

**Out of scope:**
- Any code changes to the Fabrik engine
- Any changes to test infrastructure beyond the single appended line

## Open Questions

*(none — spec is complete)*

## Prior Art / Context

Companion issue is #602 (Issue G, pipeline smoke test #4). This issue is blocked by #602 via the `addBlockedBy` GraphQL mutation. Relevant merged fixes: #576, #578, #580, #587, #588, #589, #598, #599.

## Risks / Dependencies

- **Blocked on #602**: No work can proceed until Issue G (#602) closes.
- If Issue G is itself stalled or paused, this issue will remain blocked indefinitely.

## Acceptance Criteria

- [ ] `fabrik:blocked` cleared automatically (not manually) when #602 closes.
[...]

Issue #602, opened 2026-05-06
test: pipeline smoke test #4 — Issue G (no blocker)


verveguy:
## Problem

Pipeline smoke tests #1–#3 measured baseline behavior (awaiting-review stall latency, markPRReady failure recovery) before fixes #598 and #599 landed. This run captures a post-fix timing measurement to verify the improvements are working as expected in production.

## Summary

Append a single timestamped entry to `tests/pipeline-runs.md` to record smoke test #4, Issue G (the blocker in a blockedBy chain). Issue H is the companion that depends on this issue, validating the full blockedBy chain traversal.

## Requirements

- `tests/pipeline-runs.md` must contain the exact line:
  ```
  - 2026-05-06 Run G — smoke test #4, blockedBy chain head (post-#598/#599 verification)
  ```
- No other files may be modified.
- CI must pass and PR auto-merges via `fabrik:yolo`.

## Scope

**In scope:**
- Appending one line to `tests/pipeline-runs.md`.

**Out of scope:**
- Any code changes.
- Analysis of the timing results (separate observation step).
- Issue H's corresponding entry (tracked separately in its own issue).

## Open Questions

None.

## Prior Art / Context

- Smoke tests #1–#3 established the baseline timing (~18–20 min wall-clock, ~10–12 min awaiting-review stall).
- #598 fixes the awaiting-review stall by synchronously clearing `LinkedPRReviewRequests` on Copilot review.
- #599 adds bounded retry to `markPRReady` for transient 5xx errors.
- Post-fix target: ~5–6 min wall-clock, <30s awaiting-review stall.

## Risks / Dependencies

- Issue H (`blockedBy` this issue) will not advance until this issue merges — the chain validates the dependency-unblocking logic.

verveguy:
🏭 **Fabrik — stage: Specify**
*branch: fabrik/issue-602 | commit: bdf7ba7e | main: bdf7ba7e172b30e80c6136b63ca8625140185e73 | 2026-05-06 22:46 UTC*

The spec is clear and complete with no ambiguities. This is a minimal, well-defined smoke test task with explicit acceptance criteria and no open questions.

---
Used 5/50 turns, 1k input / 0k output tokens.

verveguy:
🏭 **Fabrik — stage: Research**
*branch: fabrik/issue-602 | commit: bdf7ba7e | main: bdf7ba7e172b30e80c6136b63ca8625140185e73 | 2026-05-06 22:47 UTC*

[...]

PR #611, opened 2026-05-07
fabrik: Implement stage posts output to issue before draft PR is created

### LINKED ISSUE #608: fabrik: Implement stage posts output to issue before draft PR is created
## Problem

When the Implement stage completes, the engine sometimes posts output to the issue instead of the linked PR, despite the stage being configured with `post_to_pr: true`. The log shows the fallback message followed immediately by PR creation:

```
[#603 warn] no open PR found, posting on issue instead
[#603 worktree] pushed fabrik/issue-603 to origin
[#603 worktree] pushed fabrik/issue-603 to origin
[#603 pr] created draft PR #606
[#603 pr] updated PR #606 Verification section
[#603 pr] marked PR #606 ready-for-review
```

The Implement stage output (intended for the PR per `post_to_pr: true`) ends up on the issue instead. PR #606 is then created seconds later, leaving the issue cluttered with output that belongs on the PR.

Smoke test #4, issue #603 demonstrated this. Issue #602 (run earlier in the same smoke test) did NOT exhibit the bug — for #602, PR #604 was created before the post step ran. The behavior is order-sensitive but not yet deterministically reproducible outside that test.

## Summary

The engine's post-stage flow in `engine/item.go` posts Claude's output before creating the draft PR. When a stage is configured with both `post_to_pr: true` and `create_draft_pr: true` (as Implement is), the post step runs first, finds no open PR, and falls back to posting on the issue. The draft PR is created afterward. The fix is to reorder operations so the draft PR is created before output is posted when the stage completes.

## Root Cause (Confirmed)

In `engine/item.go`, the output post happens unconditionally at approximately line 815–835, before the `if completed { ... }` block at line 928–942 where `ensureDraftPR` is called. On a completing run with `create_draft_pr: true` and `post_to_pr: true`, the post always races ahead of the PR creation.

`postOutputToPR` (in `engine/pr.go:230`) searches for an existing open PR linked to the issue. When none exists, it logs the fallback warning and posts on the issue instead.

## Requirements

- When a stage completes with both `create_draft_pr: true` and `post_to_pr: true`, the draft PR must be created (via `ensureDraftPR`) **before** `postOutputToPR` is called.
- When a stage completes with `create_draft_pr: true` but not `post_to_pr: true`, the existing behavior (post first, create PR after) is acceptable since the output goes to the issue by design — no change needed for that path.
- The `updatePRVerification` call (which requires the PR number from `ensureDraftPR`) must remain associated with the same PR creation step, preserving its current semantics.
- The `markPRReady` call must still happen after PR creation, as it does today.
- Non-completing runs (max_turns exceeded, error, blocked) must not trigger early PR creation — `ensureDraftPR` must only be called when `completed` is true.
- Existing test coverage for the Implement post-stage flow should be updated or added to verify that the PR is created before output is posted.

## Scope

**In scope:**
- Reordering `ensureDraftPR` to run before `postOutputToPR` on completing runs where both `create_draft_pr` and `post_to_pr` are true.
- Updating or adding tests to catch this ordering regression.

**Out of scope:**
- Changing the fallback behavior of `postOutputToPR` when no PR is found on non-completing runs — that behavior is intentional for mid-stage partial output.
- Any changes to how `wait_for_ci` or `mark_pr_ready_on_complete` interact with the PR lifecycle.

## Open Questions

None.

## Risks / Dependencies

- The fix touches the post-stage sequencing in `engine/item.go`, which is a high-traffic code path. The change is a simple reorder, not a logic change, so risk is low.
- The `prNumber` variable is currently declared inside `if completed` after the post step; moving `ensureDraftPR` earlier requires restructuring that variable's scope.

verveguy:
## Summary

The engine's post-stage flow in `engine/item.go` posts Claude's output before creating the draft PR. When a stage is configured with both `post_to_pr: true` and `create_draft_pr: true` (as Implement is), the post step runs first, finds no open PR, and falls back to posting on the issue. The draft PR is created afterward. The fix is to reorder operations so the draft PR is created before output is posted when the stage completes.

## Problem

When the Implement stage completes, the engine sometimes posts output to the issue instead of the linked PR, despite the stage being configured with `post_to_pr: true`. The log shows the fallback message followed immediately by PR creation:

```
[#603 warn] no open PR found, posting on issue instead
[#603 worktree] pushed fabrik/issue-603 to origin
[#603 worktree] pushed fabrik/issue-603 to origin
[#603 pr] created draft PR #606
[#603 pr] updated PR #606 Verification section
[#603 pr] marked PR #606 ready-for-review
```

The Implement stage output (intended for the PR per `post_to_pr: true`) ends up on the issue instead. PR #606 is then created seconds later, leaving the issue cluttered with output that belongs on the PR.

Smoke test #4, issue #603 demonstrated this. Issue #602 (run earlier in the same smoke test) did NOT exhibit the bug — for #602, PR #604 was created before the post step ran. The behavior is order-sensitive but not yet deterministically reproducible outside that test.

### Approach

The fix is a minimal, targeted reorder within `processItem` in `engine/item.go`. No new types, interfaces, or files are needed.

**The change in three steps:**

1. **Lift `var prNumber int`** out of the `if completed` block and declare it before the post-output block, so it is in scope for both the early guard and the existing completion logic.

2. **Add an early `ensureDraftPR` guard** immediately before the post-output block:
   ```go
   if completed && stage.CreateDraftPR && stage.PostToPR {
       prNumber = e.ensureDraftPR(item, baseBranch)
   }
   ```
   This guarantees the PR exists before `postOutputToPR` calls `FindPRForIssue`. The guard is strictly inside a `completed` check, so non-completing runs (max_turns, error, blocked) are unaffected.

3. **Guard the `ensureDraftPR` call in `if completed`** with `prNumber == 0` so the call is skipped when the PR was already created by the early guard:
   ```go
   if stage.CreateDraftPR {
       if prNumber == 0 {
           prNumber = e.ensureDraftPR(item, baseBranch)
       }
       e.updatePRVerification(item, prNumber, extractSummary(output))
   }
   ```

The `updatePRVerification` and `markPRReady` calls are unchanged in position and semantics. The `!PostToPR` path (post goes to issue by design) is unaffected.

### New/Modified Files

| File | Change |
|------|--------|
| `engine/item.go` | Move `var prNumber int` before post-output block; add early `ensureDraftPR` guard; add `prNumber == 0` check in `if completed` |
| `engine/pr_test.go` | Add `TestProcessItem_PostToPR_CreatesDraftPRBeforePosting` |

### Key Decisions

- **Conditional early call, not unconditional move**: The early `ensureDraftPR` is guarded by `completed && stage.CreateDraftPR && stage.PostToPR`. If we moved it unconditionally (before the `completed` check), it would fire on non-completing runs, violating the spec requirement. The `PostToPR` condition is also required — for stages with `create_draft_pr: true` but `post_to_pr: false`, output goes to the issue by design and the existing order is correct.

- **`prNumber == 0` skip in `if completed`**: `ensureDraftPR` is idempotent, so calling it twice would be safe. The `== 0` guard avoids a redundant `FindPRForIssue` round-trip and makes the intent explicit: "only create if not already created."

- **No ADR warranted**: This is a bug fix to call ordering within a single function. No new abstractions, config options, or constraints are introduced. The fix is self-evident from reading the code.

- **Documentation impact**: None. This fixes internal engine sequencing. No new config options, labels, markers, or user-visible behaviors are introduced. `docs/state-machine.md` and `docs/stage-lifecycle.md` do not need updating — the fix brings the implementation in line with what a reader would already expect from those docs (PR exists before output is posted).

### Task Checklist

- [ ] Task 1: In `engine/item.go`, move `var prNumber int` to before the post-output block (~line 813) and add the early guard `if completed && stage.CreateDraftPR && stage.PostToPR { prNumber = e.ensureDraftPR(item, baseBranch) }` immediately before the `if postOutput != ""` block
- [ ] Task 2: In `engine/item.go`, remove the now-redundant `var prNumber int` from inside `if completed` (~line 934) and add `if prNumber == 0 {` guard around the existing `prNumber = e.ensureDraftPR(...)` call, keeping `e.updatePRVerification(...)` unconditional within `if stage.CreateDraftPR`
- [ ] Task 3: Add `TestProcessItem_PostToPR_CreatesDraftPRBeforePosting` in `engine/pr_test.go` — use a stateful mock where `FindPRForIssue` returns `0` on the first call (simulating no PR at post time) and `prNum` on subsequent calls (simulating PR created by `ensureDraftPR`); `createDraftPRFn` records it was called and returns `prNum`; verify that `createDraftPRFn` was called and `AddComment` was called with `prNum` (output posted to PR, not issue fallback)
- [ ] Task 4: Run `go test -race ./engine/...` and confirm all tests pass including the new one and the existing `TestProcessItem_ImplementStage_UpdatesVerificationOnComplete`

### Risks

- **`ensureDraftPR` pushes the branch internally** (`pr.go:38`): when called early, it pushes before the outer `wm.PushBranch` at `item.go:897`. The outer push then becomes a no-op (same HEAD). This is safe — identical to a double push of the same ref.
- **Existing test unaffected**: `TestProcessItem_ImplementStage_UpdatesVerificationOnComplete` has `PostToPR: false`, so the new early guard (`stage.PostToPR` is false) never fires for it. Its mock and assertions need no changes.
- **Race risk**: nil — `prNumber` is a local variable in `processItem`; no shared state is introduced.

---
Used 9/50 turns, 3k input / 7k output tokens.

## Verification

Fixed the ordering bug in `engine/item.go` where `postOutputToPR` ran before `ensureDraftPR` on completing `post_to_pr + create_draft_pr` stages, causing output to fall back to the issue. The fix lifts `var prNumber int` before the post-output block and adds an early `ensureDraftPR` guard so the PR always exists when `postOutputToPR` calls `FindPRForIssue`. Added `TestProcessItem_PostToPR_CreatesDraftPRBeforePosting` to catch this regression; all tests pass.

---

Closes #608

### DIFF
--- a/engine/item.go
+++ b/engine/item.go
@@ -811,6 +811,13 @@ func (e *Engine) processItem(ctx context.Context, board *gh.ProjectBoard, item g
 		postOutput = strings.TrimSpace(postOutput)
 	}
 
+	// When completing a stage that posts output to a PR and creates a draft PR,
+	// ensure the PR exists before posting so postOutputToPR can find it.
+	var prNumber int
+	if completed && stage.CreateDraftPR && stage.PostToPR {
+		prNumber = e.ensureDraftPR(item, baseBranch)
+	}
+
 	// Post Claude's output
 	if postOutput != "" {
 		footer := formatStatsFooter(usage, completed)
@@ -930,10 +937,12 @@ func (e *Engine) processItem(ctx context.Context, board *gh.ProjectBoard, item g
 		// Clear retry tracking for this stage — no longer needed after success.
 		e.store.Apply(itemstate.StageRetryCleared{Repo: repoStr, Number: item.Number, StageName: stage.Name})
 		e.store.Apply(itemstate.EngineUnpaused{Repo: repoStr, Number: item.Number, StageName: stage.Name})
-		// Post-stage: create draft PR and/or mark ready now that commits exist
-		var prNumber int
+		// Post-stage: create draft PR and/or mark ready now that commits exist.
+		// prNumber may already be set if the early guard above ran (PostToPR + CreateDraftPR path).
 		if stage.CreateDraftPR {
-			prNumber = e.ensureDraftPR(item, baseBranch)
+			if prNumber == 0 {
+				prNumber = e.ensureDraftPR(item, baseBranch)
+			}
 			e.updatePRVerification(item, prNumber, extractSummary(output))
 		}
 		if stage.MarkPRReadyOnComplete {
--- a/engine/pr_test.go
+++ b/engine/pr_test.go
@@ -858,6 +858,9 @@ func TestProcessItem_PostToPR_CreatesDraftPRBeforePosting(t *testing.T) {
 	// before the draft PR has been created — this is the bug: full stage output
 	// falls back to the issue instead of the PR.
 	var issueCommentBeforePR bool
+	// prCommentPosted is set when AddComment is called on the PR number,
+	// confirming output was actually routed to the PR.
+	var prCommentPosted bool
 
 	client := &mockGitHubClient{
 		findPRForIssueFn: func(owner, repo string, issueNumber int) (int, error) {
@@ -875,6 +878,9 @@ func TestProcessItem_PostToPR_CreatesDraftPRBeforePosting(t *testing.T) {
 			if issueNumber == issueNum && !prExists {
 				// AddComment on the issue before the PR existed — this is the fallback bug.
 				issueCommentBeforePR = true
+			}
+			if issueNumber == prNum {
+				prCommentPosted = true
 			}
 			return issueNumber*10 + 1, nil
 		},
@@ -947,6 +953,9 @@ func TestProcessItem_PostToPR_CreatesDraftPRBeforePosting(t *testing.T) {
 	// by the time any AddComment on the issue number fires (summary, not fallback).
 	if issueCommentBeforePR {
 		t.Error("AddComment called on issue before draft PR was created — output fell back to issue instead of PR")
+	}
+	if !prCommentPosted {
+		t.Error("expected AddComment to be called on the PR — output was not posted to PR")
 	}
 }
 
--- a/engine/pr_test.go
+++ b/engine/pr_test.go
@@ -821,6 +821,135 @@ func TestMarkPRReady_AllTransientExhausted(t *testing.T) {
 	}
 }
 
+// TestProcessItem_PostToPR_CreatesDraftPRBeforePosting verifies that when a stage
+// completes with both create_draft_pr and post_to_pr true, the draft PR is created
+// before postOutputToPR runs — so output is posted to the PR, not the issue fallback.
+func TestProcessItem_PostToPR_CreatesDraftPRBeforePosting(t *testing.T) {
+	skipIfNoGit(t)
+
+	const issueNum = 60
+	const prNum = 300
+
+	// Set up a repo with a real remote so ensureDraftPR can push the branch.
+	// initBareRepo creates a plain local repo; we use it as the "remote" and
+	// clone it so the working repo has origin configured.
+	remoteDir := initBareRepo(t)
+	workingDir := t.TempDir()
+	runCmd := func(dir string, args ...string) {
+		t.Helper()
+		cmd := exec.Command(args[0], args[1:]...)
+		cmd.Dir = dir
+		if out, err := cmd.CombinedOutput(); err != nil {
+			t.Fatalf("cmd %v: %s: %v", args, out, err)
+		}
+	}
+	runCmd("", "git", "clone", remoteDir, workingDir)
+	runCmd(workingDir, "git", "config", "user.email", "test@test.com")
+	runCmd(workingDir, "git", "config", "user.name", "Test")
+
+	wm := NewWorktreeManager(workingDir)
+
+	// prExists tracks whether the draft PR has been created yet.
+	// FindPRForIssue returns 0 until CreateDraftPR is called, simulating the bug
+	// scenario: no PR exists at stage start, PR is created by ensureDraftPR.
+	var prExists bool
+	var draftPRCreated bool
+	// issueCommentBeforePR is set when AddComment is called on the issue number
+	// before the draft PR has been created — this is the bug: full stage output
+	// falls back to the issue instead of the PR.
+	var issueCommentBeforePR bool
+
+	client := &mockGitHubClient{
+		findPRForIssueFn: func(owner, repo string, issueNumber int) (int, error) {
+			if prExists {
+				return prNum, nil
+			}
+			return 0, nil
+		},
+		createDraftPRFn: func(owner, repo, title, head, base, body string, issueNumber int) (int, error) {
+			draftPRCreated = true
+			prExists = true
+			return prNum, nil
+		},
+		addCommentFn: func(owner, repo string, issueNumber int, body string) (int, error) {
+			if issueNumber == issueNum && !prExists {
+				// AddComment on the issue before the PR existed — this is the fallback bug.
+				issueCommentBeforePR = true
+			}
+			return issueNumber*10 + 1, nil
+		},
+		getIssueBodyFn: func(owner, repo string, issueNumber int) (string, error) {
+			if issueNumber == prNum {
+				return "## Verification\n\n(Populated by Implement)\n\n---\n\nCloses #60", nil
+			}
+			return "issue body", nil
+		},
+		updateIssueBodyFn: func(owner, repo string, issueNumber int, body string) error {
+			return nil
+		},
+		fetchLabelsFn: func(owner, repo string, issueNumber int) ([]string, error) {
+			return nil, nil
+		},
+	}
+
+	claude := &mockClaudeInvoker{
+		invokeFn: func(stage *stages.Stage, issue gh.ProjectItem, comments []gh.Comment, resume bool, workDir string, opts InvokeOptions) (string, bool, TokenUsage, error) {
+			return "Implementation done.\nFABRIK_STAGE_COMPLETE", true, TokenUsage{}, nil
+		},
+	}
+
+	stgs := []*stages.Stage{
+		{
+			Name:          "Implement",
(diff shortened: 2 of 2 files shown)


verveguy:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-608 | commit: 614f8cbf | main: cc4cf43678b8f953d3fde6cbc3a02ce3bf21b4c4 | 2026-05-07 00:21 UTC*

## Review Findings

### Fixed

- **Issue**: Test `TestProcessItem_PostToPR_CreatesDraftPRBeforePosting` only asserted the negative (no fallback to issue before PR existed) but not the positive outcome — that output was actually routed to the PR. A regression that broke `postOutputToPR` entirely would pass undetected. **Fix**: Added `prCommentPosted` tracking and assertion that `AddComment` is called on the PR number, confirming output reaches the PR.

### Dismissed

- **Copilot finding (engine/item.go:819)** — The early `ensureDraftPR` call now runs while the lock is still held, slightly extending lock hold time vs. the original where `ensureDraftPR` ran after `releaseLock()`. The fix is correct; the alternative (moving the post step into the `if completed` block after `releaseLock()`) would be a larger refactor out of scope for this bug fix. The lock is already held for the entire Claude invocation, so a few additional API calls are negligible. Dismissed.

### Verified

- Implementation exactly matches the plan: `var prNumber int` lifted before post-output block; early `ensureDraftPR` guard (`completed && CreateDraftPR && PostToPR`) added; `prNumber == 0` check prevents redundant call in `if completed`.
- Non-completing runs unaffected — early guard is inside `if completed`.
- `!PostToPR` path unaffected — early guard also checks `stage.PostToPR`.
- Existing test `TestProcessItem_ImplementStage_UpdatesVerificationOnComplete` has `PostToPR: false`, so early guard never fires for it — no mock changes needed.
- All 12 packages pass with race detector (204s for engine, no races).
- Rebased onto latest main (branch was already current).
- Copilot review addressed: one finding fixed (test coverage), one dismissed (lock hold time).

---

Reviewed the ordering fix for issue #608. Fixed one test gap (Copilot finding): added a positive assertion that output is actually posted to the PR, not just that the fallback-to-issue bug didn't occur. All tests pass with race detector across 12 packages. PR is ready for human review.

---
Used 36/50 turns, 5k input / 9k output tokens.

verveguy:
*branch: fabrik/issue-608 | commit: 614f8cbf | main: 203d95e98a1bc29ac486007ca34d7583be3183ae | 2026-05-07 00:22 UTC*

Both threads resolved. Here's the disposition of the two Copilot findings:

**Copilot thread (engine/item.go:819) — lock hold time:** Dismissed. The early `ensureDraftPR` call does run while the lock is held, but this is an acceptable trade-off — the lock is already held for the entire Claude invocation (minutes), so a few extra API calls are negligible. Moving the post step into the `if completed` block after `releaseLock()` would require a more invasive refactor out of scope for this bug fix. Thread resolved.

**Copilot thread (pr_test.go:950) — missing positive assertion:** Fixed in the Review stage. Added `prCommentPosted` tracking and a `t.Error` assertion that `AddComment` is called on the PR number, verifying output actually reaches the PR (not just that the fallback bug didn't occur). Committed as `614f8cb`. Thread resolved.

No code changes needed in this session — both were already addressed. No issue body update required.

---
**Threads addressed:**
- `engine/item.go:819` — resolved
- `engine/pr_test.go:950` — resolved

verveguy:
*branch: fabrik/issue-608 | commit: 26cc2f00 | main: 203d95e98a1bc29ac486007ca34d7583be3183ae | 2026-05-07 00:29 UTC*

Branch is pushed. All checks pass. Here is the full validation report.

---

## Validation Report

### Requirements: 6/6 passed

- [x] **Draft PR created before `postOutputToPR` on completing `create_draft_pr + post_to_pr` stages**: Early `ensureDraftPR` guard added at `engine/item.go:814–820`, guarded by `completed && stage.CreateDraftPR && stage.PostToPR`. The new test log confirms order: `created draft PR #300` → `detailed Implement output posted to PR #300`.
- [x] **`create_draft_pr: true` but `post_to_pr: false` is unaffected**: Early guard checks `stage.PostToPR`; the existing `TestProcessItem_ImplementStage_UpdatesVerificationOnComplete` test (which has `PostToPR: false`) passes without modification.
- [x] **`updatePRVerification` semantics preserved**: Still called after `ensureDraftPR` inside `if stage.CreateDraftPR`, using `prNumber` populated by whichever path ran.
- [x] **`markPRReady` still happens after PR creation**: Unchanged — still in the `if completed` block after `ensureDraftPR`.
- [x] **Non-completing runs don't trigger early PR creation**: Early guard is inside `if completed`; non-completing runs skip it entirely.
- [x] **Test added to catch ordering regression**: `TestProcessItem_PostToPR_CreatesDraftPRBeforePosting` uses a stateful `prExists` mock — `FindPRForIssue` returns 0 until `CreateDraftPR` fires, then `prNum`. Asserts no fallback-to-issue before PR exists, and confirms output reaches the PR.

### Test Suite: PASSED

- 12 packages, all pass with race detector
- `go vet ./...`: clean
- `go build ./...`: clean
- `TestProcessItem_PostToPR_CreatesDraftPRBeforePosting`: PASS (2.39s)
- `TestProcessItem_ImplementStage_UpdatesVerificationOnComplete`: PASS (2.34s)
- Full engine package (205s): no races, no failures

Only `engine/item.go` and `engine/pr_test.go` changed. The change is a variable scope lift (`var prNumber int`) and a conditional call insertion — no interface changes, no new types, no shared state touched.

### Branch State

- Rebased onto latest main (`203d95e`, `d789f50` incorporated)
- Force-pushed to `origin/fabrik/issue-608`
- Working tree clean

### Verdict: READY TO MERGE

Fixed the ordering bug where `postOutputToPR` ran before `ensureDraftPR` on completing `post_to_pr + create_draft_pr` stages, causing stage output to fall back to the issue instead of the PR. The fix lifts `var prNumber int` before the post-output block and adds an early `ensureDraftPR` guard so the PR always exists when output is posted. All 12 packages pass with race detector; `TestProcessItem_PostToPR_CreatesDraftPRBeforePosting` confirms the ordering is correct.

---
Used 20/50 turns, 7k input / 5k output tokens.

PR #788, opened 2026-05-24
fix(engine): record Duration on comment-run InvocationRecorded

arbeithand:
## Summary

Comment-driven Fabrik invocations were recording zero wall-clock duration, so every 💬 entry in the TUI history pane showed `00:00` instead of the actual time spent.

## Root cause

`engine/comments.go:processComments` applies `itemstate.InvocationRecorded` at two sites — the mid-function "finished" record (line 228) and the late "completed" record (line 393). Neither set the `Duration` field, so it defaulted to `0`. `LastInvocationDuration` then propagated through `InvocationObserver` → `JobCompletedEvent` → `HistoryEntry.Duration` → TUI render as `00:00`.

The stage-run path in `engine/item.go:1015` was already correct, capturing `workerStartedAt` at function entry and passing `Duration: time.Since(workerStartedAt)`.

## Fix

Mirror the stage-run pattern: capture `commentStartedAt := time.Now()` at the top of `processComments` and pass `Duration: time.Since(commentStartedAt)` to both `InvocationRecorded` applications.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./engine/ -run 'TestProcessComments|TestAddComment' -count=1` passes
- [ ] After merge: TUI history shows non-zero elapsed time for 💬 entries

### DIFF
--- a/engine/comments.go
+++ b/engine/comments.go
@@ -38,7 +38,6 @@ func (e *Engine) findNewComments(item gh.ProjectItem) []gh.Comment {
 // processComments handles new user comments on an issue.
 // Flow: 👀 reactions → editing label → invoke Claude → perform actions / update issue body → remove editing label → 🚀 reactions
 func (e *Engine) processComments(ctx context.Context, board *gh.ProjectBoard, item gh.ProjectItem, stage *stages.Stage, comments []gh.Comment, onPIDReady ...func(int)) error {
-	commentStartedAt := time.Now()
 	owner, repo := itemOwnerRepo(item, e.defaultRepo())
 
 	// Merge any unresolved PR review thread comments into the working slice.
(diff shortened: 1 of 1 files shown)


arbeithand:
## Review feedback addressed

Thanks @gemini-code-assist for the catch. All 3 line comments pointed at the same redundancy and have been resolved in commit d271748f.

**The issue:** my initial fix introduced a new `commentStartedAt := time.Now()` at the top of `processComments`, but a `startedAt` variable already exists at line 77 — the same value used to populate the TUI `JobStartedEvent`. Using a separate timestamp would have produced a `HistoryEntry.Duration` that was a few microseconds-to-milliseconds off from the elapsed time the TUI showed in real-time during the run.

**The fix:** dropped `commentStartedAt`, reused the existing `startedAt`. `HistoryEntry.Duration` now matches `JobStartedEvent.StartedAt` exactly, so the post-run history value is anchored to the same clock as the live "Active Jobs" timer.

Net: `+2 −3` (one fewer variable, one consistent timestamp).

Issue #797, opened 2026-05-24
Pre-Implement spawn fails on cross-repo work after fabrik restart — should warm worktreeManagers from on-disk bare clones at startup


verveguy:
## Problem

When Fabrik restarts, the in-memory `worktreeManagers` map is empty. It is populated lazily — only the first time Fabrik spawns Implement against a given repo in the current process lifetime.

If a cross-repo issue is the first to attempt Implement against a repo that hasn't yet been lazily registered in the current session, the pre-Implement spawn check fails with a comment like:

> 🏭 **Fabrik — pre-Implement spawn failed**
>
> The following repos have not yet been initialized by Fabrik (not in worktreeManagers). Fabrik discovers repos lazily — ensure Fabrik can clone each repo (SSH key, webhook config) and that at least one issue from each repo has been processed through the pipeline, then remove `fabrik:paused` to retry:
>
> - verveguy/liminis-graph

The repo *is* fully initialized on disk — bare clone in `.fabrik/repos/verveguy-liminis-graph.git`, worktrees dir in `.fabrik/worktrees/verveguy-liminis-graph/`, and prior PR history shows ~25 successfully-processed fabrik issues against that repo. The only thing missing is an in-memory entry for the current process.

## Why this is wrong

The pre-Implement spawn check is the first place Fabrik asks *"what repos can I work?"* in advance. Up until that point, all repo handling has been reactive (process an issue → discover the repo it belongs to → lazily init). For a multi-repo workflow, that lazy model breaks down for cross-repo issues whose Implement scope spans multiple repos: by definition, at least one of those repos may not have been touched yet in the current session.

Fabrik already knows which repos it can work — they're sitting right there as bare clones in `.fabrik/repos/`. The on-disk evidence is authoritative: if Fabrik cloned it before, Fabrik can clone it again (SSH key is verified to work).

## Proposed fix

At Fabrik startup, walk `.fabrik/repos/*.git` and warm the `worktreeManagers` map with one entry per existing bare clone. This makes the in-memory state match the on-disk state immediately, instead of waiting for lazy discovery.

Optional: log a one-line summary at startup, e.g.

```
[startup] warmed worktreeManagers from 7 on-disk bare clones: verveguy/liminis, verveguy/liminis-graph, verveguy/liminis-framework, …
```

## Reproduction

1. Start Fabrik against a multi-repo project board.
[...]

PR #798, opened 2026-05-24
fix(github): correct addBlockedBy GraphQL mutation name and input field

verveguy:
## Summary

- GitHub's GraphQL mutation is `addBlockedBy` (not `addBlockedByIssue`), with input field `blockingIssueId` (not `blockedById`).
- Verified by schema introspection against `api.github.com/graphql` on 2026-05-24.
- The spec at `specs/sub-issue-decomposition/spec.md` already uses the correct name — the bug was confined to the implementation in `github/project.go`.

## Reproducer

Fabrik instance at `verveguy/#5` paused issue verveguy/liminis#798 (cross-repo decomposition) with:

> 🏭 **Fabrik — pre-Implement spawn failed**
>
> Failed to link child verveguy/liminis-graph#67 as blocked-by of parent: `GraphQL error: Field 'addBlockedByIssue' doesn't exist on type 'Mutation'`
>
> Created so far: verveguy/liminis-graph#67
>
> Manually close any orphaned children, remove `fabrik:paused`, then re-advance to retry.

This is the orphan-then-pause pattern described in the spec under "Partial spawn failure" (line 126). The child issue was created (REST POST works fine), but the GraphQL linking call hit a nonexistent field and crashed.

## Why the bug was subtle

The asymmetric API naming. The **read** field is `Issue.blockedBy(first: N)`, so the original implementer reasonably assumed the **write** mutation would be `addBlockedByIssue`. GitHub instead exposes:

| Op | Name |
|---|---|
| Read | `Issue.blockedBy(first: N)` |
| Mutation | `addBlockedBy(input: { issueId, blockingIssueId })` |
| Payload | `{ issue, blockingIssue, clientMutationId }` |

The code comment even hedged: *"Verify against the GitHub GraphQL explorer if this mutation ever stops working."* It never worked — the verification appears to have skipped the mutation and only checked the read field.

## Test plan

- [x] `go test ./...` passes (full suite: 12 packages, all green; engine tests took ~55s)
- [ ] Verify end-to-end against verveguy/liminis#798 once this lands and the production fabrik binary is upgraded — fabrik should now successfully spawn + link the cross-repo child without orphaning.

## Related

- handarbeit/fabrik#797 — separate bug in the same pre-Implement spawn path (`worktreeManagers` not warmed from on-disk bare clones at startup). Independent root cause.

### DIFF
--- a/github/project.go
+++ b/github/project.go
@@ -1482,20 +1482,21 @@ mutation($projectId: ID!, $contentId: ID!) {
 // blocked by blockerNodeID. After this call, blockerNodeID will appear in
 // issueNodeID's blockedBy(first: N) GraphQL field.
 //
-// NOTE: The mutation name "addBlockedByIssue" matches GitHub's GraphQL API
-// as of the empirical verification documented in the sub-issue decomposition
-// spec (handarbeit/fabrik#769). Verify against the GitHub GraphQL explorer if
-// this mutation ever stops working.
+// GitHub's GraphQL mutation is asymmetrically named relative to its read
+// field: the read is `Issue.blockedBy(first: N)`, but the write is
+// `addBlockedBy` (no "Issue" suffix), with input shape
+// `{ issueId: ID!, blockingIssueId: ID! }`. Verified via schema
+// introspection against `api.github.com/graphql` on 2026-05-24.
 func (c *Client) AddBlockedByIssue(issueNodeID, blockerNodeID string) error {
 	query := `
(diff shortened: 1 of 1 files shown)


arbeithand:
Closing in favour of #800, where the same fix is being processed through Fabrik's own pipeline so the resulting commit is attributed to the bot (arbeithand) rather than a personal handle. The diagnostic content from this PR is preserved in #800's body. Thanks @verveguy for the diagnosis.

PR #809, opened 2026-05-25
feat(release): auto-bump plugin versions on cut-release when source changed

arbeithand:
## Summary

Extend \`scripts/cut-release.sh\` to detect plugin source changes since the previous release tag and auto-bump the affected plugin's \`.claude-plugin/plugin.json\` version. The bump commit lands before the tag, so the version increment is captured in this release.

Fires only inside \`cut-release.sh\` — so the trigger is *exactly* \"on main, for tagged releases.\" No background workflow polling main pushes.

## Why

Claude Code's \`/plugin update\` checks the version field in the plugin manifest at \`ref: main\` of marketplace.json. Same version number → no refresh fires → users keep the cached older content.

This was the bug behind the shadoworg→handarbeit migration: the source URLs were updated in \`plugin/fabrik/skills/*/SKILL.md\` but \`version\` stayed at \`0.1.0\`, so users kept seeing \`fabrik.shadoworg.dev\` until PR #790 manually bumped to \`0.1.1\`. This automation prevents the same trap happening again on every future plugin edit.

## Mechanism

\`\`\`bash
bump_plugin_if_changed plugin/fabrik
bump_plugin_if_changed plugin/fabrik-workflows
\`\`\`

For each path:
1. Find previous \`v*\` tag (\`git describe --tags --abbrev=0\`)
2. \`git diff --name-only \$prev_tag..HEAD -- \$dir ':(exclude)\$manifest'\` — looks for any change *except* the manifest itself (avoids loops if the previous bump commit is the only diff)
3. If anything changed, jq-bump the patch version, commit as \`@arbeithand\` (same credential-helper-nuking pattern the release-notes push uses), push to main
4. Otherwise emit one \`info\` line and move on

The new commit lands before the tag step, so the tag captures it.

## Dry run

Ran the diff logic by hand against current main vs \`v0.0.66\` — neither plugin shows source changes, so the next release wouldn't bump. Working as designed.

## Test plan

- [x] \`bash -n scripts/cut-release.sh\` syntax clean
- [x] Dry-run: with no plugin changes since prior tag, function returns early with \`info\` line
- [ ] First real release after merge: any genuine plugin/source change since the prior tag should produce a visible \"Bump plugin/X version A → B\" step in cut-release output

### DIFF
--- a/scripts/cut-release.sh
+++ b/scripts/cut-release.sh
@@ -64,6 +64,7 @@ cd "$REPO_ROOT"
 
 step() { printf '\n\033[1;34m▶ %s\033[0m\n' "$*"; }
 ok()   { printf '  \033[1;32m✓\033[0m %s\n' "$*"; }
+info() { printf '  \033[1;36m·\033[0m %s\n' "$*"; }
 warn() { printf '  \033[1;33m!\033[0m %s\n' "$*"; }
 die()  { printf '\n\033[1;31m✗ %s\033[0m\n' "$*" >&2; exit 1; }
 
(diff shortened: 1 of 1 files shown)


arbeithand:
Filing as #816 for proper review and implementation via the Fabrik pipeline. This PR stays as a prototype reference — please don't merge as-is. Acceptance criteria, test plan, and open questions live on the issue.

arbeithand:
Closing — the proper implementation is tracked at #816 (currently in cruise mode at Specify). This PR was a prototype reference; the comment posted at PR-open already points at #816 for the canonical work. No further use for the open prototype.

Issue #1056, opened 2026-07-24
Fix base:<branch> linkage-drift cache thrash in runProbeAndDeepFetch (4th #1046 surface; unblocks #1053)


arbeithand:
## Problem

Fourth surface of the `base:<branch>` root cause (#1046), caught end-to-end by the new `TestBaseBranchPipeline` e2e scenario (#1053). GitHub only populates `closedByPullRequestsReferences` for PRs targeting the repository's **default** branch, so the **shallow board probe** reports `LinkedPRNumber == 0` for a non-default-base item even when a real linked PR exists.

`runProbeAndDeepFetch` (`engine/terminal.go:103-127`) compares the probe's `pi.LinkedPRNumber` against the cached PR number and, on the **warm-cache** path, treats a mismatch as "linkage drift" and **invalidates the deep cache**:

```
engine/terminal.go:109   if pi.LinkedPRNumber != cachedPRNum { ...
engine/terminal.go:125     // Warm cache (has been deep-fetched): real linkage drift — invalidate.
engine/terminal.go:126     e.logf(..., "probe: linkage drift (was PR #%d, now PR #%d) — invalidating deep cache", ...)
```

On a `base:<branch>` item the probe **always** returns `0` while the deep cache correctly holds the real PR (via the base-independent linkage from #1047). So every poll: probe sees `0` ≠ cached `#N` → "linkage drift" → invalidate deep cache → re-deep-fetch → base-independent path re-derives `#N` → next probe sees `0` again → **perpetual invalidation thrash**. Observed live: hundreds of `linkage drift (was PR #3609, now PR #0) — invalidating deep cache` lines for a single issue, and a close-on-merge delayed well past 30 min (the issue *does* eventually close, but slowly). Effect: continuous deep-fetch churn (GraphQL burn — ironic given #1046's whole family is about the empty-field cost) and degraded terminal-advance latency for every `base:<branch>` repo.

Neither #1047 (verify-linkage) nor #1050 (review-data feed) touched this probe/drift path.

## Summary

Make the linkage-drift detection in `runProbeAndDeepFetch` recognize that, for a `base:<branch>` item, a shallow-probe `LinkedPRNumber == 0` is **structurally expected** (not a real linkage change) and must not invalidate a warm deep cache that holds a genuine linked PR.

## Requirements

[...]

PR #1064, opened 2026-07-25
fix(engine): verify base:<branch> against remote before falling back to default

### LINKED ISSUE #1058: Harden base:<branch> resolution: ls-remote before falling back to default (5th #1046 surface)
## Problem

`base:<branch>` resolution falls back to the default branch when the base branch is absent from Fabrik's **local bare clone**, without first consulting the remote. `branchExists` (`engine/worktree.go:498`) is `git rev-parse --verify <branch>` against the local clone — no `git ls-remote` / fetch of the base branch. So a base branch that **exists on the remote** but hasn't yet been fetched into the clone is treated as non-existent, and the item silently falls back to the default branch (logging `base: label branch "…" not found on remote, falling back to default` — the "on remote" wording is itself misleading, since it never checks the remote).

Surfaced by `TestBaseBranchPipeline` (#1053): a base branch created seconds before an issue is filed races the clone's fetch and gets missed. Rare for real users (whose base branch is a long-lived trunk, long since fetched — so #1046's @bdueck is unaffected), but it is a genuine correctness gap: falling back to the default branch when the requested base branch DOES exist is exactly the kind of silent wrong-target the `base:<branch>` family (#1046) is about.

## Summary

Before falling back from a `base:<branch>` label to the default branch, verify the branch against the **remote** (`git ls-remote --heads origin <branch>`, or fetch the branch and re-check) — so a base branch that exists remotely is always honored even if the local clone is momentarily stale.

## Requirements

- In `baseBranchForItem` (`engine/item.go:1561`), when the local `branchExists("origin/"+candidate)` check fails, perform a remote check (`git ls-remote --heads origin <candidate>` or a targeted `git fetch origin <candidate>` followed by a re-check) before deciding to fall back.
- If the remote check finds the branch, honor it as the resolved base branch (fetching/updating the local clone as needed so subsequent worktree operations — e.g. `updateWorktreeFromMain`, rebase — succeed against it).
- Only fall back to the default branch (with the warning log + user comment) when the branch is genuinely absent on the remote per the `ls-remote`/fetch check.
- Update the warning log message and the posted issue comment (`engine/item.go:1591,1595`) so their wording reflects an actual remote check having been performed (rather than implying a remote check happened when only the local clone was consulted).
- Keep the fast path fast: the local `branchExists` check remains the first check, and the additional remote check (network round-trip) runs only on a `branchExists`-miss, not on every resolution call.
- Update `docs/state-machine.md` (base:<branch> label row, ~line 94 and ~line 270) if the described fallback behavior changes in a way that affects the as-built description, per this repo's canonical-docs convention.

## Scope

**In scope:**
- The `base:<branch>` resolution path in `baseBranchForItem` (`engine/item.go`) and its local-existence check `branchExists` (`engine/worktree.go:498`).
- The warning/comment wording tied to that fallback decision.
- Test coverage for: (a) branch absent locally but present on remote → resolved correctly, no fallback; (b) branch genuinely absent on remote → fallback with warning + comment, as today.

**Out of scope:**
- The other `base:<branch>` family surfaces already tracked separately: #1047 (verify-linkage), #1050 (review-data), #1056 (probe thrash). Reference #1046/#1053 for context; do not close or fix #1046 itself.
- Any change to how `base:<branch>` labels are parsed, selected among multiples, or inherited by spawned children.
- The #1053 e2e test itself (this issue unblocks/robustifies it as a side effect, but modifying that test is not required here).

## Open Questions

None — the issue is fully specified against the current code path (`engine/item.go:1561-1599`, `engine/worktree.go:498`), consistent with existing docs (`docs/state-machine.md` base:<branch> rows) and the related `#1046` family issues.

## Prior Art / Context

- The codebase already uses `git ls-remote --heads origin` elsewhere as an authoritative, read-only network check for branch existence (e.g. `engine/worktree.go:464` for merge-train branch discovery, `engine/worktree.go:776` for default-branch detection via `ls-remote --symref`) — establishing a precedent for a targeted `ls-remote` call as the appropriate mechanism here, rather than a full `git fetch`.
- `docs/state-machine.md:94,270` currently document the fallback behavior as "falls back to default if not on remote" / "if branch not found on remote" — already consistent with the fixed (remote-checked) behavior this issue implements, so the docs describe the *intended* semantics correctly today; only the code needs to catch up. The docs section may still need a note clarifying that the check is now an authentic remote probe (ls-remote/fetch), if the fix changes any user-visible timing or wording.

## Risks / Dependencies

- Base:<branch> family (#1046): #1047 (verify-linkage), #1050 (review-data), #1056 (probe thrash) — this is the resolution/fork surface. Reference #1046/#1053; do NOT `Fixes #1046`.
- Also unblocks/robustifies the #1053 e2e test (a fresh throwaway base branch would be picked up).
- Adds a network round-trip (`ls-remote` or `fetch`) to the fallback path — must be scoped to the miss case only so the common case (branch already present locally) stays fast, per the fast-path requirement above.

arbeithand:
Closes #1058

## Summary

`baseBranchForItem` previously fell back to the repository's default branch whenever a `base:<branch>` label's candidate branch was missing from Fabrik's **local bare clone** — without ever checking whether the branch actually exists on the remote. A base branch created moments before an issue is filed could race the clone's fetch cycle and get silently (and incorrectly) treated as nonexistent.

## Changes

- **`engine/worktree.go`**: Added `remoteBranchExists` (a read-only `git ls-remote --heads origin <branch>` probe, fail-safe to `false` on any error) and `resolveBaseLabelBranch` (orchestrates local-check → remote-check → fetch → re-verify, under `wm.mu`). The fast path (branch already present in the local clone) is unaffected — it still only performs a single local `rev-parse`. The remote round-trip only happens on a local miss.
- **`engine/item.go`**: `baseBranchForItem` now calls `wm.resolveBaseLabelBranch` instead of the raw local `branchExists` check. Updated the fallback warning log and posted issue comment wording to accurately reflect that an authoritative `git ls-remote` check was performed, not just a local-clone lookup.
- **`engine/worktree_test.go`**: Added `TestRemoteBranchExists` and `TestResolveBaseLabelBranch_{LocalHit,RemoteOnly,Absent}`, covering the probe and the full local-miss → remote-hit → fetch path, using a real bare-clone + origin fixture (`setupTrainRepo`).
- **`engine/engine_test.go`**: Reworked `TestBaseBranchForItem` to use a real origin remote (instead of a repo with no remote configured at all) and added a case proving a branch absent locally but present on the remote resolves correctly without falling back, with an assertion that the local clone can resolve it afterward (proving `EnsureWorktree` would succeed downstream). Also asserts the fallback comment wording mentions the `ls-remote` check.
- **`docs/state-machine.md`** (+ regenerated `docs/llms-full.txt`): Updated both `base:<branch>` label rows to describe the local-check → remote-probe → fetch resolution sequence.

## How to test

```bash
go build ./...
go vet ./...
go test -race ./engine/... -run 'TestBaseBranchForItem|TestRemoteBranchExists|TestResolveBaseLabelBranch'
go test -race ./...
```

All 2438 tests pass repo-wide.


arbeithand:
*branch: fabrik/issue-1058 | commit: fd0168ea | main: e170077e85cffefe3d554d97098f47a97c532d22 | 2026-07-25 00:11 UTC*

No existing PR. I'll emit the `FABRIK_PR_CREATE` marker and finish.


Implemented `resolveBaseLabelBranch`/`remoteBranchExists` on `WorktreeManager` so `base:<branch>` resolution now falls back to the default branch only after an authoritative `git ls-remote` probe (with fetch-on-confirmation) against origin, not just a stale local clone check. Added test coverage for the local-miss/remote-hit path, reworked `TestBaseBranchForItem` against a real origin remote, updated `docs/state-machine.md` wording (+regenerated `docs/llms-full.txt`), and checked off all Plan tasks. Full repo test suite (2438 tests) passes with `-race`; branch pushed, PR marker emitted.

---
Used 48/50 turns, 6k input / 18k output tokens.

arbeithand:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-1058 | commit: fd0168ea | main: e170077e85cffefe3d554d97098f47a97c532d22 | 2026-07-25 00:16 UTC*

## Review Findings

### Verified
- Rebased onto latest `main` (already up to date, no conflicts)
- No actionable external review feedback (only Gemini quota-limit notices, no substantive comments)
- Implementation matches the plan exactly: `remoteBranchExists` (ls-remote probe) and `resolveBaseLabelBranch` (local check → remote probe → fetch → re-verify, held under `wm.mu`) added in `engine/worktree.go`; `baseBranchForItem` in `engine/item.go` simplified to call it, with warning log and issue-comment wording updated to accurately reflect the authoritative remote check
- No deadlock risk: `branchExists` doesn't itself lock `wm.mu`, and none of `baseBranchForItem`'s five call sites (`ci.go`, `comments.go`, `context.go`, `item.go`, `merge_gate.go`) hold `wm.mu` before calling it
- Fast path preserved: local `branchExists` hit returns immediately without any network call
- Fail-safe behavior confirmed: `ls-remote`/fetch errors or a missing `origin` remote all resolve to `false` (fallback), never a hard error
- Test coverage is thorough: `TestRemoteBranchExists` (present/absent/no-origin), `TestResolveBaseLabelBranch_LocalHit/_RemoteOnly/_Absent`, and `TestBaseBranchForItem` reworked against a real local-origin repo with a new "absent locally, present on remote" case plus an assertion that the fallback comment mentions `ls-remote`
- `docs/state-machine.md` updated at both the label-summary row (~line 94) and Reachable-States row (~line 270) to describe the local-check → remote-probe → fetch sequence; `docs/llms-full.txt` regenerated and confirmed byte-identical to a fresh regen (no drift)
- `go test -race -timeout 5m ./...` — 2438 tests passed across 14 packages
- Branch already pushed and in sync with origin

### Fixed
None needed — no issues found.

### Blocking
None.

Reviewed the base:<branch> remote-verification fix. Implementation, tests, and docs all match the plan precisely — no issues found. Rebase clean, full race-enabled suite (2438 tests) passes, docs bundle in sync, no unresolved external feedback. PR is ready for human review.

---
Used 21/50 turns, 5k input / 3k output tokens.

verveguy (commented):
## Review (independent, manual — posted to unblock the Gemini-quota-stalled review gate)

I read the diff (`engine/worktree.go`, `engine/item.go`, `engine/worktree_test.go`, `engine/engine_test.go`, docs) against #1058's acceptance criteria and did some hands-on verification of the new git subprocess calls. Overall the design is right (local-check → remote ls-remote probe → fetch → re-verify, fast path preserved), and the docs/test additions are solid. However, I found one confirmed, exploitable command-injection bug that should block merge as-is.

### BLOCKING: `git fetch origin <candidate>` is vulnerable to argument injection → arbitrary shell command execution

`resolveBaseLabelBranch` (`engine/worktree.go`, new code) does:

```go
cmd := exec.Command("git", "fetch", "origin", candidate)
```

`candidate` is `strings.TrimPrefix(label, "base:")` from `baseBranchForItem` (`engine/item.go:1561`) — the raw, unsanitized suffix of a GitHub label. Anyone able to add a label to a Fabrik-managed issue (triage/write access, and on some repos less) fully controls this string, and it is passed to `git fetch` as a bare positional argument with no `--` separator and no leading-dash validation.

I reproduced this locally against a real bare clone (git 2.50.1):

```
$ git fetch origin '--upload-pack=touch /tmp/proof_marker2'
fatal: Could not read from remote repository...   (exit 128)
$ ls /tmp/proof_marker2
/tmp/proof_marker2   # <-- file was created; the injected command executed via git's upload-pack shell-out, despite the eventual fatal error
```

So a label like `base:--upload-pack=<arbitrary shell command>` on any tracked issue causes Fabrik to execute that command on the host the next time `baseBranchForItem` runs (which is frequently — every stage dispatch, comment processing, CI-fix, rebase-reinvoke, and convergence-failure path all call it). This is the same class of bug as CVE-2017-1000117 (git argument injection via unsanitized refspec/URL arguments). Note `git ls-remote --heads origin <candidate>` (also new, in `remoteBranchExists`) did *not* execute in my quick test with the same payload, but it should still be hardened defensively — its option-parsing behavior isn't a security boundary to rely on.

**Fix** (verified locally): insert a `--` separator before the user-controlled argument in both new call sites so git stops option-parsing and treats `candidate` strictly as a positional refspec/pattern:

```go
// remoteBranchExists
cmd := exec.Command("git", "ls-remote", "--heads", "origin", "--", branch)

// resolveBaseLabelBranch's fetch
cmd := exec.Command("git", "fetch", "origin", "--", candidate)
```

I confirmed `git fetch origin -- '--upload-pack=touch /tmp/x'` correctly fails with `fatal: invalid refspec` and does not execute. Recommend also adding a regression test asserting a `base:` label value starting with `-` is rejected/neutralized rather than reaching git as an option.

Note the pre-existing `branchExists("origin/"+candidate)` local check is *not* vulnerable — the `"origin/"` prefix guarantees the argument can never start with `-`. The new code is the first place in this file where a fully unprefixed, attacker-controlled string reaches a git subprocess argument list, which is presumably how this slipped through.

### Minor: fallback wording conflates "genuinely absent" with "present but fetch failed"

`resolveBaseLabelBranch` collapses two different failure modes into a single `bool`: (a) the branch is genuinely absent per `ls-remote`, and (b) the branch *is* confirmed present on `ls-remote` but the subsequent `git fetch` errored (transient network issue, permissions, etc. — logged internally via `wm.logf`, but not surfaced to the caller). `baseBranchForItem` (`engine/item.go`) then unconditionally logs/posts "Fabrik checked the remote (`git ls-remote`) and could not find branch `%s`" in both cases — which is factually wrong for case (b): the branch does exist, Fabrik just failed to fetch it. Since the fallback comment is deduplicated per-issue-per-candidate (`baseBranchWarnedSet`, `LoadOrStore`), a transient fetch failure produces a permanently-wrong, unretracted user-facing message even after a later successful resolution. Consider having `resolveBaseLabelBranch` return a tri-state (absent / present-but-fetch-failed / resolved) so the two cases get distinct wording, or at least add a test for the "ls-remote hits, fetch fails" path (currently only `LocalHit` / `RemoteOnly` / `Absent` are covered — the fetch-failure branch inside `resolveBaseLabelBranch` has no test coverage at all).

### Minor: repo-wide serialization on every miss, every call

`resolveBaseLabelBranch` runs its network round-trip (ls-remote, and on hit, fetch) while holding `wm.mu` — the same per-repo mutex that serializes worktree/branch creation for *every* concurrent issue on that repo (per `CLAUDE.md`'s documented concurrency model). Since `baseBranchForItem` is called on essentially every dispatch path (stage run, comment processing, CI-fix, rebase-reinvoke, convergence-failure), an issue with a persistently-wrong `base:` label will re-probe the remote on every call, each time blocking all other concurrent workers on that repo for the round-trip duration. None of the git subprocesses in this file use `exec.CommandContext`/timeouts (pre-existing gap, not introduced by this PR — `GIT_SSH_COMMAND` gives a 10s connect timeout for SSH remotes only, nothing bounds HTTPS stalls), so a slow/hanging remote here stalls the whole repo's concurrency, not just this issue. To the PR's credit, the issue text explicitly scopes the repeated-probe cost to #1056 ("probe thrash") as a known follow-up, so I'm not blocking on this, but it's worth being aware the blast radius is "all workers on this repo," not just the one issue with the bad label.

### What's solid
- Fast path is genuinely unaffected — confirmed the local-hit case in `resolveBaseLabelBranch` returns before any network call.
- `TestResolveBaseLabelBranch_{LocalHit,RemoteOnly,Absent}` and `TestRemoteBranchExists` give good coverage of the intended paths, using a real bare-clone + origin fixture rather than mocks.
- The `TestBaseBranchForItem` rework (real origin, `release/remote-only` created only on `srcDir`) is exactly the right way to prove the local-miss → remote-hit → fetch path end-to-end, including the fetch actually populating `refs/remotes/origin/*` (`wm.branchExists` assertion at the end).
- `docs/state-machine.md` + regenerated `docs/llms-full.txt` updates accurately describe the new resolution sequence and match the acceptance criteria in #1058.
- Wording/log changes correctly reflect that a real remote check now happens (modulo the case-(b) nuance above).

### Verdict: BLOCKING ISSUES

The command-injection bug is a real, reproducible arbitrary-command-execution vector reachable by anyone with label-write access to a Fabrik-tracked issue, and should be fixed (adding `--` before the two new user-controlled git arguments, per above) before merge. The two minor points are not blockers but worth a follow-up or a quick fix in the same PR.

arbeithand:
*branch: fabrik/issue-1058 | commit: b7ac159b | main: 638b18f3f3392e9e2dbb17e108c0dc75fedf8864 | 2026-07-25 06:04 UTC*

Clean working tree, all files aligned with the requirements. All checks pass. Writing the validation report now.

## Validation Report

### Requirements: All verified
- [x] `resolveBaseLabelBranch` performs local check → remote `ls-remote` probe → fetch → re-verify, only on local miss (`engine/worktree.go:541-563`)
- [x] Branch confirmed on remote is honored and fetched into the local clone so downstream worktree ops succeed (`resolveBaseLabelBranch`)
- [x] Fallback to default only occurs when `ls-remote` confirms genuine absence (or errors, fail-safe) (`engine/item.go:1582-1596`)
- [x] Warning log and issue comment wording updated to say "verified via ls-remote" (`engine/item.go:1589,1593`)
- [x] Fast path preserved — local `branchExists` hit returns immediately, no network call (`engine/item.go:1582`)
- [x] `docs/state-machine.md` updated at both cited rows (~94, ~270); `docs/llms-full.txt` regenerated and confirmed byte-identical to fresh regen (no drift)
- [x] Test coverage: local-hit, remote-only-hit, absent-both cases (`TestResolveBaseLabelBranch_*`, `TestRemoteBranchExists`, reworked `TestBaseBranchForItem`)

### Additional fix applied during Validate
A command-injection vulnerability was flagged in manual PR review by @verveguy: the new `git fetch`/`git ls-remote` calls passed the untrusted `base:<branch>` label value as a bare positional argument, so a label like `base:--upload-pack=<script>` would be parsed as a git option and execute an arbitrary local program. This fix (`--` separator before the untrusted argument in both calls, plus a regression test `TestResolveBaseLabelBranch_RejectsOptionInjection`) was present uncommitted in the worktree from the Review stage's response to that finding — I verified it was correct and complete, committed it (`b7ac159`), and included it in this validation pass.

### Test Suite: PASSED
- 2438/2439 tests across 14 packages, `-race` enabled
- The 1 intermittent failure (`TestExecute_ConfigYAMLApplied`, a SIGINT-timing test in `cmd/`, unrelated to this change) passed reliably in isolation and repeated runs — confirmed pre-existing flake under load, not a regression

### Branch state
- Rebased onto latest `main` (638b18f)
- All changes committed and pushed (force-with-lease, since local was rebased ahead of origin's copy of the same commits)

### Verdict: READY TO MERGE

Validated the `base:<branch>` remote-verification fix — all requirements met, docs in sync, full race-enabled suite passes (2438/2439, sole failure a confirmed pre-existing unrelated flake). Also found and committed an uncommitted security fix from the worktree (command-injection via `--` separator, per @verveguy's PR review) that hadn't yet made it into a commit. Rebased, pushed, ready to merge.

---
Used 38/50 turns, 7k input / 8k output tokens.

Issue #1069, opened 2026-07-25
--auto-upgrade installs the new binary but never re-execs the running daemon (ran v0.0.73 for 7 days while disk was 0.0.75)


verveguy:
## Summary

`--auto-upgrade` successfully downloads and installs a new release binary, but **never re-execs the running daemon**. The process keeps executing the version it started with — indefinitely — while the binary on disk advances underneath it. There is no warning, and no log line indicating a version mismatch.

Observed live: a daemon ran **v0.0.73 for 7 days** while the on-disk binary went `0.0.73 → 0.0.74 → 0.0.75`. Features shipped in those releases silently did not run.

## Root Cause (confirmed)

`SemverGreater` returns `false` for any running version carrying a non-numeric suffix (e.g. `+dirty`, or a Go pseudo-version like `v0.0.74-0.20260716173320-6198e8102f90+dirty`), because it parses each dot-separated segment with `strconv.Atoi` and bails out on the first non-numeric segment. A daemon whose running-version string carries such a suffix is therefore judged "up to date" against every newer release and never upgrades — this fully explains the observed 7-day staleness independent of whether the re-exec path itself works.

Implementation work for this bug (the `SemverGreater` fix, upgrade/skew logging, safe-point re-exec gating, and re-exec robustness) is tracked in spec-kit issue #1074.

## Evidence

**The download half works — and works well.** Binary was replaced 27 seconds after the release was published:

| Event | Time |
|---|---|
| `v0.0.75` published upstream | 2026-07-24 **23:54:28Z** |
| `~/go/bin/fabrik` mtime | 2026-07-24 **23:54:55Z** (+27s) |

That 27-second gap is auto-upgrade reacting to the release, not a manual `go install`.

**The re-exec half never happens:**

```
$ ps -o pid=,lstart=,command= -p 50456
50456  Sat Jul 18 19:14:59 2026  fabrik --auto-upgrade

$ fabrik --version
0.0.75

$ go version -m ~/go/bin/fabrik | grep mod
    mod  github.com/handarbeit/fabrik  v0.0.75+dirty
```

- Daemon start: **Jul 18 19:14:59** → it launched on **v0.0.73** (v0.0.73 released Jul 16; v0.0.74 didn't exist yet).
- Releases published *during* its uptime: **v0.0.74** (Jul 20), **v0.0.75** (Jul 24).
- Disk is at 0.0.75; the process is still running 0.0.73 code from memory. Same PID, same start time, 7 days later.

[...]

PR #1076, opened 2026-07-25
fix(upgrade): tolerate suffixed running versions, add version-skew watchdog and re-exec hardening

### LINKED ISSUE #1074: --auto-upgrade never picks up newer releases: SemverGreater returns false on suffixed running versions (+dirty/pseudo-version); + re-exec observability/robustness
## Problem

A `fabrik --auto-upgrade` daemon can run **stale code indefinitely** while newer releases ship underneath it. Reported in #1069: a daemon ran **v0.0.73 for 7 days** while the on-disk binary advanced `0.0.73 → 0.0.74 → 0.0.75`; the 0.0.75 "sweep closed items to Done" feature silently never ran (17 issues stuck in `Queued`). No warning, no log line.

Investigation surfaced a concrete, proven defect in the version comparison — matching the "version string matching, changed recently" suspicion.

### Root cause (proven, confirmed still present on `main`): `SemverGreater` returns false for any non-numeric version suffix

`SemverGreater` (`engine/upgrade.go:30`) splits each version on `.` and `strconv.Atoi`s every segment; **any segment that isn't a plain integer makes `Atoi` fail and the function `return false`** ("not greater" → treated as "up to date"):

```
SemverGreater("v0.0.75", "0.0.73")         = true    ← clean: upgrades
SemverGreater("v0.0.75", "0.0.73+dirty")   = false   ← BUG: never upgrades
SemverGreater("v0.0.75", "0.0.73-abc1234") = false   ← any SHA/pseudo-version segment
```

`engine/upgrade_test.go`'s existing `TestSemverGreater` table has no case with a `+`/`-` suffix, confirming this regression is untested and unfixed.

So whenever the **running** binary's version string carries a non-numeric suffix, `PerformReleaseUpgrade` (`engine/upgrade.go:136`) concludes "up to date" and returns without downloading or re-execing — **silently, on every poll, forever.**

This is exactly the shape modern Go produces. `versionWithSHA` (`cmd/version.go`) even documents it: Go 1.24+ stamps VCS pseudo-versions like `v0.0.74-0.20260716173320-6198e8102f90+dirty` into `Main.Version`, and a `go install pkg@vX.Y.Z` build carries its resolved module version there. Any such daemon (or one built against a dirty tree) has a version string that breaks `SemverGreater`. It **compounds with the dirty-release bug (#1073)**: `+dirty` is precisely the poison suffix.

### Regression timeline (verified via `git tag --contains`)

- **Latent since v0.0.2** (`e7cf0eaf`, #178, the original `fabrik upgrade` self-upgrade implementation) — the `Atoi`-per-segment logic has never tolerated a non-numeric suffix. Harmless for months because nothing fed it a suffixed running version.
- **Became live in v0.0.72** (`6b9978a2`, #994, "fall back to `Main.Version` for `go install` builds", 2026-07-15). Before this, a `go install`-built daemon reported `Version = "dev"`, so `checkAndUpgrade` took the dev git-rebuild path and never called `SemverGreater`. From v0.0.72 on, such a daemon reports `Main.Version`; for a **non-tag install** (`@main`/branch/pseudo-version) that string is `v0.0.7X-0.<timestamp>-<sha>` — not `dev`-prefixed, so it takes the release-download path, where `SemverGreater` chokes on the suffixed segment and silently reports "up to date."
- v0.0.74 (`b3e57712`, 2026-07-19) refined the working-tree-vs-installed distinction but kept the exposure open.

**Affected matrix:**

| Daemon build | Running version string | Auto-upgrade |
|---|---|---|
| Release tarball (GitHub Releases / prior auto-upgrade) | clean `0.0.7X` | always worked |
| `go install …@vX.Y.Z` (clean tag) | clean `vX.Y.Z` | always worked |
| `go build` from source checkout | `dev(<sha>)` | separate dev-rebuild path (not this bug) |
| `go install …@main`/branch/pseudo-version, ≥ v0.0.72 | `v0.0.7X-0.<ts>-<sha>` | broken — silently never upgrades |

This means the original #1069 report's "binary on disk *was* replaced while the daemon didn't" is consistent with a **fleet sharing `~/go/bin/fabrik`** — a different, cleanly-versioned instance (or a manual `go install`) updated the file on disk while the broken daemon kept running its own in-memory image. The `SemverGreater` suffix bug alone explains the stale daemon; it does not require the re-exec path to have run or failed.

### Secondary gaps — status as of this spec (verified against current `main`, not just the original #1069 report)

- **Observability — already substantially implemented.** `PerformReleaseUpgrade` (`engine/upgrade.go:125-256`) logs the decision and every phase/failure via its `logf` parameter (new-release found, downloading, upgraded, re-executing, and each error branch). The poll-loop caller's discarded error (`engine/upgrade.go:401`, `_ = PerformReleaseUpgrade(...)`) is intentional and documented in an adjacent comment: failures are already logged inside `PerformReleaseUpgrade` itself, and the caller's contract is non-fatal. **This requirement is likely already met** — treat it as "verify full coverage of the decision/outcome messages the issue calls for, close any specific gap found" rather than a from-scratch build.
- **Safe-point gating — already implemented.** `engine/poll.go:1116-1140` already defers `checkAndUpgrade()` (and thus re-exec) until a poll cycle dispatches zero new workers, has zero in-flight workers (checked via the engine's `Store`), and has done so for `idleUpgradeThreshold` (2) consecutive polls — with an explicit comment noting this exists specifically because `syscall.Exec` would otherwise kill in-flight workers. **This requirement is likely already met.** One nuance worth Research's attention: whether a webhook-triggered concurrent poll could dispatch a new worker while the synchronous download+exec of `checkReleaseUpgrade` is in flight on another poll cycle — if so, closing that race is in scope; if not, no further change is needed here.
- **Re-exec robustness — still genuinely missing.** No codesign/`xattr`/quarantine handling exists anywhere in the codebase. The macOS-arm64 freshly-downloaded-binary re-signing scenario and any `os.Executable()`/symlink path-mismatch case are unaddressed. `syscall.Exec` failure is currently logged and the error returned/discarded (per above) but there's no explicit "never silently stale" safeguard beyond that.
- **Version-skew warning — still genuinely missing.** No code reads the on-disk binary's version or compares it to the running version at poll time; there is no periodic skew warning anywhere in `engine/`.
- **Explicit reload trigger — partially exists, differently than described.** `SIGHUP` (`engine/sighup_unix.go`) already does a full worker-drain + `syscall.Exec` restart **of the currently-installed on-disk binary** (not merely a cache re-bootstrap, contrary to the original report) — so if a newer binary is already on disk, `SIGHUP` does pick it up. What it does *not* do is trigger a fresh `PerformReleaseUpgrade` check against GitHub Releases. This remains explicitly optional (see Scope).

## Summary

Make `--auto-upgrade` reliably pick up newer releases: fix the version comparison so a suffixed running version doesn't defeat it (the one confirmed, unfixed defect), add the missing version-skew warning, and investigate/handle re-exec robustness (macOS codesign, executable-path mismatches) so a successful install never leaves the daemon silently stale. The observability and safe-point-gating gaps described in the original report appear to already be resolved on `main`; this work should verify that and close only the specific gaps Research finds, not rebuild them.

## Requirements

- **Fix `SemverGreater`** to compare only the numeric `MAJOR.MINOR.PATCH` core, ignoring any pre-release/build/pseudo-version suffix (`-…`, `+…`) on either operand. A running version of `0.0.73+dirty`, `v0.0.74-0.2026…-sha+dirty`, or `0.0.73-abc1234` must still compare as older than `v0.0.75`. Add table-driven tests covering clean, `+dirty`, SHA-suffixed, pseudo-version, and `v`-prefix-mismatch cases. (Consider a small, well-tested parse rather than `Atoi`-per-segment.)
- **Regression test at the decision level:** a daemon whose running version carries a non-numeric suffix must upgrade to a newer release — cover this with a test at the `PerformReleaseUpgrade` level, mocking `FetchLatestRelease`. Include a case matching the confirmed real-world exposure: a `go install …@main`/branch pseudo-version running string (e.g. `v0.0.72-0.20260716173320-6198e8102f90+dirty`) comparing older than a newer clean tag — this is the exact shape that has been silently broken in the wild since v0.0.72 (see Regression Timeline).
- **Verify upgrade-path observability is complete:** confirm `PerformReleaseUpgrade`'s existing `logf` calls cover the decision (`new release vX (running vY) — downloading/upgrading`), the install outcome, and the re-exec attempt at INFO level; close any specific gap found. Do not re-architect logging that already works.
- **Verify safe-point gating is complete:** confirm the existing idle+no-in-flight-workers gate in `engine/poll.go` reliably defers re-exec while any stage is `in_progress`, including under concurrent webhook-triggered polls; close any race found.
- **Guarantee re-exec-or-loud-failure:** after a successful atomic replace, the daemon must re-exec into the new binary. If `syscall.Exec` fails, log it prominently and leave the process running the old binary rather than dying — but never leave it silently stale. Investigate and handle the macOS-arm64 freshly-downloaded-binary codesign/quarantine case and any `os.Executable()`/symlink path mismatch that could cause a replaced-but-not-re-exec'd state.
- **Version-skew warning (net new):** at poll time, determine the on-disk binary's version and compare it to the running version; if they differ, log a periodic warning so a stuck/deferred upgrade is discoverable.
- **(Optional, from #1069, out of primary scope)** an explicit reload trigger (`fabrik reload` or a signal) that forces a fresh `PerformReleaseUpgrade` check against GitHub Releases, distinct from `SIGHUP`'s existing in-place restart of whatever binary is already on disk.

## Scope

- **In:** `SemverGreater` + tests; verifying/closing gaps in the existing observability and safe-point-gating in `engine/upgrade.go` (`PerformReleaseUpgrade`, `checkReleaseUpgrade`, `checkAndUpgrade`) and its poll-loop caller; re-exec robustness (macOS codesign, path-mismatch handling); version-skew warning.
- **Out:** the dev-build (`dev(<sha>)`) git-pull rebuild path except where the same `SemverGreater`/logging fix applies; the `+dirty` release-artifact stamp itself (fixed separately in #1073 — but the `SemverGreater` fix must be robust to it regardless); extending `SIGHUP` or adding a new `fabrik reload` trigger to force a fresh release check (optional follow-up, not required for this issue).

## Prior Art / Context

Verified against `main` at the time of this spec (commit `8ce4f20`):
- `engine/upgrade.go:125-256` (`PerformReleaseUpgrade`) already logs decision/outcome/failure at each phase via its `logf` parameter.
- `engine/upgrade.go:394-401` (`checkReleaseUpgrade`) discards `PerformReleaseUpgrade`'s error deliberately and documents why (failures already logged inside; non-fatal contract).
- `engine/poll.go:1116-1140` already gates `checkAndUpgrade()` on zero dispatched workers, zero in-flight workers, and `idleUpgradeThreshold` (2) consecutive idle polls, with an explicit comment tying this to `syscall.Exec` safety.
- `engine/sighup_unix.go` already performs a full drain + `syscall.Exec` restart of the current on-disk binary on `SIGHUP` — it restarts in place, it does not fetch a new release.
- No code anywhere reads the on-disk binary's version for comparison against the running version (searched `engine/` for version-skew/on-disk-version logic — none found).
- No codesign/`xattr`/quarantine handling exists anywhere in the codebase (searched for `codesign`, `xattr`, `com.apple.quarantine`, `Gatekeeper` — no matches outside `tests/e2e/README.md`).
- Regression confirmed live since v0.0.72 (`6b9978a2`, #994) via `git tag --contains`; latent since v0.0.2 (`e7cf0eaf`, #178). See Research findings' Regression Timeline for the affected-configuration matrix.

## Risks / Dependencies

- Surfaced by **#1069** (diagnostic report). Compounds with **#1073** (the `+dirty` release stamp) — but the `SemverGreater` fix must stand on its own since pseudo-versions produce non-numeric suffixes even with a clean release process.
- This is **critical daemon-lifecycle code** (self-replacing process via `syscall.Exec`); a regression could brick the upgrade path fleet-wide. Prefer a human merge (cruise) and careful review of the re-exec path.
- Reference #1069 and #1073 in the PR body; do NOT `Fixes #1069` beyond the auto-upgrade portion (the dirty-release half is #1073).
- Because two of the four "secondary gaps" in the original report already appear resolved on `main`, Research should re-confirm this against the `main` state at implementation time (not just this spec) before scoping Plan — behavior may have shifted further since this spec was written.

## Labels

stage:Specify:complete, fabrik:cruise

arbeithand:
Closes #1074

## What this does

Fixes the confirmed root cause of `--auto-upgrade` silently never firing when the running binary's version string carries a non-numeric suffix (`+dirty`, a SHA suffix, or a full Go pseudo-version like `v0.0.72-0.20260716173320-6198e8102f90+dirty`). Also adds a version-skew watchdog and hardens the re-exec path so a stuck upgrade is never silent.

References #1069 (original diagnostic report) and #1073 (the `+dirty` release-artifact stamp this fix is robust against, but does not itself fix).

## Key changes

- **`SemverGreater` fix** (`engine/upgrade.go`): now parses only the numeric `MAJOR.MINOR.PATCH` core — strips a leading `v` and everything from the first `-`/`+` before splitting on `.` — instead of `Atoi`-ing every dot-separated segment. Previously any non-numeric segment made the whole comparison silently return `false` ("not an upgrade"). Table-driven tests cover `+dirty`, SHA suffixes, full pseudo-versions, suffixed-vs-suffixed equal cores, and non-numeric garbage.
- **Regression test at the decision level**: a `PerformReleaseUpgrade`-level test using the exact confirmed real-world exposure (a `go install …@main`/branch pseudo-version running string) against a newer clean release tag, asserting the download path is reached — plus a companion "not eagerly upgraded" case to guard against overcorrecting in the other direction.
- **Version-skew watchdog** (`Engine.checkVersionSkew`, new): piggybacked on the existing idle+no-in-flight-workers gate in `poll.go` (no new counter/threshold). Spawns `<exe> --version` with a 5s timeout and compares it to the running process's version. A mismatch — from a stuck upgrade, a failed re-exec, or an externally-replaced binary (e.g. a fleet sharing `~/go/bin/fabrik`) — is recorded as a persistent `warnings.Entry` (following the existing `checkAllowAutoMerge` pattern) so it survives restarts and surfaces in the TUI Warnings panel, with a suggested `kill -HUP <pid>` fix.
- **Re-exec hardening**: `syscall.Exec` is now called through a testable package-level seam (`upgradeExecFn`); on failure the log line is upgraded to an unambiguous `CRITICAL:` message stating the binary was replaced on disk but the process is still running the old one. On `darwin`, the freshly-downloaded binary is best-effort re-signed (`xattr -cr` + `codesign --force --sign -`) before the re-exec, working around a documented Apple Silicon AMFI quirk (see `tests/e2e/README.md`) where a binary materialized via a fresh write can be killed on exec. This is hardening based on documented precedent, not verifiable in this sandbox or Linux CI — flagging for manual verification on real Apple Silicon hardware if possible.
- **Verified, not rebuilt**: confirmed `PerformReleaseUpgrade`'s existing `logf` calls already cover every decision/phase/failure point, and confirmed there is no concurrency race between webhook-triggered polls and the synchronous upgrade/exec path (the poll loop is single-threaded for upgrade purposes) — both documented in `docs/state-machine.md` rather than re-implemented.
- Docs: `docs/state-machine.md` (as-built) and `docs/USER_GUIDE.md` (user-facing) updated; `docs/llms-full.txt` regenerated per repo convention.

## How to test

- `go test -race ./engine/... -run 'SemverGreater|PerformReleaseUpgrade|CheckVersionSkew|ResignDarwin'`
- Full suite: `go build ./... && go vet ./... && go test -race ./...` — passes except one pre-existing, network-sandboxed failure in `cmd.TestExecute_ConfigYAMLApplied` (confirmed reproducible on `main` before this branch's changes; unrelated to this work — it depends on a live GitHub GraphQL call that hangs in this sandbox).


arbeithand:
*branch: fabrik/issue-1074 | commit: 7a13c78d | main: 638b18f3f3392e9e2dbb17e108c0dc75fedf8864 | 2026-07-25 05:44 UTC*

I'll pause here and wait for the background test run to finish — the harness will notify me automatically when it completes, and I'll then finalize the PR (task checkbox update was skipped gracefully since no formatted Plan comment exists on GitHub yet).No PR exists yet — I need to emit the `FABRIK_PR_CREATE` marker for the engine to create it. Let me finalize with a summary and the completion markers.


Fixed `SemverGreater` to compare only the numeric core, ignoring `+dirty`/SHA/pseudo-version suffixes — the confirmed root cause of `--auto-upgrade` silently never firing. Added a version-skew watchdog (persistent TUI warning when the on-disk binary's version differs from the running process's) and hardened the re-exec path (testable exec seam, CRITICAL failure logging, best-effort macOS codesign re-sign). Verified the observability and safe-point-gating gaps from the original report were already resolved on `main` and documented that instead of rebuilding it. All code, tests, and docs are committed and pushed; full test suite passes except one pre-existing, network-sandboxed `cmd` test unrelated to this change.

---
Used 72/100 turns, 11k input / 24k output tokens.

arbeithand:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-1074 | commit: 7a13c78d | main: 638b18f3f3392e9e2dbb17e108c0dc75fedf8864 | 2026-07-25 06:14 UTC*

Everything is verified. Final output:

## Review Findings

### Verified
- Rebased onto latest `main` — branch was already up to date (`fabrik/issue-1074` is `origin/main` + this work).
- No actionable external review feedback — the only PR comments are automated `gemini-code-assist` daily-quota-limit notices.
- `go test -race -timeout 5m ./...` — full suite passes. `engine` package tests hit an intermittent 5-minute timeout twice (goroutine leak in unrelated merge-train semaphore tests, `engine/merge_train.go:206`, `prepareTrainWorker`), but this diff touches only `engine/upgrade.go`/`engine/poll.go`. Confirmed pre-existing and unrelated by running the exact same package against `origin/main` in a scratch worktree and repeating both branches multiple times: `main` 2/2 clean, this branch 3/4 clean, with the same leaked-goroutine signature in both hangs — not a regression introduced here.
- The one consistently-failing test, `cmd.TestExecute_ConfigYAMLApplied`, is a known network-sandboxed failure (depends on a live GitHub GraphQL call) — matches what the PR description already discloses, `cmd/` package is untouched by this diff.
- `docs/llms-full.txt` regeneration confirmed in sync with `docs/state-machine.md` and `docs/USER_GUIDE.md` via `scripts/generate-llms-full.sh` (no diff produced).
- `SemverGreater`/`semverCore` fix: correct numeric-core parsing, strips `v` prefix and everything from the first `-`/`+`, zero-pads and compares — preserves all pre-existing behavior (mismatched segment counts, equal versions → false, parse error → false) while fixing the suffix bug. Table-driven tests cover `+dirty`, SHA suffix, full pseudo-version, suffixed-vs-suffixed equal core, and non-numeric garbage on both operands.
- `PerformReleaseUpgrade`-level regression tests use the exact confirmed real-world exposure string and include a companion "not eagerly upgraded" guard.
- Re-exec hardening: `upgradeExecFn`/`upgradeExecutableFn` seams are properly save/restored in tests; the exec-failure test verifies both the CRITICAL log line and that the on-disk replace still happened despite the exec failure. `resignDarwinBinary` is correctly gated to `darwin`, uses `exec.LookPath` guards, and is non-fatal; verified as a true no-op on non-darwin.
- `checkVersionSkew`: correctly throttled (piggybacked on the existing idle+no-in-flight gate, no extra spawn on busy polls), 5s subprocess timeout, uses the same `os.Executable`+`EvalSymlinks` pattern as the existing exec paths, and follows the established `warnings.Record`/`warnings.Clear` pattern (`checkAllowAutoMerge` precedent) including the `_ = ` discarded-error convention already used there. Confirmed `--version` (`cmd/root.go:195-196`) prints only `Version` to stdout with no other output ahead of it, so the subprocess-output comparison is reliable for both release and dev-build version strings.
- Docs (`docs/state-machine.md`, `docs/USER_GUIDE.md`) accurately describe the shipped behavior, including the "no gap found" note for the safe-point-gating race investigation.

### Fixed
- None — no issues found requiring a code change.

### Blocking
- None.

Nothing to commit; working tree is clean and already pushed. PR #1076 is open (not draft) against `main` and ready for human review — `mark_pr_ready_on_complete: true` is already satisfied since it was never converted to draft.

Reviewed the `SemverGreater` fix, version-skew watchdog, and re-exec hardening for issue #1074. Rebase was already current, build/vet are clean, and the full test suite passes (one intermittent engine-package timeout was traced to a pre-existing, unrelated merge-train goroutine leak and confirmed reproducible on `main` too, not a regression from this work). No issues found — PR #1076 is ready for human review.

---
Used 23/50 turns, 0k input / 8k output tokens.

arbeithand:
*branch: fabrik/issue-1074 | commit: 4c3cae63 | main: 540d97a76f7976d561a0f005bdbe3fe2b44c0055 | 2026-07-25 12:30 UTC*

Everything is in sync, clean, and pushed. No issue body changes are needed — these were review-thread findings on the PR, not spec changes.

## Summary

Processed 8 `gemini-code-assist` inline review findings from the `/gemini review` trigger:

**Fixed (3 code changes, committed as `4c3cae6`):**
- `checkVersionSkew` now accepts the poll loop's `ctx` instead of `context.Background()`, so its subprocess is cancelled on engine shutdown — updated `poll.go`'s call site and all 3 test call sites.
- `resignDarwinBinary`'s `xattr`/`codesign` calls now run under a 5s `context.WithTimeout` to prevent an indefinite hang blocking the synchronous upgrade path.
- `SemverGreater` now treats a clean version as greater than a suffixed one when their numeric cores are equal (e.g. `v0.0.73` > `v0.0.73-alpha`/`v0.0.73+dirty`), so a daemon running a `+dirty`/pseudo-version build of the same core version upgrades once the clean release exists, rather than comparing "equal" and staying stuck. Added 3 new table-driven test cases.

**Dismissed (1 finding, with evidence):**
- The "high" priority `syscall.Exec` Windows-compilation finding: verified via `GOOS=windows go build ./engine/...` that the `engine` package already fails to cross-compile for Windows for unrelated pre-existing reasons (`syscall.Flock` in `poll.go`), so this finding introduces no new breakage. A platform split is out of scope for this issue.

Build, vet, and the full `engine`/`cmd` test suites all pass. Committed and force-pushed (the branch had already been rebased onto a newer `main` in this worktree before this turn). No `FABRIK_STAGE_COMPLETE` emitted per this stage's contract.

---
**Threads addressed:**
- `engine/upgrade.go:327` — resolved
- `engine/upgrade.go:168` — resolved
- `engine/upgrade.go:80` — resolved
- `engine/poll.go:1137` — resolved
- `engine/upgrade.go:470` — resolved
- `engine/upgrade_test.go:50` — resolved
- `engine/upgrade_test.go:777` — resolved
- `engine/upgrade_test.go:814` — resolved
- `engine/upgrade_test.go:856` — resolved

Resolved 9 review thread(s) across 9 comment(s).

arbeithand:
*branch: fabrik/issue-1074 | commit: 4c3cae63 | main: 540d97a76f7976d561a0f005bdbe3fe2b44c0055 | 2026-07-25 12:33 UTC*

No unresolved review threads. All checks are complete — everything passes.

## Validation Report

### Requirements: 6/6 met
- [x] **`SemverGreater` fix**: `engine/upgrade.go:44-87` — new `semverCore` helper strips `v` prefix and everything from the first `-`/`+`, compares only the numeric core, zero-padded. Also handles the "equal core, one suffixed" edge case (clean beats suffixed) added during review. Table-driven tests (`engine/upgrade_test.go:25-67`) cover clean, `+dirty`, SHA suffix, full pseudo-version, `v`-prefix mismatch, suffixed-vs-suffixed equal core, and non-numeric garbage — all pre-existing cases preserved.
- [x] **Regression test at decision level**: `TestPerformReleaseUpgrade_SuffixedVersionUpgrades` uses the exact confirmed real-world exposure string (`v0.0.72-0.20260716173320-6198e8102f90+dirty`) and asserts the download path is reached; companion `TestPerformReleaseUpgrade_SuffixedVersionNotEagerlyUpgraded` guards against over-eager regression.
- [x] **Observability verified**: confirmed already complete on `main`, no rebuild — matches Research's finding.
- [x] **Safe-point gating verified**: confirmed race-free (single-threaded poll loop) — no code change, documented in `state-machine.md`.
- [x] **Re-exec-or-loud-failure**: `upgradeExecFn` seam + `CRITICAL:` log line (`upgrade.go:324-327`) on exec failure; `resignDarwinBinary` (darwin-gated, `exec.LookPath`-guarded, 5s timeout) handles the macOS codesign/quarantine case. Tested via `TestPerformReleaseUpgrade_ExecFailureLogsAndReturnsError` and `TestResignDarwinBinary_NoopOnNonDarwin`.
- [x] **Version-skew warning**: `checkVersionSkew` (`upgrade.go:468-506`) spawns `<exe> --version` with a 5s timeout, compares to running version, records/clears a `warnings.Entry` via the established `warnings.Record`/`Clear` pattern. Wired into `poll.go:1137` inside the existing idle+no-in-flight gate. Tested for match/mismatch/error paths.

### Test Suite: PASSED
- 2461 passed, 1 skipped, 1 failed across 14 packages (`go test -race -timeout 5m ./...`)
- The 1 failure (`cmd.TestExecute_ConfigYAMLApplied`) is a known network-sandboxed test requiring live GitHub GraphQL access — `cmd/` is untouched by this diff (confirmed via `git diff origin/main..HEAD --name-only`), pre-existing and unrelated.
- Build: clean. Vet: clean.

Diff is scoped to `engine/upgrade.go`, `engine/poll.go`, `engine/upgrade_test.go`, and docs. No shared interfaces changed.

### Documentation
`docs/state-machine.md` and `docs/USER_GUIDE.md` both updated with accurate descriptions of the fix and new watchdog; `docs/llms-full.txt` regeneration confirmed byte-identical to committed version (no drift).

### Branch/PR State
Rebased and up to date with `main` (no-op rebase). Working tree clean. HEAD matches `origin/fabrik/issue-1074`. PR #1076 is open, not draft, mergeable. All 9 prior review threads resolved, 0 unresolved.

### Verdict: READY TO MERGE

Validated PR #1076 for issue #1074: all 6 requirements confirmed implemented and tested (SemverGreater suffix fix, regression test at PerformReleaseUpgrade level, verified observability/safe-point-gating already complete, re-exec hardening with macOS codesign, version-skew watchdog). Full test suite passes (one pre-existing, unrelated network-sandboxed test failure in `cmd/`), build/vet clean, docs bundle in sync, branch rebased and pushed, PR mergeable with all review threads resolved. Ready to merge.

---
Used 32/50 turns, 17k input / 6k output tokens.

PR #1105, opened 2026-07-26
release: guard against publishing artifacts built from a dirty tree

### LINKED ISSUE #1070: release: guard against publishing artifacts built from a dirty tree (verify vcs.modified on the built binary)
## Problem

The published `v0.0.75` release artifact was built from a **dirty git working tree** — its Go build info records `vcs.modified=true`. The binary therefore corresponds to no committed revision and cannot be reproduced from the tagged source.

```
$ go version -m ~/go/bin/fabrik
    mod   github.com/handarbeit/fabrik   v0.0.75+dirty
    build vcs.revision=e170077e85cffefe3d554d97098f47a97c532d22
    build vcs.time=2026-07-24T23:53:12Z
    build vcs.modified=true            <-- dirty tree at build time
$ fabrik --version
0.0.75                                  <-- ldflags masks the +dirty
```

Why it matters: the artifact is **not reproducible** (rebuilding `e170077` will not produce it), **not auditable** (uncommitted code shipped, so `git bisect` cannot find a regression introduced that way), and **silently masked** — `-ldflags -X .../cmd.Version` makes `--version` print a clean `0.0.75`, so nothing surfaces it to a user or maintainer. It only appears in `go version -m`.

The bad `v0.0.75` artifact will be superseded by the next release, so re-cutting it is neither necessary nor safe: this is a public Go module and `proxy.golang.org` has already cached that version's zip and checksum, so republishing different content under the same tag would give users a checksum mismatch. **This issue is scoped to the guard, not the artifact.**

Root cause is **not yet identified**, and the guard must not depend on identifying it:
- `scripts/cut-release.sh` only tags and pushes; it does **not** build or upload the published artifacts (its `go build ./...` at line ~132 is purely a local sanity check, run before the tag exists).
- `.github/workflows/release.yml` builds via goreleaser and contains **no tree-mutating step** before the build (checkout → verify tag on main → setup-go → verify release notes → goreleaser).
- `.goreleaser.yaml` has **no `before:` hooks**.
- `cut-release.sh`'s dirty-tree pre-flight (line ~79) deliberately **allowlists** `release-notes/<VERSION>.md` and `plugin/known_embedded_versions.go` as permitted-dirty. That allowlist is a plausible contributor if any build ever happens locally, and is worth reviewing — but is not confirmed as the cause.

Because the source is unknown, the guard must verify the **built artifact**, not merely the working tree.

## Summary

Add a verification step to the release-publishing path (goreleaser / `.github/workflows/release.yml`) that inspects every binary produced by the release build — via `go version -m` or equivalent — and asserts `vcs.modified=false` and `vcs.revision` equal to the tag's target commit, for each one, before any artifact is published. A failure aborts the release loudly, naming the offending binary and its actual vs. expected values. Separately, review and document the disposition of `cut-release.sh`'s dirty-tree allowlist, and update release-process docs to describe the new guard.

## Requirements

- **Verify the artifact, not the tree.** After the release binaries are built and before they are published, assert that each built binary reports `vcs.modified=false`.
- **Check every binary in the release matrix**, not just one — the goreleaser build produces multiple `GOOS`/`GOARCH` targets (currently darwin/linux × arm64/amd64), and any one of them being dirty must fail the release.
- **Assert the revision matches the tag**, per binary: `vcs.revision` must equal the commit the pushed tag points at. A clean build of the *wrong* commit is equally unpublishable.
- **Fail the release loudly, before publish.** A dirty artifact or revision mismatch must abort the release with a non-zero exit — no GitHub Release, checksums, or Discussions announcement gets created. It must not warn-and-continue.
- **Failure output must name**: the offending binary (path or target triple), its recorded `vcs.revision` / `vcs.modified`, and the expected revision (the tag's commit SHA).
- **Guard runs in the publishing path** — i.e. somewhere in `.github/workflows/release.yml` / `.goreleaser.yaml`'s build-to-release flow — not only in `scripts/cut-release.sh`, since that script does not build what ships.
- **A clean release publishes unchanged** — no new failures, no new required manual steps, when every binary is `vcs.modified=false` at the tag's commit.
- **Review the `cut-release.sh` allowlist** (line ~79, `release-notes/<VERSION>.md` and `plugin/known_embedded_versions.go`): decide whether each should remain permitted-dirty, and document the reasoning in-repo. If either stays allowlisted, explicitly note that it cannot affect the CI-built artifact, since the release workflow does a fresh checkout of the tag and never carries local working-tree state into the goreleaser build.
- **Update release-process documentation** to describe the guard: what it checks, where it runs, and what a maintainer should do when it trips (note explicitly that this does not mean re-cutting `v0.0.75` — see Scope).
- **Test coverage**:
  - Positive: a binary built from a clean checkout passes the guard.
  - Negative: a binary built from a deliberately dirtied tree (touch a tracked file before building, in a temp clone) fails the guard.
  - Negative: a binary built from a commit other than the tag's target fails the revision check.
  - Use stdlib `testing` where the guard logic is Go; otherwise a shell-level test invoked from CI.
- **Optional, decide during Plan**: have `fabrik --version` append `+dirty` when the running binary's own build info reports `vcs.modified=true`, so a locally-built or improperly-released binary self-identifies instead of masquerading as a clean release. This is a user-visible change to `--version` output — decide deliberately, and only within the bound below (no other change to version-injection via ldflags).

## Scope

**In scope:**
- Release-pipeline verification of built artifacts (all binaries in the goreleaser build matrix).
- Failure behavior (loud abort, before publish, with named diagnostics).
- The `cut-release.sh` allowlist review and its documented disposition.
- Release-process documentation updates.
- The optional `+dirty` runtime surfacing in `--version` (decision deferred to Plan).

**Out of scope:**
- Re-cutting or replacing the existing `v0.0.75` artifact (unsafe — see Problem).
- Root-causing the specific `v0.0.75` dirt (desirable but not required; the guard must work without it).
- Any change to version-injection via ldflags beyond the optional `+dirty` surfacing above.

## Open Questions

None — the requirements above are fully specified. The one deliberate judgment call (`+dirty` surfacing in `--version`) is intentionally deferred to the Plan stage per the requirement above, not left ambiguous.

## Prior Art / Context

- `cmd/version.go` already inspects `debug.BuildInfo.Settings` for `vcs.revision` (and its test file, `cmd/version_test.go`, already constructs fixtures with `vcs.modified: "true"`) as part of the existing dev-build version-string logic. This is directly relevant prior art for whichever stage implements the guard's own build-info parsing, or the optional `+dirty` surfacing — the pattern for reading these settings out of `debug.BuildInfo` already exists in this codebase.
- `go version -m <binary>` is the standard way to inspect embedded build info (module version, `vcs.revision`, `vcs.modified`, `vcs.time`) on an already-built binary without re-running `go build`.
- goreleaser supports per-build hooks (e.g. `builds[].hooks.post`) that run after a binary is built and before it is archived/published — a plausible integration point, to be evaluated in Research.

## Risks / Dependencies

- `proxy.golang.org` has already cached `v0.0.75`'s zip and checksum — any approach that touches that tag's content is unsafe; this issue must not attempt to fix `v0.0.75` itself.
- The guard's insertion point depends on goreleaser's hook/lifecycle model supporting a fail-fast check between build and publish; Research should confirm the exact mechanism before Plan commits to an approach.
- The next real release (v0.0.76 or later) is the first live test of this guard in production — a clean result narrows the original incident to a one-off; a dirty result reproduces it with fresh evidence.

arbeithand:
Closes #1070

## Summary

Adds a verification step to the release-publishing path that inspects every binary in the goreleaser build matrix and asserts it was built from a clean tree at the tag's target commit — before any artifact is archived, checksummed, or published. This guards against a recurrence of the `v0.0.75` incident, where the published artifact was silently built from a dirty working tree (`vcs.modified=true`), undetectable except via `go version -m`.

## What changed

- **`tools/verify-release-artifact/`** — new standalone Go program (mirrors the existing `tools/print-plugin-hash/` pattern). Reads a built binary's embedded VCS build info via `debug/buildinfo.ReadFile` (works cross-arch without executing the binary — critical since CI is `ubuntu-latest` and must inspect darwin binaries) and asserts `vcs.modified != "true"` and `vcs.revision == <expected>`. On failure it names the offending binary, its actual `vcs.revision`/`vcs.modified`, and the expected revision, then exits non-zero.
- **`.goreleaser.yaml`** — builds the verifier once, natively, via a global `before.hooks` entry (into `.release-tools/`, not `dist/` — goreleaser requires `dist/` to be empty entering the build phase, confirmed empirically). Each `builds[].hooks.post` then invokes that native binary per target, before archiving.
- **`.github/workflows/release.yml`** — exports `TAG_COMMIT` (already computed by the existing "Verify tag is on main" step) via `$GITHUB_ENV` so the goreleaser hook can read it as `{{ .Env.TAG_COMMIT }}`.
- **`.gitignore`** — adds `/.release-tools/`. This was a real self-inflicted bug I caught during implementation: the verifier's own build output, if left untracked in the working tree, would itself make every subsequently-built `fabrik` binary report `vcs.modified=true` — reproducing the exact bug this guard exists to catch.
- **`cmd/version.go`** — `fabrik --version` now appends `+dirty` when the running binary's own build info reports `vcs.modified=true` (the optional surfacing called out in the issue, decided in Plan: implement it). Append-only; no change to ldflags injection.
- **`scripts/cut-release.sh`** — documents the reviewed disposition of the dirty-tree allowlist (`release-notes/<version>.md`, `plugin/known_embedded_versions.go`): both are committed before the tag is pushed, so they can't affect the CI-built artifact. No code change.
- **`docs/USER_GUIDE.md`** — new "Release-artifact verification guard" subsection: what it checks, where it runs, what a failure means, and an explicit note that a trip does **not** mean re-cutting `v0.0.75` (its checksum is already cached by `proxy.golang.org`).
- **`adrs/070-release-artifact-vcs-verification.md`** — records the hook-mechanism design, including two non-obvious traps found during implementation: `go run` inside a per-build hook would try to cross-compile-and-execute the verifier under that target's `GOOS`/`GOARCH` (fails on the runner), and building into `dist/` conflicts with goreleaser's own empty-directory check.

## Testing

- `tools/verify-release-artifact/main_test.go` — unit tests build real binaries in temp git repos via the real `git` binary (`skipIfNoGit` convention), covering clean-checkout-passes, dirty-tree-fails, and wrong-revision-fails.
- `cmd/version_test.go` — new cases for `appendDirtyIfModified`.
- **End-to-end smoke test**, run locally against the actual goreleaser config in a plain clone (matching `actions/checkout`, not a linked worktree — see note in the ADR about a worktree-specific VCS-stamping quirk that produced a misleading result before I switched to a plain clone): `goreleaser build --snapshot --clean` with all 4 real matrix targets, confirming all three required cases —
  - clean commit → all 4 hooks pass, build succeeds
  - dirty tree → all 4 hooks fail with named diagnostics, build aborts before archiving
  - wrong `TAG_COMMIT` → all 4 hooks fail with named diagnostics, build aborts before archiving
- `go build ./...`, `go vet ./...`, `go test -race ./...` — clean except two pre-existing, network-dependent test failures in `cmd/...` (GraphQL calls unavailable in this sandbox), confirmed unrelated by reproducing them identically with my changes stashed out.

## Known gap

The Plan called for a short cross-reference note in `.claude/skills/cut-release/SKILL.md`. That edit was denied by a harness permission policy protecting `.claude/skills/` files from agent modification, so it was skipped rather than worked around. The equivalent (and more complete) guidance is in `docs/USER_GUIDE.md`, which is the canonical documentation the issue actually requires.


arbeithand:
*branch: fabrik/issue-1070 | commit: 7ff39236 | main: 958a339e198ecb2c5e552369b985e03f5ec4997a | 2026-07-26 19:20 UTC*

Working tree is clean and all commits are pushed.


Added `tools/verify-release-artifact`, wired into `.goreleaser.yaml`'s build hooks, to verify every release binary is built clean at the tag's commit before publish — with a matching `+dirty` surfacing in `fabrik --version`, allowlist review, ADR-070, and USER_GUIDE.md docs. Verified end-to-end locally via `goreleaser build --snapshot` against all three required cases (clean/dirty/wrong-revision), and caught a self-inflicted bug during testing (the verifier's own build output needed gitignoring, or it would have caused the exact `vcs.modified=true` bug it exists to catch). All work committed and pushed; `go build`/`vet`/`test -race` clean except two pre-existing, unrelated network-dependent test failures.

---
Used 30/50 turns, 0k input / 18k output tokens.

arbeithand:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-1070 | commit: a5d8224f | main: 2a63727a6af058e5cb1cf740c02d173a631be335 | 2026-07-26 19:53 UTC*

PR is clean and mergeable. Here's the review output.

## Review Findings

### Fixed
- **Issue**: ADR-070 collided with the already-merged `adrs/070-bot-service-notice-filter.md` — this branch's ADR was drafted before that file landed on `main`, and the rebase onto latest `main` surfaced the duplicate number. **Fix**: renamed to `adrs/071-release-artifact-vcs-verification.md`, updated the in-file title, and fixed the two remaining path references in `.goreleaser.yaml` and `scripts/cut-release.sh`.
- **Issue**: The new ADR claimed the guard's distinction from `cut-release.sh`'s own gate was "documented in `docs/USER_GUIDE.md` and `.claude/skills/cut-release/SKILL.md`," but the `SKILL.md` edit was actually skipped (Implement's own PR notes say it was denied by harness permission policy). I independently re-attempted the edit and confirmed the same denial. **Fix**: corrected the ADR's Consequences and Related Work sections to accurately describe the skip instead of a change that never happened.

### Verified
- Rebased onto latest `main` already (done prior to this session); `git diff origin/main..HEAD` shows only the expected 11 files.
- `go test -race -timeout 5m ./...`: full suite passes except two pre-existing, network-dependent failures in `cmd` (`TestExecute_ConfigYAMLApplied`, `TestExecute_ValidStagesReachesEngine` — both fail on a GraphQL board-fetch call unavailable in this sandbox). Confirmed unrelated: this PR touches neither `cmd/root.go` nor `cmd/root_test.go`, and an earlier one-off `FAIL` in `engine` on this run did not reproduce on rerun (flaky, unrelated to this diff).
- `gofmt -l` clean on all new/changed Go files.
- `docs/llms-full.txt` regeneration verified in sync (`bash scripts/generate-llms-full.sh` produces byte-identical output).
- Read through `tools/verify-release-artifact/main.go`, its tests, `.goreleaser.yaml`, `.github/workflows/release.yml`, `cmd/version.go`, and the ADR in full — the design matches every requirement in the issue: per-target check via `debug/buildinfo.ReadFile` (no execution of cross-arch binaries), `TAG_COMMIT` computed once and passed identically to all four hook invocations, named diagnostics on failure, loud non-zero abort before archiving/publishing, and a documented (not silently patched) allowlist disposition for `cut-release.sh`.
- `cmd/version.go`'s `+dirty` addition doesn't destabilize the existing `dev`-prefix auto-upgrade gate or `SemverGreater` — confirmed `SemverGreater` already strips `+`-suffixes (fixed under #1074, prior art correctly cross-referenced in the new ADR).
- PR is `MERGEABLE` / `CLEAN` against `main`.

Reviewed the release-artifact VCS verification guard. Found and fixed an ADR numbering collision (070 was already taken by a merged ADR) by renumbering to 071, and corrected a factual inaccuracy in the ADR claiming a `SKILL.md` doc update happened when it was actually skipped. Build, vet, and race-tested test suite are clean aside from two known pre-existing network-dependent test failures unrelated to this change. PR #1105 is mergeable and ready for human review.

---
Used 9/50 turns, 0k input / 3k output tokens.

arbeithand:
*branch: fabrik/issue-1070 | commit: a5d8224f | main: 2a63727a6af058e5cb1cf740c02d173a631be335 | 2026-07-26 20:00 UTC*

Working tree clean, local matches remote. Everything checks out — all requirements verified, tests pass (one unrelated pre-existing network-dependent failure), docs and ADR are thorough and accurate. Ready to sign off.

## Validation Report

### Requirements: 9/9 passed
- [x] Verify the artifact, not the tree — `tools/verify-release-artifact` uses `debug/buildinfo.ReadFile` on the built binary, not `git status`.
- [x] Checks every binary in the matrix — wired via `builds[].hooks.post` in `.goreleaser.yaml`, runs once per target (darwin/linux × arm64/amd64).
- [x] Asserts revision matches the tag — `TAG_COMMIT` computed once in `release.yml`'s "Verify tag is on main" step, exported via `$GITHUB_ENV`, passed identically to all 4 hook invocations.
- [x] Fails loudly before publish — per-build hook failure aborts the whole `goreleaser release` invocation (archiving/checksumming/GitHub Release/Discussions never run); verified empirically per ADR-071's Testing section.
- [x] Failure output names binary path, actual `vcs.revision`/`vcs.modified`, and expected revision — confirmed in `main.go`'s error messages.
- [x] Guard runs in the publishing path, not just `cut-release.sh` — lives in `.goreleaser.yaml`/`release.yml`.
- [x] Clean release publishes unchanged — no new manual steps for the success path.
- [x] `cut-release.sh` allowlist reviewed and documented in-repo (both allowlisted files are committed before the tag is pushed, so CI's fresh checkout never sees them).
- [x] Docs updated — `docs/USER_GUIDE.md` §`/cut-release` describes what the guard checks, where it runs, and explicitly notes a trip doesn't mean re-cutting v0.0.75; `docs/llms-full.txt` regenerated and confirmed byte-identical to a fresh run.
- [x] Test coverage — `tools/verify-release-artifact/main_test.go` covers all three required cases (clean/dirty/wrong-revision) via real temp git repos + real builds; `cmd/version_test.go` covers `appendDirtyIfModified`.
- [x] Optional `+dirty` surfacing — implemented in `cmd/version.go`, applies unconditionally (release or dev builds) based on the running binary's own `vcs.modified`, no change to ldflags injection.

### Test Suite: PASSED (one unrelated pre-existing failure)
- `go build ./...` — clean
- `go vet ./...` — clean
- `gofmt -l` — clean on every file this PR touches (pre-existing formatting drift exists elsewhere in the repo, untouched by this branch)
- `go test -race -timeout 5m ./...` — 2551 passed, 1 skipped, 1 failed (`cmd.TestExecute_ConfigYAMLApplied`, fails on a GraphQL board-fetch call unavailable in this network-restricted sandbox; this branch touches only `cmd/version.go`/`cmd/version_test.go`, not `cmd/root.go`/`cmd/root_test.go` — confirmed unrelated, matches Review's prior finding)

### Branch state
- Rebased onto latest `main` (already up to date, no conflicts)
- Working tree clean, all commits pushed
- PR #1105: `MERGEABLE` / `mergeStateStatus: CLEAN`, no ADR-number collision (071 is correctly free)

### Verdict: READY TO MERGE

All 9 requirements verified against the issue spec: per-binary VCS guard wired into goreleaser's build hooks, loud pre-publish abort with named diagnostics, `cut-release.sh` allowlist reviewed and documented, docs updated, and the optional `+dirty` `--version` surfacing implemented — all backed by passing unit tests. Build/vet/gofmt clean; full race-tested suite passes except one pre-existing, network-dependent, unrelated test failure. PR #1105 is mergeable and ready for human merge.

---
Used 26/50 turns, 15k input / 5k output tokens.

PR #1115, opened 2026-07-26
Annotate turn-cap retry successors with a provenance notice

### LINKED ISSUE #1081: A turn-cap kill produces an untrustworthy retry: the successor reports work it never performed
## Summary

When a stage exceeds `max_turns`, Fabrik SIGKILLs it and retries after a cooldown. The retry starts
against a worktree where the previous run's file changes already exist — and it can post a
**complete, confident stage report describing verification work it never performed**.

This is worse than the wasted spend, because the output is indistinguishable from a real one.

## Observed

A Review stage was killed at the cap:

```
[#NNN claude] used 51 turns, $5.17
[#NNN stats] used 51/50 turns, 0k input / 22k output tokens
[#NNN warn] claude invocation issue: claude exited with error: exit status 1
[#NNN wait] stage "Review" did not complete — will retry after 5m0s
```

The retry then "completed" in **2 turns / 4k output tokens** and posted a full review write-up
containing a gate table:

| Gate | Result |
|---|---|
| `npm run check` | 0 errors, 48 warnings |
| `eslint` | clean |
| `prettier --check` | clean |
| `npm run test:unit` | 2215 passed / 3 failed |

Four separate commands, each of which takes a turn, reported from a two-turn run. It also concluded
**no defects found, no commits pushed** — while sibling stages that completed normally the same
evening each found a real bug (one a build-breaking export, one a silent data-loss path).

Whether those numbers were carried from the killed session or fabricated, a reader of the PR comment
has no way to tell. Nothing in the posted output indicates the predecessor died.

## Why this is structural

The failure is invisible at exactly the moment it matters:

- The stage is marked **complete** and `Success: true` in `history.json`.
- The PR comment reads like every other review.
- The only trace is two log lines and a `TurnsUsed`/`MaxTurns` pair a human must go looking for.
- Downstream, `cruise` advances to Validate and the PR presents as reviewed.

The retry mechanism is right — the problem is that its output carries no provenance.

## Proposed mechanism

Make the discontinuity visible, without judging the retry's content:

1. **Annotate the posted stage output** when the immediately preceding run for the same
   issue+stage was turn-capped. One line is enough: *"⚠️ The previous attempt at this stage was
   stopped at its turn limit (51/50). This run resumed after that interruption; treat any claim
   about locally-executed commands accordingly."*
2. **Record it in `history.json`** as an explicit field (e.g. `AfterCappedRun: true`) so the
   condition is queryable rather than inferable from adjacent rows.
3. **Optionally, flag the suspicious shape**: a run that completes in very few turns immediately
   after a capped predecessor is the exact signature — worth a log warning at minimum.

Deliberately *not* proposed: refusing the retry, or auto-raising `max_turns`. The retry is often
correct and cheap; the ask is only that its provenance travel with its output.

## Related

The turn-cap kill itself is documented behaviour and sizing `max_turns` from `history.json` is the
right operator response (raising the budget removed the kills here). This issue is specifically
about the *trust* problem in the retry's report, which survives any budget choice — a cap can always
be hit by an unusually large diff.

## Environment

- fabrik 0.0.75
- Review stage at the shipped `max_turns: 50`; 4 of 10 runs capped-and-died before raising it

arbeithand:
Closes #1081

## What this does

When a stage hits `max_turns`, Fabrik SIGKILLs it and retries on a later poll — resuming the same Claude session against a worktree that already has the killed run's partial work. Today the retry's posted comment is indistinguishable from a normal, from-scratch completion: nothing signals that the predecessor died mid-stage, so a reviewer (or `cruise` auto-advancing) can't tell whether reported verification (test runs, lint, etc.) was actually performed in this invocation or just echoed from the truncated prior session.

This PR makes that discontinuity visible without judging the retry's content:

- **`StageState.LastRunCapped map[string]bool`** (new, stage-scoped) records whether the most recently finalized invocation of a given stage was turn-capped: `usage.MaxTurns > 0 && usage.TurnsUsed >= usage.MaxTurns && !completed` — deliberately independent of `err`, since the real-world turn-cap kill exits non-zero and would be missed by the existing `err == nil`-gated extension-loop check. Stage-scoped so an intervening comment-processing invocation (or a different stage's own run) can't clobber the signal.
- **Comment annotation.** When the immediately preceding invocation of a stage was capped, the successor's posted comment (issue and PR paths both) is prefixed with a one-line provenance notice. No turn numbers, no judgment of the retry's findings — just a heads-up that verification claims may be inherited.
- **`history.json` field.** `InvocationRecorded.AfterCappedRun` mirrors the same flag through to `tui.JobCompletedEvent`/`tui.HistoryEntry`, rendered as a dim `⚠cap` marker in the History pane (and picked up for free by `fabrik watch`).
- **Two log-only heuristics** (exploratory, never gate behavior): a successor that completed in under a quarter of the stage's turn budget immediately after a capped predecessor, and stage output that names the "inherited a prior session" situation in its own language.

Scoped to full-stage invocations only — the structurally identical comment-processing retry path (`comment_max_turns`) is a deliberate, separate follow-up (a human is already reading the thread live there).

## Key files

- `internal/itemstate/{itemstate,mutation,store,snapshot}.go` — new mutation, storage, and reader
- `engine/item.go` — read-before-dispatch / write-after-finalize wiring, annotation prepend, log heuristics
- `engine/claude.go` — annotation text and self-admission-language detector
- `engine/observers.go`, `tui/{events,history,model}.go` — TUI/history plumbing
- `docs/state-machine.md` (new "Turn-Cap Retry Provenance" subsection), `docs/USER_GUIDE.md` (History icon table), `docs/llms-full.txt` (regenerated)
- `adrs/071-turn-cap-retry-provenance.md`

## How to test

- `go test -race ./internal/itemstate/... ./engine/... ./tui/...` — all new/updated tests pass, including multi-run capped-chain and intervening-comment-invocation scenarios (`engine/capped_run_test.go`).
- Manually: configure a stage with a low `max_turns`, let an invocation hit the cap, then let the cooldown retry fire — the retry's posted comment should carry the provenance notice, and the History pane should show `⚠cap` on that row.

### DIFF
--- a/adrs/071-turn-cap-retry-provenance.md
+++ b/adrs/071-turn-cap-retry-provenance.md
@@ -1,0 +1,73 @@
+# ADR 071: Turn-Cap Retry Provenance
+
+**Date**: 2026-07-26
+**Status**: Accepted
+**Issue**: #1081 — A turn-cap kill produces an untrustworthy retry: the successor reports work it never performed
+
+## Context
+
+When a stage hits `max_turns`, the Claude CLI's own `--max-turns` self-termination exits non-zero (`completed=false`, `err != nil`). This falls through the progress-based extension loop (ADR-030) — which requires `err == nil` — straight to the cooldown/retry path: a later poll re-invokes the same stage with `resume=true`, against a worktree and Claude session that already carry the killed run's partial work.
+
+The retry mechanism itself is correct and often cheap. The problem is trust: the successor's posted comment is indistinguishable from a normal, from-scratch completion. In the reported incident, a Review stage was killed at 51/50 turns and its retry "completed" in 2 turns with a full gate table (`npm run check`, `eslint`, `prettier`, `npm run test:unit`) — four commands, each costing a turn, reported from a two-turn run. A second, independent installation corroborated the same pattern concentrated in Implement (6 of 19 runs capped, all `Completed: false`), including cases where the successor's own summary named the situation outright ("Found the implementation already complete from a prior interrupted session") while still reporting verification as first-hand. Nothing in the posted output, `history.json`, or the board state signals that a predecessor died mid-stage.
+
+## Decision
+
+**1. Stage-scoped capped-run signal, not item-scoped.** `StageState.LastRunCapped map[string]bool` (keyed by stage name) records whether the most recently finalized invocation of that stage was turn-capped. It is written via a new `StageCappedRunRecorded` mutation, applied unconditionally in `finalizeStageOutcome` on every finalized invocation — so a normal completion clears a stale flag, and a chain of consecutive capped runs (the reported `51/51/23/18/18` case) keeps annotating each successor, not just the first.
+
+**2. Detection condition drops `err == nil`.** The flag is set from `usage.MaxTurns > 0 && usage.TurnsUsed >= usage.MaxTurns && !completed` — independent of `err`. Reusing the extension loop's `hitLimit` condition (which requires `err == nil`) would silently never fire for the real-world turn-cap kill, since that process exits non-zero.
+
+**3. Comment-only annotation; nothing injected into the resumed session's prompt.** `processItem` reads `snap.StageCapped(stage.Name)` at dispatch time — before the current invocation runs — and when true, `finalizeStageOutcome` prepends a one-line callout to the posted output (shared by both the issue-comment and PR-comment paths):
+
+> ⚠️ **Provenance notice:** the previous attempt at this stage was stopped after hitting its turn limit. This run resumed that session — treat any claims about freshly-run commands or verification accordingly.
+
+The annotation carries no turn numbers and is skipped when `postOutput` is empty (degenerate output is never annotated).
+
+**4. `history.json` field, not just a log line.** `InvocationRecorded.AfterCappedRun` mirrors the same pre-dispatch flag into `ItemState.LastInvocationAfterCappedRun`, forwarded by `InvocationObserver` into `tui.JobCompletedEvent`/`tui.HistoryEntry` — rendered as a dim `⚠cap` marker in the History pane and available to `fabrik watch` for free.
+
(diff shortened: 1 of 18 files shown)


arbeithand:
*branch: fabrik/issue-1081 | commit: b6c44cba | main: 2a63727a6af058e5cb1cf740c02d173a631be335 | 2026-07-26 19:57 UTC*

Working tree is clean and no PR exists yet. Everything is implemented, tested, documented, and pushed. Time to signal PR creation and completion.


Implemented turn-cap retry provenance per the plan: a stage-scoped `LastRunCapped` signal drives a one-line comment annotation and a new `AfterCappedRun` `history.json`/TUI field on the invocation immediately following a turn-capped predecessor, plus two log-only "suspicious shape" heuristics. All 17 plan tasks are complete — code, tests (`go test -race` clean across `internal/itemstate`, `engine`, `tui`), `docs/state-machine.md` and `docs/USER_GUIDE.md` updates, regenerated `docs/llms-full.txt`, and new `adrs/071-turn-cap-retry-provenance.md` are committed and pushed.

---
Used 16/50 turns, 0k input / 7k output tokens.

arbeithand:
Closing unmerged — #1081 has been resolved as **not a bug**.

This PR builds a provenance disclaimer for behaviour that is working as designed. A turn-capped stage is retried with `--resume` (`engine/item.go:682` → `engine/claude.go:480-482`), so the successor is the *same* Claude session continuing, not a fresh agent inheriting a dirty worktree. A retry that reports a gate table in two turns is summarising commands that session ran itself before the cap — continuation, not fabrication.

Three specific problems if this landed:

1. The annotation would prefix roughly a third of stage comments (6 of 19 Implement runs were capped in the reporter's second installation) with a notice implying their verification may be fabricated.
2. The few-turns-after-a-capped-predecessor heuristic is the signature of a resume *succeeding* — it would warn on the healthy path every time.
3. The phrase-list heuristic flags output mentioning "prior session" / "interrupted session", penalising exactly the runs that transparently disclose the resume. ADR-071 cites one such honest self-report as evidence of untrustworthiness.

**Salvageable and worth re-filing separately:** the `AfterCappedRun` field in `history.json` plus the `⚠cap` History-pane marker. Queryable provenance with no editorialising is genuinely useful; the prose annotation and both heuristics are what should go.

The branch is retained, so that piece can be cherry-picked rather than rewritten.

Separately, this PR's review surfaced a real defect: `engine/claude.go:481` silently drops `--resume` when the session file is missing or empty, so the retry starts fresh against a predecessor's worktree with nothing logged. That is being filed on its own.

PR #1116, opened 2026-07-26
docs(stage-lifecycle): explain turn-cap retry session resumption

arbeithand:
Documents what happens when a turn-capped stage is retried — the gap that led to #1081 being filed as a bug.

`docs/stage-lifecycle.md` previously mentioned `--resume` in a single line and never explained the implication: the retry **continues the same Claude session**, so it can legitimately summarise commands its predecessor ran before the cap, and a very short retry is the expected shape of a resume working rather than evidence of fabricated output. Both of those read as defects from the posted comment alone.

Also documents the one real failure mode: `--resume` is silently dropped when the session file is missing or empty (`engine/claude.go:480-484`), leaving a fresh session on a worktree full of a predecessor's changes with nothing logged.

Docs-only. `docs/llms-full.txt` regenerated in the same commit per CLAUDE.md.

Related: #1081 (closed as not-a-bug), PR #1115 (closed unmerged).

https://claude.ai/code/session_01J4L5oQDHChWDF3c6oj1haw

### DIFF
--- a/adrs/002-github-as-state-store.md
+++ b/adrs/002-github-as-state-store.md
@@ -46,4 +46,9 @@ Fabrik maintains local ephemeral state only:
 - Session files (`~/.fabrik/sessions/`): Claude Code session IDs for resumption.
 - Worktrees (`.fabrik/worktrees/`): Git worktrees for issue isolation.
 
+> **Superseded in part by [ADR-023](023-cwd-relative-runtime-state.md)**: session files and
+> logs moved from `~/.fabrik/` to `<cwd>/.fabrik/` in v0.0.32. The paths above describe the
+> layout at the time of this decision, not current behaviour. The decision this ADR records —
+> GitHub as the sole authoritative state store — is unaffected.
+
 None of this is authoritative — it's a cache that can be safely deleted.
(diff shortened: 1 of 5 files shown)


arbeithand:
@coderabbitai both findings are valid and are fixed in `3be914f0`. Thanks — the first one caught a genuine inaccuracy rather than a wording nit.

**1. Untrimmed length check.** Correct: `len(sessionID) > 0` runs on the raw bytes, so a whitespace-only file passes the guard and then `TrimSpace` yields `""` — meaning `--resume` is appended with an *empty value*, not omitted. My text claimed missing-or-empty always falls back to a fresh session, which is wrong for that third case. The docs now enumerate all three outcomes explicitly, and the addendum on #1117 records that the empty-value case may be a distinct bug from the silent-fallback one, since what the CLI does with an empty `--resume` isn't established yet.

**2. `ReadSessionID` divergence.** Also correct, and a better catch than it first looks: `ReadSessionID` (`engine/claude.go:286-302`) trims before returning and so treats a whitespace-only file as absent, while the inline read in `buildClaudeArgs` does not. `cmd/resume.go` uses the former, the invocation path the latter. Now documented, and #1117 asks whether the invocation path should call `ReadSessionID` outright — with the caveat that it derives its own path from `(repo, issueNumber, stageName)` while `buildClaudeArgs` receives `sessFilePath` pre-built, so it's a small refactor rather than a drop-in.

Your review also surfaced a pre-existing error neither finding named directly: the docs said the session file lives at `~/.fabrik/sessions/...`, but it is `<cwd>/.fabrik/sessions/...` and gains an `<owner>-<repo>` segment on multi-repo projects (`engine/claude.go:227-242`). Corrected in the same commit.

PR #1078, opened 2026-07-25
feat(stages): declarative `unmanaged: true` stage type for parking columns (Backlog)

arbeithand:
Closes #973

## Summary

Introduces a third special-stage category — `unmanaged: true` — mirroring the existing `cleanup_worktree`/`holding_stage` flags. It declares a board column Fabrik recognizes but never dispatches a worker to and never auto-advances items out of. Backlog is now a real (if inert) stage YAML instead of a hardcoded string, generalizing the pattern to any parking column (Icebox, On Hold, etc.) a team might add.

## Key changes

- **`stages.Stage.Unmanaged bool`** (`stages/stages.go`) — new flag; `loadOne`'s prompt/skill requirement is waived when set, mirroring the `CleanupWorktree`/`HoldingStage` guard.
- **`stages/examples/backlog.yaml`** — new default stage (`name: Backlog`, `order: -1`, `unmanaged: true`), auto-installed by `fabrik init` (which already extracts every embedded example file).
- **`engine/startup.go`** — unmanaged stages are unconditionally exempt from the "missing board column" required-set (a board without a `Backlog` column no longer fails startup once `backlog.yaml` is configured); the hardcoded `"Backlog"` recognition in the extra-column check remains as the compat net for installs without `backlog.yaml`.
- **`engine/item.go`** — the three `HoldingStage` dispatch guards (`itemMayNeedWork`, `itemNeedsWork`, `processItem`) now also check `stage.Unmanaged`.
- **`engine/poll.go`** — the catch-up (auto-advance) loop skips unmanaged stages; `runProbeAndDeepFetch`'s stage-membership guard now treats unmanaged stages as "unconfigured" for deep-fetch purposes, preserving the deep-fetch-avoidance optimization from issue #778 (this would have silently regressed once `backlog.yaml` made `FindStage("Backlog")` return non-nil).
- **Docs** — `CLAUDE.md`, `README.md`, `docs/USER_GUIDE.md`, `docs/stage-lifecycle.md`, `docs/state-machine.md` document the new flag and its dispatch-suppression mechanism; `docs/llms-full.txt` regenerated. `USER_GUIDE.md` also documents that `refresh-stages` cannot introduce brand-new files (only patch missing keys into existing ones) — existing installs without `backlog.yaml` must copy it in manually to opt into the declarative form; the hardcoded compat net keeps them silent either way.

## Testing

[...]

PR #1123, opened 2026-07-26
fix(ci): pin Claude PR review to claude-sonnet-5 + honest failure fallback

arbeithand:
## Problem

The Claude PR reviewer regressed to posting a useless stub (`🤖 Automated test-repo review (Claude). No blocking issues surfaced by the agent step.`) on every PR. Two independent causes:

1. **Bad model default.** The workflow didn't pin a model, so `anthropics/claude-code-action@v1` picked its own default — which drifted to **`claude-opus-5[1m]`**, an id this key can't use. Every review run failed with `is_error:true` after 1 turn (confirmed in run logs), so no formal review was posted. Claude *did* post real reviews here before the drift (e.g. #1078), and `verveguy/liminis-context-graph` still works today with the exact config this PR adopts.

2. **Dishonest fallback.** The safety-net posted *"No blocking issues surfaced"* whenever no formal review existed — reporting a **failed or skipped** review as a clean one. That false all-clear is what masked the outage.

## Fix

1. Pin `--model claude-sonnet-5` (matches the known-good liminis-context-graph config; Sonnet is capable for diff review and far cheaper per-PR than Opus).
2. Replace the fallback message with an honest **"Claude review did not post — this is NOT a clean bill of health"** that links the run logs. It still posts a formal review so Fabrik's review gate (`engine/reviews.go`) clears rather than hangs.

## Cannot self-test — validate after merge

`claude-code-action` **skips any PR that modifies its own workflow file** (a security guard against secret exfiltration): *"your workflow will begin working once you merge your PR."* So this PR's own review run is skipped by design. Validation is: **merge, then confirm the next normal PR gets a real Sonnet review.**

## Follow-ups

- `shadoworg/fantasy`'s `claude-review.yml` has **no** model pin either — same one-line fix needed there.
- Open PRs branched before this merges (e.g. #1121) need a rebase to pick up the fixed workflow, else the validation guard skips their review too.

### DIFF
--- a/.github/workflows/claude-review.yml
+++ b/.github/workflows/claude-review.yml
@@ -44,21 +44,32 @@ jobs:
             Then submit a FORMAL pull-request review (not an issue comment) with:
               gh pr review ${{ github.event.pull_request.number }} --comment --body-file <your-findings-file>
             Add inline comments for specific line-level issues where useful.
-          claude_args: "--allowedTools Bash(git:*),Bash(gh:*),Read,Grep"
+          # Pin the model explicitly. Without --model, claude-code-action@v1
+          # picks its own default, which drifted to an unusable id
+          # (claude-opus-5[1m]) and made every review run fail with
+          # is_error:true after 1 turn — the safety-net step below then posted a
+          # useless "no blocking issues" stub. Sonnet is the intended reviewer
+          # (capable for diff review, far cheaper than Opus per-PR).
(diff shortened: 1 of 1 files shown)


arbeithand:
@coderabbitai review

Heads-up: the two prior findings were on commit a8b38316, which has been reverted. That commit's `claude-review.md` file-handoff (and the forged-review risk you flagged) no longer exists — the workflow is back to the agent producing the review in-session and posting via `gh pr review`, matching the known-good liminis-context-graph config, plus an honest failure fallback. Please re-review the current head.

Issue #1183, opened 2026-07-27
Usage-limit detection false-positives on stage output prose, suspending Claude dispatch account-wide


arbeithand:
## Problem

`#1119`'s usage-limit detection matches on **stage output prose**, so a stage whose output merely *mentions* a usage-limit message is misclassified as having hit one. It then suspends Claude dispatch **account-wide**, halting all work.

This fired on its first day live and stopped the pipeline for ~11 hours.

`engine/claude.go:34`:

```go
var usageLimitHitRE = regexp.MustCompile(`(?i)hit your \S+ limit`)
```

## Observed 2026-07-27

```
15:01:06 [#1178 claude-limit] usage-limit exit detected
         (resetTime="10:20pm (America/Edmonton)", turns=51, cost=$2.2766): hit your session limit
15:01:06 [#1178 claude-limit] account usage-limit hit; suspending Claude dispatch
         account-wide until 2026-07-27T22:20:00-06:00 (parsed from CLI reset time)
```

Three independent facts show this was **not** a usage limit:

1. **`turns=51`, `cost=$2.2766`.** A genuine usage-limit exit fails immediately with 0 turns and $0.00. This invocation ran a full budget and did real work.
2. **The invocation's own terminal state was a turn cap.** From its stage log:
   ```
   "terminal_reason":"max_turns"
   "subtype":"error_max_turns"
   ```
3. **The "parsed" reset time was `10:20pm (America/Edmonton)`** — the verbatim example string from #1084's issue body. The operator is not in that timezone. The regex matched text the stage had *written*, not text the CLI had emitted.

Confirmed mechanism: that stage log contains `hit your session limit` **4 times** and `America/Edmonton` **5 times**, because #1178's Implement stage was writing code and tests about usage-limit detection and quoting #1084's example as a fixture.

## Why this is severe

- **Blast radius is account-wide, not per-issue.** One false positive suspends *every* repo and every issue this engine manages, for as long as the bogus reset time says (here ~11 hours).
- **Fabrik is uniquely prone to it.** This repository's issues, specs, ADRs, tests and skill files are full of these exact strings — #1084, #1119, #1120, #1146, #1178 all quote them. Any stage touching that work self-triggers.
- **It looks authoritative.** The log reports a parsed reset time, so an operator reasonably believes a real limit was hit.
[...]

Issue #1178, opened 2026-07-27
TUI renders turn-budget exhaustion as "(error)" when it is an incomplete, resumable run


arbeithand:
## Summary

The TUI History pane (and `fabrik watch`, if it renders this field) renders a turn-budget exhaustion identically to a genuine invocation failure: a red `✗` with `(error)`. A turn-cap exit is not a fault — it is an incomplete, resumable run that made real progress and will continue in the same Claude session on the next attempt. Reading several capped runs as red `(error)` rows drives a wrong diagnosis ("failed attempts burning money" instead of "one session making incremental progress across budget windows"), which implies the opposite remedy (investigate a fault vs. simply raise the budget). The two cases need visually distinct renderings, discriminated using the Claude CLI's own structural result classification rather than inferred from turn counts.

## Requirements

- A run that ended by exhausting its turn budget must render distinctly from a genuine error, in both the TUI History pane and `fabrik watch` (if it renders this field).
- Suggested rendering: `↻` with `(turn limit)` — reusing the existing non-alarming retry style rather than introducing a new color. `✗ (error)` remains reserved for genuine faults.
- Detect the turn-cap case from the Claude CLI's own structural result classification — `subtype: "error_max_turns"` (and/or the more explicit `terminal_reason: "max_turns"`) — not by inferring it from turn counts (`MaxTurns > 0 && TurnsUsed >= MaxTurns && !Completed`). The turn-count heuristic is workable but fragile: the CLI's own accounting can report a turn count past the configured cap (observed: `num_turns: 51` against `max_turns: 50`), so an inference built on `>=` is riding on slightly odd arithmetic, and it can't cleanly distinguish "hit the cap" from "legitimately used the last turn and finished." The CLI's own classification has neither problem.
- The structural classification must be threaded from the parsed Claude response through to whatever state `history.json` / the TUI / `fabrik watch` currently read to make this decision (today: `LastInvocationErrored` → `Success`), so the distinction is available wherever that state is read, not just re-derived locally in the TUI.
- A genuine non-zero exit that is *not* a turn cap must still render as `✗ (error)`.
- Decide **where** to fix it, and state the reasoning:
  - **TUI-side** (`tui/history.go`): reorder so the turn-cap check precedes `!Success`. Smallest change, but leaves the underlying `Success: false` misclassification in place for any other consumer.
[...]

Issue #1119, opened 2026-07-26
Detect Claude usage-limit exhaustion and stop misattributing it as a stage failure


arbeithand:
## Problem

When the underlying Claude session hits its account usage limit, the CLI exits immediately with a message naming the reset time:

```
You've hit your session limit · resets 10:20pm (America/Edmonton)
```

Fabrik has **no detection for this condition anywhere**. The invocation is treated as an ordinary stage failure, which produces three distinct harms:

**1. The retry budget is consumed by a condition that is not a failure.** `engine/item.go:1137-1152` sets `claudeRan` to `false` only for process *start* failures (`exec.Error`, `os.PathError`, `exec.ErrNotFound`). A usage-limit exit is a process that started fine and exited non-zero, so `claudeRan = true`, `StageRetryIncremented` fires (`:1279`), and `max_retries` is spent. A subsequent genuinely-transient failure then has no budget left.

**2. The retries cannot succeed.** The cooldown is `PollSeconds * 10` (`engine/item.go:1276`) — 5 minutes at the default `poll: 30`. All three retries burn within ~15 minutes of a limit that resets in *hours*.

**3. The issue is durably mislabelled.** The item ends with `stage:<Stage>:failed`, `fabrik:paused`, and a comment reading *"stage `<Stage>` failed 3 time(s) — pausing issue"*. Nothing distinguishes that from a branch whose stage genuinely failed three times. As the reporter put it: **a failure attributed to the wrong cause is worse than a visible outage.** Later readers — and Fabrik's own later stages, which re-read the issue — draw the wrong conclusion.

Because the limit is **account-wide**, every concurrent worker hits it in the same window, so each in-flight issue independently burns its own retry budget and pauses. With `max_concurrent: 4` that is four issues needing manual recovery.

Reported by @bdueck in #1084 with `history.json` evidence (`1/150 turns, $0.00, failed`, twice per issue across two issues).

## Summary

Detect the usage-limit exit, exclude it from the `max_retries` budget, and label/report it as what it is rather than as a stage failure. **Backoff-to-reset and account-wide dispatch suspension are deliberately out of scope** — they are the follow-up issue, which this one blocks.

## Requirements

[...]

Issue #1084, opened 2026-07-26
Quota exhaustion is misclassified as a stage failure: burns max_retries, mislabels the issue, and cascades across concurrent workers


bdueck:
## Summary

When the underlying Claude session hits its usage limit, the invocation exits immediately with a
message that **states when it resets**:

```
You've hit your session limit · resets 10:20pm (America/Edmonton)
```

Fabrik treats this as an ordinary stage failure. It retries on the normal cooldown, exhausts
`max_retries`, escalates, and pauses the issue with `stage:<Stage>:failed`.

Three things go wrong, in increasing order of cost.

## 1. The retries cannot possibly succeed

From `history.json`, one stage across two issues:

| Issue | Stage | Turns | Cost | Result |
|---|---|---|---|---|
| #A | Review | 1/150 | $0.00 | failed |
| #A | Review | 1/150 | $0.00 | failed |
| #B | Review | 1/150 | $0.00 | failed |
| #B | Review | 1/150 | $0.00 | failed |

Each retry fires within minutes of a limit that resets in **hours**. The reset timestamp is right
there in the error text, so the condition is not merely detectable — its clearing time is *known*.

## 2. It consumes the retry budget that exists for real failures

`max_retries: 3` is the runaway backstop for genuinely flaky stages. Quota exhaustion spends all
three against a condition that is not the stage's fault and cannot clear. After the limit resets,
the issue is paused with its retries already gone — so a subsequent *real* transient failure has no
budget left.

## 3. It cascades, and it lies in the issue history

The usage limit is **account-wide**, so with `max_concurrent > 1` every in-flight worker hits it in
the same window. Two issues here; with four workers it would be four. Each independently burns its
own retry budget and pauses.

The durable damage is the labelling. Both issues now carry `stage:Review:failed` and a timeline
reading *"stage Review failed 3 time(s) — pausing issue"*. **Nothing distinguishes that from a
branch whose review genuinely failed three times.** We had to hand-write comments on both issues
explaining that the failures were quota, not code — otherwise the next person to read them (or the
next Fabrik stage, which re-reads the issue) would draw exactly the wrong conclusion.

That is the part worth fixing even if the retry behaviour stays: **a failure attributed to the wrong
cause is worse than a visible outage.**

## Proposed mechanism

Recognise the condition and give it its own state, rather than folding it into stage failure:

1. **Detect** the usage-limit exit — it is a distinctive message on an invocation that consumed ~1
[...]

PR #1184, opened 2026-07-27
fix(engine): don't classify a turn-capped run as a usage-limit exit

arbeithand:
Refs #1183 — this is a targeted stop-gap; the issue stays open for the remaining Definition of Done items (structural detection, reset-time trust, restart-free suspension clearing).

## The bug

`detectUsageLimitExit` matches a regex against **raw invocation output**, which necessarily includes text the assistant itself wrote. So any stage whose output merely *discusses* usage limits self-triggers an account-wide Claude dispatch suspension.

That is not hypothetical in this repository. On 2026-07-27, #1178's Implement stage — writing tests *about* usage-limit detection, quoting #1084's example message as a fixture — exited on a plain turn cap:

```json
"subtype": "error_max_turns", "terminal_reason": "max_turns", "num_turns": 51
```

Fabrik classified it as a usage-limit hit and suspended dispatch **for every repo and issue** until the reset time it parsed out of that quoted fixture — `10:20pm (America/Edmonton)`, a timezone the operator isn't in. Roughly **11 hours** of pipeline downtime, caused by a turn cap.

Three facts made it unambiguous: 51 turns and $2.28 consumed (a real limit exit does 0 turns, $0.00), the invocation's own terminal state was `max_turns`, and the "parsed" reset time was verbatim from an issue body.

## The fix

Pass `TokenUsage` as an **exclusion gate**. An invocation that consumed turns *and* incurred cost cannot have exited on a usage limit, because such an exit terminates immediately, before any work is billed.

**The asymmetry is the point.** #1119's spec explicitly forbids using the "~1 turn, $0.00" shape to *confirm* a usage limit — a genuine immediate stage failure shares that shape, so that would risk false negatives. Using the *inverse* shape to *rule one out* is sound: no real usage-limit exit does 51 turns of work first. Ruling in and ruling out are different inferences, and only the former was prohibited. The existing call-site comment is updated to say so, since the distinction is easy to "correct" back.

The exclusion is conjunctive (turns **and** cost) so a partially-recorded usage struct can't suppress a genuine detection.

## Tests

Three cases added to the existing table, plus threading `usage` through it:

- **Regression:** turn-capped run quoting the message in its own output → not detected
- **Conjunctivity:** cost recorded but zero turns → still detected
- **Canonical:** zero turns, zero cost → still detected

`go build`, `go vet` clean; `go test -race ./engine/ ./internal/...` — **1,856 tests pass**.

## Scope

Deliberately a targeted stop-gap, to unblock the pipeline. The durable fix is to classify from the CLI's structured result (`subtype` / `terminal_reason`, neither of which `claudeResponse` captures today) rather than from output prose — tracked in #1178 and in #1183's requirements.

Note the suspension is in-memory (`engine/engine.go:119-120`), so restarting the engine clears a false positive; #1183 also asks for a way to clear it without a restart.

https://claude.ai/code/session_01J4L5oQDHChWDF3c6oj1haw

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -4880,9 +4880,25 @@ parsed `claudeResponse.Result` field, since it is unconfirmed whether a real occ
 in `claudeResponse.Errors[]`, or in non-JSON plain text — for Anthropic's usage-limit exit message via
 `detectUsageLimitExit()`. When Claude exits non-zero without a `FABRIK_STAGE_COMPLETE` marker and the
 message matches, `interpretClaudeResult` returns a `*claudeUsageLimitError{Message, ResetTime}`
-sentinel instead of the generic error wrapper. The turn-count/cost shape (`~1 turn, $0.00`) is logged
-alongside purely for diagnostic corroboration and never gates detection — a genuine immediate stage
-failure can share the same shape.
+sentinel instead of the generic error wrapper.
+
+The turn-count/cost shape gates detection in one direction only. It is **never** used to *confirm* a
+usage limit — a genuine immediate stage failure shares the `~1 turn, $0.00` shape, so treating that
+shape as positive evidence would produce false negatives on real failures. It **is** used to *exclude*
+one: `detectUsageLimitExit` returns `detected = false` when `usage.TurnsUsed > 0 && usage.CostUSD > 0`,
+because a real usage-limit exit terminates the invocation immediately, before any work is billed, so an
+invocation that consumed turns *and* incurred cost cannot have hit one. Ruling in and ruling out are
+different inferences; only the former is unsound here.
+
+The exclusion is conjunctive (turns **and** cost) so a partially-recorded usage struct cannot suppress
+a genuine detection.
+
+This guard exists because the message is matched against **raw invocation output**, which necessarily
+includes text the assistant itself wrote. Without it, any stage whose output merely discusses usage
+limits self-triggers an account-wide suspension — on 2026-07-27 a turn-capped run on #1178 that quoted
+#1084's example message suspended dispatch for ~11 hours (#1183). Matching prose remains the underlying
+weakness; classifying from the CLI's structured result (`subtype` / `terminal_reason`) is the durable
+fix and is not yet implemented.
 
 **Handling:** `finalizeStageOutcome()` (`engine/item.go`) detects the sentinel via `errors.As`,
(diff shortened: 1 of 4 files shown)


arbeithand:
All three addressed in `27684ea7`. Thanks — finding 3 was a real hole, not a nit.

**1. Stale canonical doc.** Correct, and a straight miss on my part: CLAUDE.md is explicit that `docs/state-machine.md` moves in the same change set. The passage now states the asymmetry directly rather than the old blanket claim — never used to *confirm* (a genuine immediate failure shares the `~1 turn, $0.00` shape, so that would risk false negatives), only to *exclude*. It records the #1183 incident as the reason the guard exists and notes that prose-matching remains the underlying weakness with structural classification as the durable fix. `docs/llms-full.txt` regenerated.

**2. `Closes` → `Refs`.** Agreed, and the reasoning is right: #1178 is scoped to TUI rendering and does **not** cover the remaining #1183 bullets, so auto-closing would have dropped structural detection, reset-time trust, and restart-free suspension clearing with no tracking anywhere. PR body updated; #1183 stays open.

**3. The test gap — this one mattered most.** You're right that the unit table hand-supplies `TokenUsage` and therefore cannot catch the field wiring breaking. I checked the specific claim and it holds exactly: my regression fixture had no `total_cost_usd`, so that payload through the real pipeline would parse `CostUSD` as `0`, leave the conjunctive gate **inert**, and reintroduce the false positive — with the test still green. A regression test that can't observe its own regression.

Added `TestInterpretClaudeResult_TurnCappedQuotingMessage_NotUsageLimit`, which drives the full `interpretClaudeResult` path from raw JSON carrying both `num_turns: 51` and `total_cost_usd: 2.2766`, and asserts `TurnsUsed`/`CostUSD` parsed non-zero **before** asserting the sentinel is absent — so if the wiring ever goes inert the test fails loudly rather than passing for the wrong reason.

I also confirmed it isn't vacuous. With the guard removed it fails, reproducing the incident's exact log line:

```
[#1 claude-limit] usage-limit exit detected (resetTime="10:20pm (America/Edmonton)", turns=51, cost=$2.2766): hit your session limit
--- FAIL: TestInterpretClaudeResult_TurnCappedQuotingMessage_NotUsageLimit
    #1183 regression
```

Restore the guard and it passes.

`go build`, `go vet` clean; `go test -race ./engine/` green.

PR #1202, opened 2026-07-27
fix(engine): detect Claude usage-limit exits structurally, not from output prose

arbeithand:
Closes #1183

## What this does

Replaces regex-based usage-limit detection (`usageLimitHitRE` over raw stage output) with a structural check on the CLI's own result object: `resp.TerminalReason == "blocking_limit"`. This closes the false-positive class described in the issue, where a stage merely *writing about* usage-limit phrasing (e.g. quoting a fixture in a test) got misclassified as an actual limit hit and suspended Claude dispatch account-wide.

Also adds two restart-free recovery mechanisms:
- `fabrik:clear-claude-limit` — an operator-applied label (mirrors `fabrik:revalidate`) that a new per-poll settle scan reads to clear an active account-wide suspension immediately, no engine restart needed.
- An account-wide settle scan that removes the informational `fabrik:claude-limit` label from every open item once the underlying suspension lifts, rather than only clearing it on a labelled issue's next dispatched invocation (the orphaned-label leak called out in the issue, same shape as #1135).

## Key changes

- `engine/claude.go`: deleted `usageLimitHitRE`/`usageLimitResetRE`/`detectUsageLimitExit`; added `classifyUsageLimitExit`, wired into `interpretClaudeResult` after the existing turn-cap check. Reset time is no longer parsed from prose — a genuine hit always falls back to the existing 1h backoff. A diagnostic log line records any unmatched, non-empty `terminal_reason` (e.g. `rapid_refill_breaker`, deliberately excluded from triggering a suspension) for future confirmation.
- `engine/claude_limit_settle.go` (new): `settleClaudeLimitClearRequests` and `settleClaudeLimitLabelSweep`, wired into the per-poll settle sequence in `engine/poll.go`.
- Canonical docs (`docs/state-machine.md`, `docs/stage-lifecycle.md`, `docs/USER_GUIDE.md`, `CLAUDE.md`, `docs/llms-full.txt`) updated to match; new ADR `adrs/1183-structural-claude-usage-limit-detection.md`, with short pointer notes added to ADR-1119/1120.

## How to test

- `go test -race ./engine/...` — covers: turn-cap output that quotes limit phrasing is *not* classified as a usage-limit exit (the issue's own regression case), a genuine `blocking_limit` structural exit is classified, the `TurnsUsed>0 && CostUSD>0` sanity guard (#1184) still applies, and both new settle scans (including a `-race` concurrent-activation test).
- `go build ./...`, `go vet ./...` clean.

## Note

[...]

Issue #1189, opened 2026-07-27
feat(pruefer): post findings as inline review thread comments so Fabrik's review-reinvoke can consume them


arbeithand:
## Problem

Pruefer submits its entire review as a single top-level `body` on the `pull_request_review`. `github/prs.go:572-586` (`SubmitPRReview`) posts only `commit_id`, `body`, and `event: COMMENT` — there is no `comments[]` array, so every finding lands as prose in one comment regardless of which file or line it concerns.

V1 scoped inline comments as "desirable but optional" (ADR-1113) and took the option. That was the right call for shipping, but it leaves Pruefer unable to participate in the loop it was built for.

**Fabrik's review-reinvoke path consumes only inline thread comments.** `buildReviewThreadComments` (`engine/reviews.go`) reads line-anchored review comments; it does **not** read the top-level review body. So a Pruefer review today:

- ✅ satisfies `wait_for_reviews: true` — it is a formal `pull_request_review`, so `hasReviews` is true and the gate clears
- ❌ produces **zero** auto-fix work — Fabrik never sees the findings, because they are in the body

Observed live on PR #1185 (2026-07-27, Pruefer's first submitted review). It flagged a real, specific issue — the precedence between the stage-complete carve-out and the turn-limit classification at `engine/claude.go:944-969`, including that `TestInterpretClaudeResult_TurnCapExit_ClassifiedAsTurnLimitError`'s fixture doesn't exercise the edge case. Good finding. Fabrik cannot act on any of it; a human has to read it and relay it by hand.

With inline comments the circuit closes: **Pruefer reviews → Fabrik auto-fixes → Pruefer re-reviews the new head SHA** (which is already handled, since review state is keyed on head SHA). That is the difference between a reviewer that participates in the pipeline and one that only observes it.

## Summary

Extend `SubmitPRReview` to post line-anchored `comments[]` alongside the summary `body` in the same `pull_request_review` request, have the review prompt emit structured findings instead of free prose, map those findings onto diff-validated anchors, and demote any finding that can't be anchored into the summary body — so a single unanchorable finding never causes the whole review to fail.

## Requirements

- Extend `SubmitPRReview` to accept an optional set of line-anchored comments, posted in the same request:

  ```json
  {
    "commit_id": "<sha>", "event": "COMMENT", "body": "<summary>",
    "comments": [{"path": "engine/claude.go", "line": 954, "side": "RIGHT", "body": "..."}]
  }
  ```

[...]

Issue #1071, opened 2026-07-25
[design] De-fragilize the review gate — it single-sources on formal-review bots (Gemini SPOF)


arbeithand:
> **Status: design question — needs further thought before implementation.** Not spec-kit-ready; do NOT pipeline yet. Captured so the analysis isn't lost.

## Problem

Fabrik's `wait_for_reviews: true` gate is **single-sourced on reviewers that submit a *formal* GitHub review**, and that fragility took down the self-hosting pipeline on 2026-07-25 (issues #1058, #1059, #1065 all paused at the review gate) the moment our one formal-review bot (Gemini) ran out of quota.

### Verified mechanics (not memory — read from source)

The gate clears only when both conditions hold (`engine/reviews.go:119`):
```go
if len(outstanding) == 0 && hasReviews { /* gate opens */ }
```
and `hasReviews` (`reviewGateOutstanding`, `engine/reviews.go:271-275`) is true **only if at least one non-DISMISSED `gh.PRReview` object exists** — i.e. a formal `pull_request_review` (APPROVE/COMMENT/REQUEST_CHANGES), the kind that appears in `gh pr view --json reviews`. On final timeout the gate **pauses**, never advances.

### The fragility

- **Gemini** (`gemini-code-assist`) submits a formal `COMMENTED` review — empirically confirmed on ~9 of the last 12 merged PRs (#1036–#1057). That is the *only* reason the pipeline has been advancing. It is a single point of failure.
- **Copilot** would also submit a formal review, but signups are currently closed.
- **Claude Code review does NOT satisfy the gate** — verified, not assumed:
  - Managed "Code Review" (claude.ai admin): posts inline comments + a check run, **no formal review object**.
  - `anthropics/claude-code-action` / the `code-review` plugin: its command source posts via `gh pr comment <PR> --body …` and inline-comment MCP — it explicitly **does not** use `gh pr review` or `POST /pulls/N/reviews`.
  - So pointing Claude review at a PR leaves comments but never flips `hasReviews`; the gate still times out and pauses.

Net: any single-reviewer outage (quota, sunset, signup freeze) silently stalls every issue at Review until a human intervenes.

## The question (needs thought)

How should the review gate be de-fragilized so it isn't hostage to one formal-review bot? Candidate directions, each with tradeoffs to think through — **not yet decided**:

[...]

PR #1211, opened 2026-07-28
yolo items can merge with unresolved review threads: findings arriving near Validate completion are lost

### LINKED ISSUE #1207: yolo items can merge with unresolved review threads: findings arriving near Validate completion are lost
## Problem

A `fabrik:yolo` item can merge with **unresolved PR review threads**, losing review findings that arrived shortly before Validate completed.

Observed on #1183 / PR #1202 on 2026-07-27:

```
21:29:03  claude[bot]              inline finding   → Fabrik reinvoked Review, resolved
21:37:13  handarbeit-pruefer[bot]  inline finding   → Fabrik reinvoked Review, resolved
21:42:10  handarbeit-pruefer[bot]  inline finding   → resolved
21:47:11  Fabrik: Validate stage completes
21:48:56  handarbeit-pruefer[bot]  inline finding   → NEVER SEEN
21:50:07  PR #1202 MERGED
```

The reinvoke loop worked for the first three — each finding triggered a Review reinvoke, was addressed, and its thread was resolved. The fourth arrived **71 seconds** before the merge and was lost. That finding (a dead `else if` branch when `TerminalReason == "blocking_limit"` is excluded by the turns/cost guard) is now on `main`, unaddressed.

### Why the existing mechanisms don't cover it

**§2.9 Review Reinvoke** fires on any `stage:<X>:complete` item with unresolved threads, and it is stage-agnostic — it is what handled the first three. But it only helps while the item is *parked* at a completed stage. With `yolo`, Validate completing does not park anything: `attemptMergeOnValidate` runs in the same flow and advances to Queued (or enables native auto-merge) immediately.

**§2.16 SHA-invalidation** compares `snap.ValidateCompletedSHA()` against `LinkedPR.HeadSHA` and un-completes Validate on mismatch. But **a review comment does not change the head SHA** — the finding was posted against the already-validated head. Nothing to detect.

So the back-edge exists for *"a commit arrived after Validate"* and not for *"a finding arrived after Validate."*

Note `fabrik:cruise` is largely unaffected: it deliberately stops at Validate, so a late finding is picked up by the normal reinvoke loop before a human merges. This is specifically a yolo/auto-advance race.

### Why it is getting worse, not better

#1189 made Pruefer post findings as inline comments, and Pruefer re-reviews on every new head SHA. Fabrik pushes a fix per reinvoke, each producing a new SHA, each triggering a fresh review. **The faster the fix loop runs, the more likely a review lands inside the merge window.** This race is a consequence of the loop working.

## Summary

A `fabrik:yolo` item must not merge while unresolved PR review threads exist on its current head commit. This requires guarding **two** points, not one: the initial decision to advance out of Validate / enable auto-merge (`attemptMergeOnValidate`), and the ongoing window between auto-merge being enabled and GitHub actually completing the merge (`checkAutoMergeConvergence`'s repeated polling). Guarding only the first point would not have caught the incident above — all three earlier findings were already resolved by the time Validate completed, and the fourth arrived after that one-shot decision had already run. If a late finding appears during the convergence window, Fabrik must disable auto-merge before GitHub can merge underneath the reinvoke loop, then let the existing review-reinvoke loop (§2.9) address it and re-enable once resolved.

This is an explicit stopgap, not a general redesign — see Risks / Dependencies.

## Requirements

- **At `attemptMergeOnValidate`:** a yolo item with an unresolved review thread on its current head does not advance out of Validate and does not have GitHub auto-merge/merge-queue enabled on its behalf.
- **At `checkAutoMergeConvergence`:** on each convergence poll, if unresolved review threads have appeared on the current head since auto-merge was enabled, Fabrik disables auto-merge (so GitHub does not merge underneath the reinvoke loop) and lets §2.9 address the findings; auto-merge is re-enabled automatically once they are resolved.
  - This is a real mutation on the PR and must be idempotent, and must not corrupt the convergence state machine. `fabrik:auto-merge-enabled` currently serves as both the idempotency guard for `attemptMergeOnValidate` and the budget-start anchor read by `checkAutoMergeConvergence` — removing and later re-adding it has consequences for that budget beyond just the merge itself, and those consequences must be checked, not assumed to round-trip cleanly.
  - **Sanctioned fallback:** if disabling-and-re-enabling proves unsafe or too disruptive to the convergence budget/state machine, a narrower version is acceptable instead: guard `attemptMergeOnValidate` only, and emit a loud log (not a disable) when unresolved threads appear during the convergence window. Do not force the full disable/re-enable mechanism if it compromises correctness elsewhere — say so and fall back.
- The check is scoped to the PR's **current head SHA** in both places — a thread opened against a commit that has since been superseded by a new push must not block.
- Once a blocking thread is resolved by the existing review-reinvoke loop (§2.9), the item advances / auto-merge is (re-)enabled on a later cycle without further manual intervention.
- `MaxReviewCycles` (default 5) continues to bound the reinvoke loop and pause on exceed — this change must not create a new state that neither advances nor escalates, at either guard point.
- Every deferral is logged explicitly, naming the SHA and the thread count (e.g. "Not advancing: N unresolved review thread(s) on `<sha>`" / "disabling auto-merge: N unresolved review thread(s) on `<sha>`") — silence here is indistinguishable from a stall.
- Reuse existing thread-unresolved detection (`buildReviewThreadComments`, the `reviewThreads { isResolved }` GraphQL data) rather than adding a parallel detection path, at both guard points.

## Scope

**In scope:**
- Closing the race for `fabrik:yolo` items specifically (native auto-merge, merge-queue enqueue, and any other path that makes Validate's completion irreversible).
- The engine-side checks and their interaction with the existing §2.9 reinvoke loop and `MaxReviewCycles`, at both `attemptMergeOnValidate` and `checkAutoMergeConvergence`.
- Disabling an already-enabled auto-merge when a late finding appears during convergence — or, if that proves unsafe (see Requirements fallback), a loud log at that point instead.

**Out of scope:**
- `fabrik:cruise` — already unaffected, since it stops at Validate for human merge.
- Changing repository-level GitHub branch protection (e.g. "require conversation resolution before merging"). Pruefer's findings are submitted as non-blocking `event: "COMMENT"` reviews (ADR-1189 Decision 5's sibling constraint — confirmed in `github/prs.go:583`), so GitHub's own merge gating does not stop native auto-merge on an unresolved thread today; this issue is about Fabrik enforcing it, not GitHub.
- Reworking how Pruefer posts findings, or the review-reinvoke mechanism itself (§2.9) — this issue only changes when/whether the merge-side action is allowed to proceed.
- The merge-train `Queued` blackout window — a separate race with a different remedy (eject-and-requeue), tracked in #1208 and chained behind this issue.

## Prior Art / Context

- `engine/stages.go:246` — `attemptMergeOnValidate`, the one-shot decision point at Validate completion.
- `engine/merge_gate.go:212` — `checkAutoMergeConvergence`, the function that polls repeatedly between auto-merge being enabled and GitHub completing the merge; the second guard point.
- `engine/stages.go:250-258` — `fabrik:auto-merge-enabled`'s dual role as idempotency guard and convergence budget-start anchor; relevant to the disable/re-enable interaction called out in Requirements.
- `engine/reviews.go:432` — `buildReviewThreadComments`, existing unresolved-thread detection, reused by §2.9's reinvoke loop.
- `docs/state-machine.md` §2.9 (Review Reinvoke), §2.16 (SHA-Invalidation Scan).
- The GraphQL `reviewThreads` fetch (`github/project.go:704-729`) carries thread-level `isResolved`/`path`/`line`/`diffSide`, but individual thread *comments* do not currently carry a per-comment original-commit SHA — only `diffHunk`, `path`, and `line`. Scoping "unresolved on current head" vs. "stale SHA" may need a small addition to the GraphQL query (e.g. `originalCommit { oid }`) unless an equivalent signal already exists elsewhere.
- #1183 / PR #1202 — the live loss that motivated this issue.
- #1189 (`adrs/1189-pruefer-inline-review-comments.md`) — inline comments via non-blocking `COMMENT`-event reviews, which is both why this race is possible (findings are now machine-consumable) and why GitHub's own branch protection doesn't already block it (the review event type isn't `CHANGES_REQUESTED`).
- #1208 — the merge-train `Queued` blackout window; a related but separate race, out of scope here, chained behind this issue.
- #1071 — carries a design direction expected to supersede this stopgap: replacing the undecidable "have objections appeared?" check (which needs a guard at every advancement point, as this issue does) with the decidable "do I hold a valid signature from every required signer for this SHA?", checked once. If that lands, this guard largely dissolves.

## Risks / Dependencies

- The disable-then-re-enable interaction with `fabrik:auto-merge-enabled`'s dual role (idempotency guard + convergence budget anchor) is the main implementation risk — it must be verified to round-trip cleanly (disable → resolve → re-enable → budget behaves correctly) before committing to the full mechanism. If it doesn't, fall back to the narrower guard-only-at-`attemptMergeOnValidate`-plus-loud-log version rather than forcing a risky mutation path.
- The `checkAutoMergeConvergence` guard must not fight the merge-queue (`PRMergeQueued`) or ejection-recovery branches already in that function.
- `MaxReviewCycles` must continue to bound and escalate the reinvoke loop regardless of which guard point triggered it.
- This is explicitly a stopgap, expected to be superseded by #1071's signature-based design. Build narrowly — do not generalize this mechanism beyond yolo/Validate, and do not invest in infrastructure meant to outlive it.

## Definition of Done

- A yolo item with an unresolved review thread on its current head does not advance out of Validate and does not have auto-merge enabled on its behalf (`attemptMergeOnValidate` guard).
- If auto-merge was already enabled and an unresolved thread appears on the current head during convergence, auto-merge is disabled before GitHub can merge underneath the reinvoke loop — or, if that proves unsafe per the sanctioned fallback, a loud log is emitted at that point instead. Whichever is implemented must be stated explicitly in the PR.
- Once the thread is resolved by the reinvoke loop, the item advances / auto-merge is re-enabled normally.
- A thread against a superseded SHA does not block indefinitely, at either guard point.
- The deferral (and, if implemented, the disable) is logged with SHA and thread count.
- `MaxReviewCycles` still bounds and escalates; no new never-advancing state at either guard point.
- Tests: unresolved-on-current-head blocks at `attemptMergeOnValidate`; resolved advances; stale-SHA thread does not block; cycle limit still escalates; the convergence-window behavior (disable-and-reenable, or the fallback log) is covered.
- `go test -race ./...` passes.

## Labels

fabrik:yolo

arbeithand:
Closes #1207

## Summary

A `fabrik:yolo` item must not merge while unresolved PR review threads exist on its current head commit. This requires guarding **two** points, not one: the initial decision to advance out of Validate / enable auto-merge (`attemptMergeOnValidate`), and the ongoing window between auto-merge being enabled and GitHub actually completing the merge (`checkAutoMergeConvergence`'s repeated polling). Guarding only the first point would not have caught the incident above — all three earlier findings were already resolved by the time Validate completed, and the fourth arrived after that one-shot decision had already run. If a late finding appears during the convergence window, Fabrik must disable auto-merge before GitHub can merge underneath the reinvoke loop, then let the existing review-reinvoke loop (§2.9) address it and re-enable once resolved.

This is an explicit stopgap, not a general redesign — see Risks / Dependencies.

## Problem

A `fabrik:yolo` item can merge with **unresolved PR review threads**, losing review findings that arrived shortly before Validate completed.

Observed on #1183 / PR #1202 on 2026-07-27:

```
21:29:03  claude[bot]              inline finding   → Fabrik reinvoked Review, resolved
21:37:13  handarbeit-pruefer[bot]  inline finding   → Fabrik reinvoked Review, resolved
21:42:10  handarbeit-pruefer[bot]  inline finding   → resolved
21:47:11  Fabrik: Validate stage completes
21:48:56  handarbeit-pruefer[bot]  inline finding   → NEVER SEEN
21:50:07  PR #1202 MERGED
```

The reinvoke loop worked for the first three — each finding triggered a Review reinvoke, was addressed, and its thread was resolved. The fourth arrived **71 seconds** before the merge and was lost. That finding (a dead `else if` branch when `TerminalReason == "blocking_limit"` is excluded by the turns/cost guard) is now on `main`, unaddressed.

### Why the existing mechanisms don't cover it

**§2.9 Review Reinvoke** fires on any `stage:<X>:complete` item with unresolved threads, and it is stage-agnostic — it is what handled the first three. But it only helps while the item is *parked* at a completed stage. With `yolo`, Validate completing does not park anything: `attemptMergeOnValidate` runs in the same flow and advances to Queued (or enables native auto-merge) immediately.

**§2.16 SHA-invalidation** compares `snap.ValidateCompletedSHA()` against `LinkedPR.HeadSHA` and un-completes Validate on mismatch. But **a review comment does not change the head SHA** — the finding was posted against the already-validated head. Nothing to detect.

So the back-edge exists for *"a commit arrived after Validate"* and not for *"a finding arrived after Validate."*

Note `fabrik:cruise` is largely unaffected: it deliberately stops at Validate, so a late finding is picked up by the normal reinvoke loop before a human merges. This is specifically a yolo/auto-advance race.

### Why it is getting worse, not better

#1189 made Pruefer post findings as inline comments, and Pruefer re-reviews on every new head SHA. Fabrik pushes a fix per reinvoke, each producing a new SHA, each triggering a fresh review. **The faster the fix loop runs, the more likely a review lands inside the merge window.** This race is a consequence of the loop working.

### Approach

Guard both points identified in Research, with a key correction to the issue's own Prior Art: the disable mutation at guard 2 cannot live inside `checkAutoMergeConvergence` as literally suggested — Research confirmed `handleReviewGate` (Phase 1 Handler 2) always claims the item and dispatches a reinvoke *before* `handleAutoMergeConvergence` (Handler 3) ever runs, whenever a fresh unresolved thread exists on an item that already carries `fabrik:auto-merge-enabled`. Placing the check only in `checkAutoMergeConvergence` would never fire in the exact scenario that produced the #1183/PR#1202 incident. So the disable mutation is added *inside* `handleReviewGate`, at the point it already decides to dispatch a review reinvoke — no reordering of `catchUpPhase1Handlers` (ADR-056 D3's structural ordering stays untouched).

Both guards reuse `buildReviewThreadComments` per the Requirements, filtered through a new `currentHeadReviewThreadComments` helper that excludes threads GitHub itself marks `isOutdated` (a thread-level GraphQL field not currently fetched). This is simpler than the alternative floated in the issue's Prior Art (adding `originalCommit { oid }` per comment and comparing SHAs manually) and needs no new comparison logic in the engine — GitHub already computes "has this thread's commented lines been superseded by a later push."

Guard 1 (`attemptMergeOnValidate`) gets the check added directly inside the function (single source of truth, matches the issue's literal code pointer), returning a new third value — `deferred bool` — alongside the existing `(enabled bool, error)`. This is a real signature change (not a boolean-overload of the existing "no action needed" `(false, nil)` case), because the caller in `handleStageComplete` needs to distinguish "blocked by a real review thread — must not advance" from the two pre-existing benign `(false, nil)` cases (cruise, no linked PR) where the existing code already lets `shouldAdvance` stay true. Overloading the existing boolean would force the call site to re-derive *why* it got `false` by independently re-checking cruise/idempotency conditions — a "boolean blindness" smell. The mechanical cost is updating ~15 existing two-value call sites in `engine/stages_test.go`; that's a one-line find/replace, not a design cost.

Guard 2's disable mutation is a *real* GitHub-side action (native `disablePullRequestAutoMerge` or `DequeuePullRequest` for queue-enabled repos, branching exactly like `reenableAutoMergeAfterRebase` already does), followed by removing `fabrik:auto-merge-enabled` — mutation first, then label removal. That ordering is deliberate: if we removed the label first and the GitHub mutation then failed, Fabrik would stop watching the PR (`handleAutoMergeConvergence` is gated on the label's presence) while GitHub might still merge it — recreating the exact race this issue closes. Doing the mutation first means a failure keeps the label in place, so the same disable is retried next poll; the correctness-relevant risk is deliberately traded so that a failure re-tries rather than exposes a silent merge window.

Removing the label lets the *existing* mechanisms do everything else without new state: `handleAutoMergeConvergence` (Handler 3) stops running entirely for this item (guarded on label presence), so it can't fight the ejection-recovery ladder or misread the dequeue as a queue ejection. The normal `dispatchWithCycleLimit`/`MaxReviewCycles` path in `handleReviewGate` still bounds the reinvoke loop unchanged. Once the thread resolves, the label's absence causes the very next poll's Phase 2 Validate branch (`poll.go:1066`, already a retry loop) to call `attemptMergeOnValidate` again, which re-enables auto-merge or re-enqueues with a fresh label — the same remove-then-reapply idiom `pauseForMergeGroupStall` already uses successfully today for the convergence-budget anchor. This means the "budget resets on every disable/re-enable cycle" behavior is accepted as-is (bounded overall by `MaxReviewCycles`, not by wall-clock) rather than building new cumulative-timestamp tracking — consistent with the issue's explicit instruction not to over-build a stopgap.

### New/Modified Files

| File | Change |
|------|--------|
| `github/project.go` | Add `isOutdated` to the `reviewThreads` GraphQL selection and its decode struct; propagate to `Comment.IsOutdated` in `applyLinkedPRs` |
| `github/types.go` | Add `IsOutdated bool` field to `Comment` |
| `github/prs.go` | Add `DisablePullRequestAutoMerge(owner, repo string, prNumber int) error` (mirrors `EnablePullRequestAutoMerge`'s two-step `prNodeID` → mutation pattern) |
| `github/prs_test.go` | Unit test for `DisablePullRequestAutoMerge` |
| `engine/interfaces.go` | Add `DisablePullRequestAutoMerge` to `GitHubClient` |
| `engine/mocks_test.go` | Add mock impl + call-recording slice for `DisablePullRequestAutoMerge` |
| `engine/reviews.go` | Add `currentHeadReviewThreadComments(item) []gh.Comment` — `buildReviewThreadComments` filtered to exclude `IsOutdated` |
| `engine/stages.go` | `attemptMergeOnValidate` signature → `(enabled, deferred bool, err error)`; add the current-head-thread guard inside it; update `handleStageComplete`'s call site and `shouldAdvance` logic |
| `engine/poll.go` | Update the `poll.go:1066` call site for the new 3-value return (discard `deferred` — Phase 1 already prevents reaching this line while blocked) |
| `engine/merge_gate.go` | Add `disableAutoMergeForReviewThreads(item, blockingCount)` — queue-aware disable + label removal + logging |
| `engine/catch_up_handlers.go` | Wire the disable call into `handleReviewGate`, before `dispatchWithCycleLimit` |
| `engine/stages_test.go` | Update ~15 existing call sites for the new signature; add guard-1 tests |
| `engine/catch_up_handlers_test.go` | Add guard-2 tests (disable + stale-SHA + cycle-limit + round-trip) |
| `docs/state-machine.md` | Update §2.9 (Review Reinvoke) to document both guards and the handler-ordering rationale |
| `docs/llms-full.txt` | Regenerate via `scripts/generate-llms-full.sh` |
| `adrs/1207-*.md` (new) | ADR for the handler-ordering decision and the stopgap's scope |

### Key Decisions

- **`isOutdated` (GitHub-computed) over manual SHA comparison**: resolves the issue's own open Technical Question. Avoids adding `originalCommit { oid }` plumbing and a new per-comment SHA-comparison point in the engine; GitHub already computes staleness from the current diff.
- **Guard 1 lives inside `attemptMergeOnValidate` itself, with a widened return signature** rather than a boolean-overloaded `(false, nil)`: the caller must distinguish "genuinely blocked, do not advance" from the two existing benign `(false, nil)` cases (cruise, no PR), and re-deriving that from outside the function would duplicate its internal branching logic.
- **Guard 2's disable mutation lives inside `handleReviewGate`, not `checkAutoMergeConvergence`**: this deviates from the issue's Prior Art pointer (`engine/merge_gate.go:212`) but is required by the Requirements' own goal ("guard both points... The second half is the important one") — Research proved the literal placement wouldn't fire in the incident scenario due to Phase 1 handler-claim ordering (ADR-056 D3). No handler reordering is used; `handleReviewGate` already runs first by design ("review threads addressed before any merge/CI gate evaluation"), so this is the natural site for the disable, not a workaround.
- **Mutation before label removal**: on partial failure (mutation succeeds, label removal fails), the worst outcome is `checkAutoMergeConvergence`'s existing "user disabled auto-merge" pause firing on the next poll — a safe, human-visible failure mode, not a silent merge. The reverse order risks the opposite: GitHub still auto-merging while Fabrik has stopped watching.
- **Convergence-budget reset on each disable/re-enable is accepted, not re-engineered**: matches `pauseForMergeGroupStall`'s existing remove-and-reapply idiom. Bounded by `MaxReviewCycles` (dispatch count), not wall-clock; building a cumulative first-enabled timestamp would be speculative infrastructure this explicit stopgap should not carry (see #1071).
- **No reuse of `pauseOpts.removeAutoMerge`**: that flag is Fabrik-label-only everywhere it's used today (never calls a real GitHub mutation). Conflating it with this issue's real disable mutation would silently change the semantics of three existing pause call sites. `disableAutoMergeForReviewThreads` is a new, narrowly-scoped function instead.

### Task Checklist

- [ ] Add `isOutdated` to the `reviewThreads` GraphQL query and decode struct in `github/project.go`; set `Comment.IsOutdated` in `applyLinkedPRs` alongside the existing `ReviewThreadID` assignment
- [ ] Add `IsOutdated bool` field to `github.Comment` (`github/types.go`)
- [ ] Add `DisablePullRequestAutoMerge(owner, repo string, prNumber int) error` to `github/prs.go` (GraphQL mutation `disablePullRequestAutoMerge(input: { pullRequestId: $prId })`, same two-step pattern as `EnablePullRequestAutoMerge`); add a unit test in `github/prs_test.go`
- [ ] Add `DisablePullRequestAutoMerge` to the `GitHubClient` interface (`engine/interfaces.go`) and to the mock in `engine/mocks_test.go` with a call-recording slice
- [ ] Add `currentHeadReviewThreadComments(item) []gh.Comment` to `engine/reviews.go` — reuses `buildReviewThreadComments`, filters out `IsOutdated` comments
- [ ] Change `attemptMergeOnValidate`'s signature to `(enabled, deferred bool, err error)`; add the current-head-thread guard (after the cruise/idempotency checks, before the merge-train/enqueue/enable logic), logging `"not advancing: N unresolved review thread(s) on <sha>"` and returning `(false, true, nil)`
- [ ] Update `handleStageComplete`'s call site (`engine/stages.go`) to thread `deferred` through and combine it with `autoMergeEnabled` in the `shouldAdvance = false` decision; update the `poll.go:1066` call site for the new 3-value return (discard `deferred`)
- [ ] Mechanically update the ~15 existing `attemptMergeOnValidate` call sites in `engine/stages_test.go` for the new signature
- [ ] Add guard-1 tests: unresolved current-head thread blocks (no label applied, no `EnablePullRequestAutoMerge`/`EnqueuePullRequest`/`MergePR` call, `deferred == true`); an all-outdated thread does not block; a resolved item proceeds normally; an end-to-end `handleStageComplete` test (not just the direct function call) proving the item does not advance past Validate when blocked
- [ ] Add `disableAutoMergeForReviewThreads(item gh.ProjectItem, blockingCount int)` to `engine/merge_gate.go`: branch on `item.LinkedPRIsMergeQueueEnabled` (`DequeuePullRequest` vs `DisablePullRequestAutoMerge`), log `"disabling auto-merge: N unresolved review thread(s) on <sha>"`, remove `fabrik:auto-merge-enabled` only on mutation success, log-and-keep-label on failure
- [ ] Wire the disable call into `handleReviewGate` (`engine/catch_up_handlers.go`): inside the existing `len(syntheticComments) > 0` branch, before `dispatchWithCycleLimit`, call `disableAutoMergeForReviewThreads` when `fabrik:auto-merge-enabled` is present and `currentHeadReviewThreadComments(item)` is non-empty
- [ ] Add guard-2 tests (`engine/catch_up_handlers_test.go`): disable fires for a native-auto-merge item with a non-outdated unresolved thread; `DequeuePullRequest` fires instead for a merge-queue-enabled item; an all-outdated thread does not trigger disable; the disable still fires on the same poll that hits `MaxReviewCycles` and pauses (not just the dispatch branch)
- [ ] Add a round-trip test: disable → thread resolved (ROCKET reaction / `CommentProcessed`) → next poll's Phase 2 Validate branch re-invokes `attemptMergeOnValidate` → auto-merge (or enqueue) re-enabled with a fresh `fabrik:auto-merge-enabled` label, no stuck state
- [ ] Update `docs/state-machine.md` §2.9 (Review Reinvoke) to document both guards, including why the guard-2 mutation lives inside Handler 2 rather than `checkAutoMergeConvergence`; regenerate `docs/llms-full.txt` via `scripts/generate-llms-full.sh`
- [ ] Create ADR (check `adrs/` for the current highest number at implement time — do not hardcode): document the handler-ordering deviation from the issue's Prior Art, the `isOutdated`-based stale-SHA scoping choice, the mutation-before-label-removal ordering, and the accepted convergence-budget-reset behavior; note this is a stopgap superseded by #1071
- [ ] Run `go build`, `go vet ./...`, `go test -race ./...`

### Risks

- **Test-signature churn**: the `attemptMergeOnValidate` 3-value return touches ~15 existing test call sites. Purely mechanical, but a missed site fails the build immediately (not a silent risk).
- **Partial-failure window in `disableAutoMergeForReviewThreads`**: if the GitHub mutation succeeds but the label-removal API call fails, the issue may get paused by `checkAutoMergeConvergence`'s existing "user disabled auto-merge" path on the next poll. This is accepted as a safe (human-visible, not silent) degraded outcome, consistent with existing patterns elsewhere in this codebase that don't guarantee atomic mutation+label round-trips.
- **Residual race is not eliminated, only bounded**: GitHub can still complete a merge between poll cycles; this reduces exposure to `PollSeconds`, not zero, as the issue itself acknowledges is out of scope to fully close (would require branch-protection changes).
- **Stopgap scope discipline**: per the issue's explicit instruction, this mechanism must not be generalized beyond yolo/Validate — Implement and Review should watch for any temptation to extract a reusable "block on unresolved threads" abstraction beyond the two call sites here.

---
Used 27/50 turns, 11k input / 48k output tokens.

## Verification

Verified all prior Implement work for #1207 is complete: both yolo-merge guards (`attemptMergeOnValidate` and the `handleReviewGate`-embedded disable, correctly deviating from the issue's own Prior Art pointer at `checkAutoMergeConvergence` per Research's handler-ordering finding) are implemented, tested (guard-1, guard-2, stale-SHA, cycle-limit, and full disable/re-enable round-trip), documented in `docs/state-machine.md` §2.9 with `llms-full.txt` regenerated cleanly, and captured in ADR-1207. Confirmed `go build`, `go vet`, and `go test -race ./...` pass (one `cmd` test fails identically on unmodified `origin/main` too — a pre-existing sandbox network/SIGINT-timing flake, unrelated to this change). Checked off all 16 remaining Plan checklist items on the Plan stage comment; branch is already fully pushed and in sync with remote.

---

Closes #1207


arbeithand:
*branch: fabrik/issue-1207 | commit: c344dd62 | main: e5251cf51947bc220504f8901fcbd7269d06a88f | 2026-07-28 00:37 UTC*

Branch is already fully in sync with remote, working tree clean, nothing to commit or push.

Verified all prior Implement work for #1207 is complete: both yolo-merge guards (`attemptMergeOnValidate` and the `handleReviewGate`-embedded disable, correctly deviating from the issue's own Prior Art pointer at `checkAutoMergeConvergence` per Research's handler-ordering finding) are implemented, tested (guard-1, guard-2, stale-SHA, cycle-limit, and full disable/re-enable round-trip), documented in `docs/state-machine.md` §2.9 with `llms-full.txt` regenerated cleanly, and captured in ADR-1207. Confirmed `go build`, `go vet`, and `go test -race ./...` pass (one `cmd` test fails identically on unmodified `origin/main` too — a pre-existing sandbox network/SIGINT-timing flake, unrelated to this change). Checked off all 16 remaining Plan checklist items on the Plan stage comment; branch is already fully pushed and in sync with remote.

---
Used 43/50 turns, 15k input / 10k output tokens.

arbeithand:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-1207 | commit: c344dd62 | main: e5251cf51947bc220504f8901fcbd7269d06a88f | 2026-07-28 00:44 UTC*

Working tree is clean and everything is already pushed (confirmed at session start: `origin/fabrik/issue-1207` matches HEAD). Review complete.

## Review Findings

### Verified
- Rebased onto latest `main` (already up to date)
- `go test -race -timeout 5m ./...` — 3041 passed, 1 failed (pre-existing `cmd` flake, confirmed to fail identically on unmodified `main` — unrelated to this change), 1 skipped
- No external bot/human review feedback pending on PR #1211
- `docs/llms-full.txt` regeneration produces no diff — already in sync with `docs/state-machine.md`
- Traced both guard points end-to-end against the actual control flow:
  - **Guard 1** (`attemptMergeOnValidate`, `engine/stages.go`): correctly placed after the cruise/idempotency checks and before the merge-train/enqueue/enable branches, using the new `currentHeadReviewThreadComments` helper (excludes GitHub's `isOutdated` threads). Widened `(enabled, deferred, err)` signature correctly threaded through `handleStageComplete` and all ~19 test call sites.
  - **Guard 2** (`handleReviewGate`, `engine/catch_up_handlers.go`): correctly placed *inside* Handler 2, not `checkAutoMergeConvergence` — verified this is necessary, since Handler 2 always claims and stops the Phase 1 chain before Handler 3 ever runs when a fresh unresolved thread exists on an auto-merge-enabled item. Placing it in `checkAutoMergeConvergence` as the issue's own Prior Art literally suggested would not have closed the actual #1183/PR#1202 race — Research/Plan correctly caught and documented this deviation.
  - `disableAutoMergeForReviewThreads` (`engine/merge_gate.go`) correctly branches on merge-queue vs. native auto-merge, mutates GitHub first and only removes the label on success (safe failure mode), and the round-trip test confirms disable → resolve → re-enable works with no stuck state.
- `MaxReviewCycles` still bounds and escalates at guard 2 — explicitly tested (`TestHandleReviewGate_DisablesAutoMerge_EvenWhenCycleLimitReached`) that the disable fires even on the cycle-limit-hit poll.
- Stale-SHA scoping (`isOutdated`) tested at both guard points, plus a full GraphQL-decode integration test (`TestFetchItemDetails_ReviewThreadIsOutdated`).
- New `DisablePullRequestAutoMerge` GraphQL mutation matches GitHub's real schema and mirrors the existing `EnablePullRequestAutoMerge` two-step pattern; unit-tested.
- ADR-1207 and `docs/state-machine.md` §2.9 accurately document the implementation, including the deliberate deviation from the issue's Prior Art and the rationale for each key decision.
- No scope creep: no reusable "block on unresolved threads" abstraction was extracted beyond the two guard call sites, per the issue's explicit stopgap instruction.

### Fixed
None — no issues found requiring a fix.

### Blocking
None.

PR #1211 is ready for human review.

Reviewed the yolo-merge review-thread guards implementation (#1207). Verified both guard points work correctly, including the load-bearing deviation from the issue's own Prior Art (guard 2 lives in `handleReviewGate`, not `checkAutoMergeConvergence`, due to Phase 1 handler-claim ordering). Build, vet, and full test suite pass (one pre-existing unrelated `cmd` flake confirmed present on `main` too). Docs, ADR, and `llms-full.txt` are in sync. No issues found — PR #1211 is ready for human review.

---
Used 50/50 turns, 16k input / 11k output tokens.

arbeithand:
*branch: fabrik/issue-1207 | commit: fceb9eb2 | main: e5251cf51947bc220504f8901fcbd7269d06a88f | 2026-07-28 00:49 UTC*

Thread resolved. The issue body doesn't require any updates — this was a review thread comment addressed with a documentation clarification, not a spec change.

## Summary

Pruefer's finding was valid: `attemptMergeOnValidate`'s guard 1, at its primary call site (`handleStageComplete`, run synchronously right after a Claude Validate invocation completes), uses the pre-invocation item snapshot for its review-thread check — which can be far staler than "a few seconds" on a long-running Validate stage, mirroring the exact staleness pattern the codebase already documents for `wait_for_reviews` a few lines below in the same function.

I traced the actual mitigation rather than assuming one: `poll.go`'s Phase 2 Validate retry (the other caller of `attemptMergeOnValidate`) always operates on deep-fetched, fresh data via `selectDeepFetchCandidates`, and guard 2 (`handleReviewGate`) runs on every subsequent poll with fresh review-thread data too — so a miss at the primary call site is closed within one poll cycle, the same `PollSeconds`-bounded residual race already accepted and documented for guard 2's convergence-window disable. Re-fetching synchronously inside `attemptMergeOnValidate` to close this fully would reintroduce the per-stage-completion GraphQL round-trip the existing Path 1/Path 2 split deliberately avoids, and isn't warranted given the backstop already exists.

I documented this explicitly (not a behavior change) in three places, committed and pushed:
- `engine/stages.go` — expanded the guard-1 comment
- `docs/state-machine.md` §2.9 — added a "Freshness caveat" paragraph
- `adrs/1207-yolo-merge-review-thread-guards.md` — added a trade-off bullet crediting Pruefer's review

Build, vet, and the full `engine`/`github` test suite pass. Resolved the review thread.

---
**Threads addressed:**
- `engine/stages.go:287` — resolved

arbeithand:
*branch: fabrik/issue-1207 | commit: fceb9eb2 | main: e5251cf51947bc220504f8901fcbd7269d06a88f | 2026-07-28 00:54 UTC*

All packages except the pre-existing flaky `cmd` package pass cleanly with the race detector. The validation is complete — everything checks out against the Definition of Done.

Validated #1207: both yolo-merge guards (attemptMergeOnValidate and the handleReviewGate-embedded disable) are correctly implemented per spec, including the load-bearing handler-ordering correction to the issue's own Prior Art. Build/vet/full test suite pass (one pre-existing `cmd` SIGINT/network flake confirmed present on unmodified origin/main too, plus a pre-existing gofmt misalignment in an unrelated struct — neither is a regression from this branch). Docs, ADR, and llms-full.txt are in sync. PR #1211 is ready to merge.

## Validation Report

### Requirements: 8/8 verified
- [x] Guard 1 (`attemptMergeOnValidate`): blocks advance/auto-merge-enable when unresolved current-head thread exists, widened `(enabled, deferred, err)` signature, logs `"not advancing: N unresolved review thread(s) on <sha>"`
- [x] Guard 2: correctly placed inside `handleReviewGate` (not `checkAutoMergeConvergence` as the issue's own Prior Art literally suggested) — verified this deviation is necessary via the Phase 1 handler-claim ordering (`catchUpPhase1Handlers`: Handler 2 always claims before Handler 3 runs)
- [x] Full disable-and-re-enable mechanism built (not the fallback), using `DisablePullRequestAutoMerge` (new mutation) / `DequeuePullRequest` (existing), mutation-before-label-removal ordering
- [x] Current-head scoping via GitHub's `isOutdated` field — stale-SHA threads never block indefinitely, at either guard point
- [x] Resolved threads advance automatically via existing retry loops (poll.go Phase 2 for guard 1, label-removal feeding the same path for guard 2) — no new plumbing
- [x] `MaxReviewCycles` still bounds/escalates — disable is additive to the existing dispatch, tested even on the cycle-limit-hit poll
- [x] Every deferral/disable logged with SHA and thread count
- [x] Reuses `buildReviewThreadComments` exclusively via `currentHeadReviewThreadComments` — no parallel detection path

### Test Suite: PASSED
- All packages pass with `-race`, except pre-existing `cmd` package flakes (SIGINT timing / sandbox network dependency) confirmed present identically on unmodified `origin/main` via a side-by-side temp worktree
- 45 targeted tests for this feature (guard 1, guard 2, stale-SHA, cycle-limit, round-trip) pass
- Build and vet clean

- `docs/llms-full.txt` regenerates with zero diff (in sync)
- gofmt flags one pre-existing misalignment in `github/types.go` (`MergeQueueEntry`, unrelated struct) — confirmed present on unmodified `origin/main`, not introduced here
- Branch fully rebased and pushed; no uncommitted changes

### Issues Found
None blocking.

### Verdict: READY TO MERGE

---
Used 34/50 turns, 18k input / 7k output tokens.

arbeithand:
*branch: fabrik/issue-1207 | commit: eab12b35 | main: a323ec771a94817b92aadf0320190bf6178d3945 | 2026-07-28 21:52 UTC*

No TODOs/FIXMEs left. Everything checks out — the implementation is complete, correct, well-tested, and well-documented. Ready to signal stage completion.

Validated #1207: rebased cleanly onto the latest main (which now includes ADR-1216's independent review-gate work), resolving one semantic conflict in `attemptMergeOnValidate`'s signature. Confirmed both yolo-merge guards work correctly — including the load-bearing deviation from the issue's own Prior Art (guard 2 lives in `handleReviewGate`, not `checkAutoMergeConvergence`, due to Phase 1 handler-claim ordering). Build/vet/full race-detector test suite pass (one `cmd` SIGINT/network test fails identically on unmodified `origin/main`, confirmed via a side-by-side worktree — pre-existing, unrelated). All 8 Definition-of-Done requirements verified against code, docs (`state-machine.md` §2.9), and the ADR; `llms-full.txt` regenerates with zero diff. Rebased branch force-pushed; PR #1211 has no unresolved review threads and is ready to merge pending CI.

---
Used 38/100 turns, 20k input / 11k output tokens.

arbeithand:
*branch: fabrik/issue-1207 | commit: daa5c3e1 | main: a323ec771a94817b92aadf0320190bf6178d3945 | 2026-07-28 22:00 UTC*

Running the thread-resolution in the background since it's rate-limited for a couple more minutes; I'll wrap up the summary now and it'll complete shortly.

Fixed Pruefer's finding: the direct-merge fallback in `attemptMergeOnValidate` (taken when the PR is already CLEAN/UNSTABLE) called `MergePR` synchronously with no convergence window, meaning guard 2's "catch it on the next poll" safety net never applied there. Added a live `FetchItemDetails` re-read immediately before that `MergePR` call — mirroring the existing re-read-before-irreversible-action pattern used elsewhere (`checkDependencies`, `recoverMissingPlanComment`, `verifyAndHealLinkage`) — so a fresh unresolved current-head thread now defers the merge instead of proceeding on stale data. Added 3 tests (defer-on-blocking, stale-SHA-doesn't-block, fetch-failure-surfaces-as-error), updated `docs/state-machine.md` §2.9 and ADR-1207 with the new decision, regenerated `llms-full.txt`, and committed/pushed (`daa5c3e1`). Full `engine`/`github` test suite passes. Thread resolution is running in the background pending a GitHub API rate-limit reset (~2 min).

---
**Threads addressed:**
- `engine/stages.go:372` — resolved

arbeithand:
*branch: fabrik/issue-1207 | commit: daa5c3e1 | main: a323ec771a94817b92aadf0320190bf6178d3945 | 2026-07-28 22:06 UTC*

Both of Pruefer's new review comments were confirmatory, not action items:

1. **Live-read failure behavior** — on a transient error from the new `FetchItemDetails` re-check, `attemptMergeOnValidate` returns an error, and `handleStageComplete` returns before applying `stage:Validate:complete` (line 130). I traced this against the existing code comment at `engine/stages.go:121-123`, which states this is *deliberate* pre-existing behavior for any error in this block ("runs BEFORE adding the completion label so that on failure the engine retries Validate"). My addition just reaches an already-intended error path through one more branch — not a new cost.
2. **Scoping check** — verified `enqueueForQueue` (merge-queue path) only calls `EnqueuePullRequest`, an async enqueue GitHub processes later under guard 2's convergence window, and `advanceToQueued` (merge-train path) only moves the board column — neither is a synchronous, no-window merge like the `MergePR` direct-fallback. Pruefer's self-assessment ("correctly scoped to the one branch that needed it") checks out.

No code changes were required. Resolved all three review threads (the original finding plus these two follow-ups) via GraphQL. Branch is clean and already in sync with `origin/fabrik/issue-1207` (commit `daa5c3e1`); no issue body update needed.

---
**Threads addressed:**
- `engine/stages.go:384` — resolved
- `engine/stages.go:384` — resolved

Issue #1217, opened 2026-07-28
e2e: exercise both merge_train modes; default (off) path is uncovered


arbeithand:
## Problem

The e2e suite only ever exercises whichever merge-train mode the test bed happens
to be configured with. It is currently `FABRIK_MERGE_TRAIN=on`, so the
`merge_train: off` landing path — **the path almost every user is actually on** —
is not covered end-to-end at all.

This is backwards from the risk profile. The merge train is opt-in and used by a
small minority; the default path carries nearly all real usage and gets the least
integration coverage.

It also produces false failures that cost real triage time. In the 2026-07-28 run
two scenarios failed purely because they encode `merge_train: off` assumptions
while running against a train-on bed:

- `TestConvergenceRace` asserts `fabrik:auto-merge-enabled` was applied. On a
  non-queue repo the internal train never applies it — by design
  (`engine/merge_gate.go:230`). Both issues completed the full pipeline and closed
  cleanly; only the label assertion failed.
- `TestYoloAutoMergeLabel` asserts the member PR ends `CLOSED`. Under the train it
  ends `merged`, because the integration merge makes the member branch's commits
  reachable from `main` and GitHub auto-marks the PR merged (verified: member PR
  #3739's `merge_commit_sha` is the train's own trial SHA, one second after the
  integration PR merged).

Neither was an engine defect, but both read as one.

Only 5 of 16 e2e files are train-aware today (`auto_merge_test.go` and the four
`mergetrain_*_test.go`). The remaining 11 silently assume whatever the bed is set
to.

## Summary

Make the e2e suite runnable in both `merge_train` modes via an explicit,
restart-based mode fixture (two sequential invocations, `E2E_TRAIN_MODE=on|off`),
generalize the existing single-scenario `readEnvFileMergeTrainMode` pattern so any
scenario can branch on mode, audit all 16 scenario files and classify each as
mode-invariant or mode-dependent, and fix the two known mode-divergent assertions
(`TestConvergenceRace`, `TestYoloAutoMergeLabel`) so both pass in both modes.
Document the two-mode run in `scripts/e2e/run.sh` and `tests/e2e/README.md` as the
pre-release gate.

## Requirements

- **FR-1** The suite MUST be runnable in both modes, and a full validation run MUST
  cover both.
- **FR-2** Mode MUST be explicit and self-describing at the suite level — not an
  ambient property of the bed's `.env` that a reader has to go look up. A scenario's
[...]

Issue #1216, opened 2026-07-28
Validate's wait_for_reviews gate never arms when wait_for_ci defers completion


arbeithand:
## Problem

`wait_for_reviews` is silently not enforced whenever `wait_for_ci` defers the same
stage's completion. A PR can merge with an outstanding requested reviewer and no
approving review, on a stage configured with **both** gates enabled.

Found by `TestConjunctiveCIReviewGate` in the 2026-07-28 e2e run against
`dev(53f9de9)`. Observed on `handarbeit/fabrik-test-alpha`:

```
PR #3747   merged_at 2026-07-28T04:34:16Z
           requested_reviewers: ["verveguy"]      <- still outstanding at merge
           reviews: 1 x github-actions[bot] COMMENTED   (no approval)
```

Engine trace for issue #3744 (Validate has `wait_for_ci: true` + `wait_for_reviews: true`):

```
04:25:00  [done]         stage "Validate" complete
04:25:01  [awaiting-ci]  deferring stage:Validate:complete until CI gate clears
04:25:0x                 (reviewer "verveguy" requested on PR #3747)
04:31:41  [ci-gate]      CI gate clears
04:31:43  [merge-train]  advancing to Queued
04:34:16                 PR merged
```

`fabrik:awaiting-review` was never applied at Validate.

### Mechanism (verified against `dev(53f9de9)`)

Two code paths each assume the other owns the gate.

**Path 1** — `engine/stages.go:150` (`handleStageComplete`) deliberately does not
seed `fabrik:awaiting-review` when `wait_for_ci: true`, and delegates:

> `fabrik:awaiting-review` is NOT seeded here when `wait_for_ci: true`. Path 2
> (`checkReviewGate` in the catch-up loop) handles the review gate after the CI
> gate clears and `stage:X:complete` is added (#617).

**Path 2** — `engine/catch_up_handlers.go:64` (`handleReviewGate`) guards on
`hasComplete`:

```go
func (e *Engine) handleReviewGate(pctx *phase1Ctx) bool {
	if !pctx.hasComplete {
		return false          // skipped during CI-await, by design (#617)
	}
```

`hasComplete` only becomes true when the CI gate clears. But the catch-up chain
order (`engine/catch_up_handlers.go:47`, `catchUpPhase1Handlers`) is:

```
dependencies -> reviewGate -> autoMergeConvergence -> mergeAndCIGates
```

so within the single poll pass that clears CI, `reviewGate` has *already* run
[...]

PR #1219, opened 2026-07-28
test(e2e): make the suite affordable — board reads via GraphQL, tunable poll base

arbeithand:
Makes the e2e suite affordable against the bed token's GraphQL budget. Test-only
— no engine code is touched.

## Why

A full `E2E_PARALLEL=4` run exhausted the bed token's entire 5,000 points/hour
GraphQL budget in ~14 minutes. Every scenario after that failed on `gh` exit 1,
which reads as a wall of engine regressions but is not — the 2026-07-27 run
looked like six failures and was entirely rate-limit noise.

The cause is that `gh project item-list` and `gh project field-list` resolve
field values with a follow-up query per item, so cost scales with the requested
limit rather than with the data needed. Measured against the live test board:

| Call | Cost |
|---|---|
| `gh project item-list --limit 200` | ~101 pts |
| `gh project field-list` | ~106 pts |
| equivalent hand-written query | ~1 pt |

At ~101 points a call on a 10-15s poll interval, a **single** scenario polling
continuously cost ~24,000 points/hour — 4.8x the whole budget on its own. Note
this could not have been fixed by lengthening the poll intervals: at that
per-call cost even a fully serial suite does not fit.

## What

**1. Board reads via direct GraphQL** (`fetchBoardItems`, `fetchStatusField`).
Replaces the two `item-list` poll loops, and collapses the `project view` +
`field-list` pair in `resolveStatusOption` into one call. Both select through
`repositoryOwner` with inline fragments on `Organization` and `User`, so they
work for either project owner type. `item-add` / `item-edit` are one-shot
mutations and stay on `gh project`.

**2. A tunable poll base.** With board reads cheap, the residual cost became the
issue/PR wait-helpers (`gh issue view --json`, `gh pr list --json` — also
GraphQL). The 13 hardcoded 15s intervals move to `pollBase()`, default 30s,
overridable via `E2E_POLL_INTERVAL`. `WaitForProjectStatus` stays at 10s: it is
now ~1 pt per call and is the helper most often on the critical path between an
engine action and the next assertion.

## Measured result

Verified through the Go path, not just the CLI: **10 board polls = 8 points**
(0.8 pts/poll), against ~101 before.

| | before | after |
|---|---|---|
| Suite, `E2E_PARALLEL=4` | budget exhausted in ~14 min | completed, 105 min |
| Bed engine (`poll: 30`) | ~480 pts/hr | ~480 pts/hr |
| Observed total | — | ~4,860 pts/hr, then ~2,400 with the poll base |

The first full green-path run of the suite came from this change: **12 pass,
4 fail, 1 skip**. Of the four failures, one is a genuine engine defect
(now #1216), two are test-expectation defects from the suite never having been
adapted to `merge_train: on` (now #1217), and one is environmental.

## Notes

`tests/e2e/README.md` gains a **GraphQL budget** section recording the measured
costs, so the next person does not have to rediscover them — including an
explicit warning against both regressions: reverting to `gh project`, and
"fixing" a future budget squeeze by lengthening intervals instead.

Unblocks #1217, which needs budget headroom before it can run the suite in both
merge-train modes.

### DIFF
--- a/tests/e2e/README.md
+++ b/tests/e2e/README.md
@@ -56,6 +56,41 @@ These tests assume:
 See `~/fabrik-oss-launch-notes.md` (under "Files and where they live") for
 the canonical setup.
 
+## GraphQL budget
+
+The bed token has 5,000 GraphQL points/hour, shared by the suite and the
+test-bed engine. Measured 2026-07-28:
+
+| Operation | Cost |
+|---|---|
+| `gh project item-list --limit 200` | ~101 pts |
+| `gh project field-list` | ~106 pts |
+| `fetchBoardItems` / `fetchStatusField` (harness.go) | ~2 pts |
+| Engine board poll, warm cache | ~4 pts |
+| Engine board poll, cold bootstrap | ~349 pts |
+
+**The harness must not read the board through `gh project` subcommands.** They
+resolve field values with a follow-up query per item, so cost scales with the
+requested limit rather than with the data needed. When the wait-helpers used
+them, a single scenario polling on a 10-15s interval cost ~24,000 pts/hour —
+nearly 5x the whole budget — and a full `E2E_PARALLEL=4` run exhausted the
+token in ~14 minutes. Everything after that fails with `gh` exit 1, which
+looks like a pile of engine regressions but is not.
+
+Use `fetchBoardItems` / `fetchStatusField` instead; they return the same data
+for ~2 points by resolving Status inline via `fieldValueByName`. At that cost a
+parallel-4 run is ~1,900 pts/hour, plus ~480 for the engine — roughly 2x
+headroom. `gh project item-add` / `item-edit` are one-shot mutations and are
+fine as-is.
+
+If you hit the limit anyway, check the reset time and wait it out rather than
+lengthening the poll intervals — at the old cost even a fully serial suite did
+not fit, so long intervals treat the symptom:
+
+```bash
+gh api rate_limit --jq '.resources.graphql'
+```
+
 ### Additional prerequisites for `TestCIFixReinvoke` and `TestCIFixReinvokeCycleLimit`
 
 5. **`ci-fix-sentinel` enrolled as a required status check** on
--- a/tests/e2e/README.md
+++ b/tests/e2e/README.md
@@ -83,9 +83,19 @@ parallel-4 run is ~1,900 pts/hour, plus ~480 for the engine — roughly 2x
 headroom. `gh project item-add` / `item-edit` are one-shot mutations and are
 fine as-is.
 
-If you hit the limit anyway, check the reset time and wait it out rather than
-lengthening the poll intervals — at the old cost even a fully serial suite did
-not fit, so long intervals treat the symptom:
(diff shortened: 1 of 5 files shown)


arbeithand:
Both review points addressed in f8a6eae9 — thanks, the first one is a real hole.

**Missing existence check on `fetchBoardItems`** — fixed. Both queries now select `id` on `projectV2` and error when it is empty.

One correction to the stated mechanism, because it changes what the check is actually guarding. I tested both misconfigurations against the live API:

```
bad owner login    -> {"data":{"repositoryOwner":null}}, no errors[], gh exit 0
wrong project num  -> NOT_FOUND in errors[],             gh exit 1
```

A wrong or stale project number is *not* silent — GitHub returns a top-level `NOT_FOUND` and `gh` exits non-zero, so `ghOutput` already surfaced it as an error before this PR. The reachable silent failure is a mistyped **owner** login: `repositoryOwner` resolves to null with no `errors` array at all and `gh` exits 0. The conclusion stands and the fix covers both (a null owner leaves `ProjectV2.ID` empty), but the guard is against `FABRIK_TEST_PROJECT_OWNER` drift rather than project-number drift.

Worth flagging a trap in the obvious version of this fix: the check has to be on the **project node**, not the item count. An empty board is a routine, legitimate state — `scripts/e2e/reset.sh` drains every item as the first step of a clean run, and the bed sat in exactly that state during last night's run. Erroring on "no items" would fail spuriously against a freshly reset bed.

**`fields(first:50)` unpaginated** — addressed, but deliberately not by paginating. The query now selects `pageInfo`, and if Status is not found while `hasNextPage` is true the error says so explicitly instead of claiming the field does not exist. A project with 50+ custom fields would be pathological, and a page-through loop would cost a round trip in every scenario's setup for a case that does not occur. If you think the loop is worth it anyway, say so and I will add it.

arbeithand:
Third finding addressed in 1e0e4d42 — a genuine miss on my part, and correctly diagnosed.

I migrated the 13 `pollSleep(15 * time.Second)` sites in `harness.go` and never checked whether the scenario files had their own polling loops. They did: seven `time.Sleep(15 * time.Second)` calls around `tryIssueLabels`/`tryIssueState`, both GraphQL via `gh issue view --json`, at exactly the lines named. All seven now go through `pollSleep(pollBase())`.

Two consequences beyond the ones you listed, both of which make the finding sharper:

- These sites called `time.Sleep` **directly**, so they also bypassed `pollSleep`'s ±20% jitter. That jitter exists specifically to desynchronise concurrent scenarios against the shared rate-limit budget (see the comment at `pollSleep`), so the loops skipping it were the ones where lockstep polling cost the most.
- `conjunctive_ci_review_gate_test.go` is the scenario that runs longest against the review gate, so its four sites were a disproportionate share of the residual burn the README claims to have halved.

The earlier two threads are now resolved: the `projectV2.id` existence check and the `fields(first:50)` `pageInfo` guard both landed in f8a6eae9. Noting again for the record that the first one's stated mechanism was inverted — a wrong project number already exits non-zero; the silent case is a mistyped **owner** — but the conclusion held and the fix covers both.

Issue #617, opened 2026-05-07
fabrik: spurious fabrik:awaiting-review re-applied at Validate-complete; survives into Done


verveguy:
## Problem

Issue #608 ended in Done state still carrying `fabrik:awaiting-review`, even though Copilot's review had been submitted and processed correctly during the Review stage. Investigation reveals the label was **spuriously re-applied** at the Validate→Done transition, then never cleared before issue close.

## Reproduction (#608 label timeline)

```
00:21:56  labeled fabrik:awaiting-review     ← set after Review complete
00:21:58  unlabeled fabrik:awaiting-review   ← cleared 2s later (Copilot reviewed) ✅
...
00:23:43  stage:Validate:in_progress
...
00:29:10  labeled fabrik:awaiting-ci         ← Validate completes, CI gate engages
00:29:11  labeled fabrik:awaiting-review     ← ⚠ SPURIOUSLY re-applied at Validate-complete
00:29:12  labeled stage:Validate:complete
00:29:12  unlabeled fabrik:awaiting-ci       ← cleared as expected
00:29:16  closed                             ← awaiting-review never cleared before close
```

Copilot did NOT fail — they reviewed within seconds and the label cleared correctly during Review. The bug is that the review-gate evaluation runs again at Validate-complete and re-applies the label even though we're already past Review.

## Summary

There are two bugs. Bug A is the root cause: `checkReviewGate` in `engine/reviews.go` is invoked by the catch-up loop in `engine/poll.go` for items that have `fabrik:awaiting-ci` set (Validate-complete / CI gate engaged), and at that point a stale `LinkedPRReviewRequests` snapshot causes the gate to re-block and re-apply `fabrik:awaiting-review`. Bug B is a defensive enhancement: transient lifecycle labels should be swept when an issue is closed so terminal-state issues don't carry leftover labels regardless of how Bug A resolves.

## Codebase Context

- **`engine/poll.go:976-984`** — The catch-up loop admits items with `fabrik:awaiting-ci` on a `wait_for_ci` stage (not just `stage:X:complete`), then unconditionally calls `checkReviewGate`. This is the call site for Bug A.
- **`.fabrik/stages/validate.yaml`** — Confirms `wait_for_reviews: true` and `wait_for_ci: true` are both enabled on the Validate stage (Review stage also has `wait_for_reviews: true`).
[...]

PR #77, opened 2026-04-02
Add max retries and escalation for failed stages

verveguy:
Closes #71

PR #1232, opened 2026-07-28
fix(engine): enforce wait_for_reviews at the Validate landing decision

### LINKED ISSUE #1216: Validate's wait_for_reviews gate never arms when wait_for_ci defers completion

arbeithand:
Closes #1216

`wait_for_reviews` was silently unenforced whenever `wait_for_ci` deferred the same stage's completion — i.e. on essentially every real PR, since CI is never already green when `FABRIK_STAGE_COMPLETE` fires. Reproduced against `dev(53f9de9)`: `fabrik-test-alpha` PR #3747 merged with `requested_reviewers: ["verveguy"]` outstanding and no approval.

## Mechanism

Two paths each assumed the other owned the gate, and they don't compose because there is no poll boundary between them:

- `pctx.hasComplete` is computed **once**, before the Phase 1 handler chain runs.
- `catchUpPhase1Handlers` orders `reviewGate` **ahead of** `mergeAndCIGates`.
- On the pass CI turns green: `reviewGate` skips itself (snapshot predates the completion label) → `mergeAndCIGates` adds `stage:X:complete` and returns unclaimed → **Phase 2 calls `attemptMergeOnValidate` in that same loop iteration**.

A second, `wait_for_ci`-independent instance: `handleStageComplete` calls `attemptMergeOnValidate` when `yoloActive && !waitForCI`, *before* its own `fabrik:awaiting-review` seeding block runs.

Under `merge_train: off` this is masked by GitHub auto-merge honouring branch protection. Under `merge_train: on` Fabrik is the merge authority and the reviewer request is simply dropped — so a repo becomes unsafe purely by turning the train on.

## Changes

- **`engine/reviews.go`** — new `reviewGateBlocksLanding`: opt-in check (zero cost when `wait_for_reviews` is unset), PR number from `item.LinkedPRNumber` with a `FetchLinkedPR` fallback for the `base:<branch>` case, **live** `FetchPRReviews`/`FetchPRReviewRequests`, same `reviewGateOutstanding` clearing condition as `checkReviewGate`. Blocks conservatively on a fetch error (both slices discarded); does not block when no PR resolves. Applies `fabrik:awaiting-review` idempotently and returns — no escalation, no timeout.
- **`engine/stages.go`** — `attemptMergeOnValidate` now uses its `stage` param and calls the gate ahead of the `merge_train` fork.
- **Docs** — `docs/state-machine.md` §6.6.6 rewritten (it described a two-poll handoff that doesn't exist for a `yoloActive` Validate item, citing a test that never reaches the failing path); Path 3 added to the review-gate path list and the label-semantics entry; `CLAUDE.md` bullet; ADR-1216; `docs/llms-full.txt` regenerated.

`catchUpPhase1Handlers` ordering is deliberately untouched — a reorder cannot fix the `handleStageComplete` instance, and once the gate lives at the choke point it is unnecessary. FR-4 holds for free: `attemptMergeOnValidate` is unreachable while CI genuinely blocks, because `mergeAndCIGates` claims the item then.

## How to test

```bash
go test -race ./engine/
go test ./engine/ -run 'TestValidateLanding|TestAttemptMergeOnValidate_ReviewGate|TestHandleStageComplete_ReviewGate'
```

The new tests were verified to genuinely catch the defect: reverting the one-line gate call in `attemptMergeOnValidate` fails 4 of the `TestAttemptMergeOnValidate_ReviewGate_*` cases and both single-`poll()` reproductions. `TestCIAndReviewGate_JointClearingHandoff`, `TestCatchUpPhase1HandlersOrder`, and both `TestCatchupLoop_*ReviewGate*` pass **unmodified** — no assertions weakened.

Key coverage: single-`poll()` reproductions for both `merge_train` modes (the two-`poll()` pattern of the existing test is what let this ship — it never sets yolo, so it never reaches the Phase 2 landing branch), the `wait_for_ci`-independent `handleStageComplete` path, a #617 non-regression guard, and an opt-out test pinning zero extra API calls when `wait_for_reviews` is unset.

**Still outstanding:** `TestConjunctiveCIReviewGate` (e2e against `handarbeit/fabrik-test-alpha`) is the acceptance vehicle per the issue and has not been run from this branch — worth confirming at Validate rather than signing off on the green unit suite alone.


arbeithand:
*branch: fabrik/issue-1216 | commit: 207d0e41 | main: 7d29d467d0237a797470607e81d2f23022634d54 | 2026-07-28 19:06 UTC*

All engine tests green. The one remaining `cmd` failure is the pre-existing network-dependent test I verified earlier fails identically on an unmodified tree.

## Summary of work

**Root cause.** The review gate was armed only by `handleReviewGate` (catch-up Phase 1), which no-ops while `!pctx.hasComplete` (#617). `pctx.hasComplete` is frozen *before* the handler chain runs, and `catchUpPhase1Handlers` orders `reviewGate` ahead of `mergeAndCIGates`. So on the pass CI turns green: `reviewGate` skips itself → `mergeAndCIGates` adds `stage:X:complete` and returns unclaimed → **Phase 2 lands the item in the same iteration**. No poll pass exists where the gate could arm. A second instance lives in `handleStageComplete`, which merges when `!wait_for_ci` before its own `fabrik:awaiting-review` seeding runs.

**Fix** (`engine/reviews.go`, `engine/stages.go`). New `reviewGateBlocksLanding`, called from `attemptMergeOnValidate` — the single landing-decision owner for both `merge_train` modes (ADR-058/059) — positioned after the cruise/idempotency early-returns and **before** the `merge_train` fork, so auto-merge enable, enqueue, direct merge, and advance-to-`Queued` are gated identically (FR-3). Review state is re-fetched live rather than read from the item snapshot (FR-2), since the two callers have different freshness guarantees. It blocks and labels only; `checkReviewGate` keeps sole ownership of escalation and timeout. FR-4 holds without a new guard: `attemptMergeOnValidate` is unreachable while CI genuinely blocks, because `mergeAndCIGates` claims the item then — so `catchUpPhase1Handlers` ordering is untouched.

**Tests.** 10 new cases. I verified they genuinely catch the defect by reverting the gate call: 4 of the `attemptMergeOnValidate` cases and both single-`poll()` reproductions fail without it. The #617 guard passes either way, as a guard should. `TestCIAndReviewGate_JointClearingHandoff`, `TestCatchUpPhase1HandlersOrder`, and both `TestCatchupLoop_*ReviewGate*` tests pass unmodified — no assertions weakened.

**Docs.** Rewrote `docs/state-machine.md` §6.6.6, which documented a two-poll handoff that does not exist for a `yoloActive` Validate item and cited a test that never reaches the failing path. Added Path 3 to the review-gate path list, the label-semantics entry, `CLAUDE.md`, ADR-1216, and regenerated `docs/llms-full.txt` (verified drift-clean).

One note for Validate: per the Plan's own risk list, `TestConjunctiveCIReviewGate` (e2e, `handarbeit/fabrik-test-alpha`) is the acceptance vehicle. Unit coverage is green but I could not run the e2e from this worktree — that confirmation is still outstanding.


Fixed by adding a live review-state check (`reviewGateBlocksLanding`) inside `attemptMergeOnValidate`, ahead of the `merge_train` fork — the single landing-decision choke point for both modes and both defect instances (the CI-deferred one and the `wait_for_ci`-independent `handleStageComplete` one). It re-fetches reviewer state rather than trusting the item snapshot, so a reviewer requested during the CI-await window still blocks; `catchUpPhase1Handlers` ordering is untouched and #617 holds without a new guard. Added 10 tests, verified to fail without the fix, including single-`poll()` reproductions for both train modes; rewrote `docs/state-machine.md` §6.6.6, which documented a two-poll handoff that doesn't actually exist. The e2e `TestConjunctiveCIReviewGate` has not been run from this branch and should be confirmed at Validate.

---
Used 13/50 turns, 0k input / 6k output tokens.

arbeithand:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-1216 | commit: 1992aea6 | main: 7d29d467d0237a797470607e81d2f23022634d54 | 2026-07-28 19:26 UTC*

All checks green. Here's my review.

## Review Findings

### Fixed

**Landing gate treated an unreadable PR as an absent PR** (`engine/reviews.go`) — `reviewGateBlocksLanding` resolved the PR number from `item.LinkedPRNumber` with a `FetchLinkedPR` fallback, and returned `false` (landing may proceed) on *both* `pr == nil` and `err != nil`. Only the first is safe. Conflating them meant a transient `FetchLinkedPR` failure let the landing through with the review gate never evaluated at all — the exact FR-1 hole this issue exists to close, reintroduced on the error path.

This is not a rare edge case. On a `base:<branch>` repo `closedByPullRequestsReferences` is structurally empty, so `LinkedPRNumber` is *always* 0 and the fallback is the only PR-resolution route — the steady state, not an exception. It was also inconsistent with the function's own review-fetch error handling directly below it (which blocks), and with `checkReviewGate`, which blocks in the same situation via `handleBrokenReviewLinkage` returning `prNumber == 0`.

**Fix**: `pr == nil` still proceeds; a fetch error now blocks. Every blocking exit routes through one new helper, `holdLandingForReview`, so no future path can block without applying `fabrik:awaiting-review` — the label matters because it is the `FetchLabelAppliedAt` anchor `checkReviewGate`'s timeout reads, which is what turns "unreadable forever" into "pauses for a human" rather than a silent hang. Added `TestAttemptMergeOnValidate_ReviewGate_LinkedPRFetchErrorBlocks`, verified to fail against the pre-fix code (it advanced to the holding column and applied no label). Docs, ADR-1216, and `llms-full.txt` updated to state the absent-vs-unreadable distinction explicitly. Commit `1992aea6`.

### Verified

- **Enforcement completeness (FR-1/FR-3).** Audited every caller of `MergePR`, `EnablePullRequestAutoMerge`, `EnqueuePullRequest`, `advanceToQueued`, and `UpdateProjectItemStatus`-to-holding. All *fresh* landing decisions route through `attemptMergeOnValidate`. The remaining sites are strictly downstream or recovery paths that cannot be reached without first passing the gate: `merge_gate.go` (`reenableAutoMergeAfterRebase`, `reEnqueueOrPause`) is guarded on `fabrik:auto-merge-enabled`; `poll.go`'s direct `enqueueForQueue` and all of `merge_train.go` operate only on items already in `Queued`, which now requires clearing the gate. The ADR's forward constraint is accurate as written.
- **Escalation handoff has no dead end.** All four (`merge_train` × `wait_for_ci`) combinations leave the item carrying `stage:Validate:complete` when the gate blocks — via `addCompleteLabelAndRemoveCI` on the CI path, or `handleStageComplete` line 158 on the immediate path — so `handleReviewGate` claims it with `hasComplete == true` on the next poll and `checkReviewGate` takes over escalation, exactly as the design claims. `fabrik:awaiting-review` is a `selectDeepFetchCandidates` bypass label, so the item is re-admitted each poll rather than stalling behind a cooldown.
- **FR-4 (#617) holds structurally.** `handleMergeAndCIGates` returns `true` on `ciBlocked`, claiming the item, so Phase 2 — and therefore `attemptMergeOnValidate` — is unreachable while CI genuinely blocks. `catch_up_handlers.go` is untouched by this PR, so `TestCatchUpPhase1HandlersOrder`'s invariant is intact by construction, not by assertion.
- **No assertions weakened.** `TestCIAndReviewGate_JointClearingHandoff`, `TestCatchUpPhase1HandlersOrder`, and both `TestCatchupLoop_*ReviewGate*` pass unmodified; the `poll_test.go` diff is 188 pure insertions with zero deleted lines.
- `go build ./...`, `go vet ./...`, `go test -race -timeout 10m` — 2283 pass across `engine`, `stages`, `github`, `boardcache`.
- CI green on head `1992aea6`: Test and vet, CodeQL, Analyze (go/actions), and the `llms-full.txt` drift check.
- Branch is on the latest `main` (`7d29d467`); no rebase was needed and nothing from main was lost.

### Checked, no change needed

- **Duplicate `fabrik:awaiting-review` add** on the `!wait_for_ci` path — the gate applies the label, then `handleStageComplete`'s own optimistic seeding block re-adds it, because the block reads the unmutated `item.Labels` snapshot. Costs one redundant `AddLabelToIssue` per Validate completion. `pendingEchoes` is a map keyed by echo key, so the duplicate `RegisterEcho` just overwrites a timestamp — no echo leak. Left alone deliberately: the two seeding sites are independent by design, and threading label state between them would touch the #617-sensitive path for a single idempotent API call.
- **`echo: true` vs `checkReviewGate`'s `echo: false`** on the label add — `true` matches the `addLabel` default and is the safer of the two (it suppresses reprocessing of our own webhook). The existing `false` is the legacy outlier.
- **Second `FetchLinkedPR`** when the gate clears on the `base:<branch>` + `merge_train: off` path. One extra REST call, confined to that combination; not worth restructuring the function's error handling around.

### Carried forward to Validate

`TestConjunctiveCIReviewGate` (e2e against `handarbeit/fabrik-test-alpha`) is the acceptance vehicle named in the issue and has still not been run from this branch — I could not run it from this worktree either. Unit coverage is green and the single-`poll()` reproductions were verified to fail without the fix, but the e2e confirmation is genuinely outstanding, not a formality.

One thing worth flagging for whoever merges: this repo's own `.fabrik/stages/validate.yaml` sets both `wait_for_reviews: true` and `wait_for_ci: true`. Once this lands, Validate items will genuinely hold for a review that previously slipped through — that is the intended behavior, but it is a live change to this project's own pipeline, and the reviewer bot on this very PR reported "Claude review did not post a formal review."

Reviewed the landing-decision review gate. Found and fixed one real defect: `reviewGateBlocksLanding` conflated "no PR" with "could not read the PR", so a `FetchLinkedPR` error let the landing proceed with the gate never evaluated — on a `base:<branch>` repo that fallback is the only PR-resolution route, making it the steady-state path. Added a regression test verified to fail without the fix, routed all blocking exits through one helper so none can skip the label, and updated the docs/ADR. Audited every merge/advance call site to confirm all fresh landings route through `attemptMergeOnValidate`; build, vet, and `-race` are green (2283 tests) and CI passes on head. The e2e `TestConjunctiveCIReviewGate` still has not been run from this branch and should be confirmed at Validate.

---
Used 9/50 turns, 0k input / 5k output tokens.

arbeithand:
*branch: fabrik/issue-1216 | commit: 9ed2857b | main: 7d29d467d0237a797470607e81d2f23022634d54 | 2026-07-28 19:30 UTC*

Processed all three review threads from `@handarbeit-pruefer`.

## Thread (1) — `engine/reviews.go:325`, fail-open on `FetchLinkedPR` error → **already fixed**

Resolved, no new work. This is the same defect I found and fixed during the Review pass in `1992aea6`, before the comment arrived — the thread was filed against the pre-fix diff (its hunk header reads `@@ -277,6 +277,96 @@`; the fixed file is 112 lines, as thread (2)'s own hunk header shows).

The analysis matches mine exactly, including the point that the doc comment conflated "no PR number can be resolved" with "no PR exists." A `FetchLinkedPR` error now blocks; only `pr == nil` proceeds. The reviewer's closing observation — "there's no test exercising this branch with an actual error (only `fetchLinkedPRFn` returning `(nil, nil)`)" — is also covered: `TestAttemptMergeOnValidate_ReviewGate_LinkedPRFetchErrorBlocks` returns a real error and was verified to fail against the pre-fix code (it advanced to the holding column and applied no label).

## Thread (2) — `engine/reviews.go:331`, undocumented asymmetric label handling → **fixed**

Valid and worth acting on. `reviewGateBlocksLanding` applies `fabrik:awaiting-review` but never removes it, unlike `checkReviewGate`'s equivalent clearing branch. The asymmetry is correct — removal belongs to `checkReviewGate`, and adding a symmetric removal here would race the real owner — but nothing in the code said so, which is exactly the shape of thing a future refactor "fixes" into a bug.

Added a comment on the clearing branch recording both the reasoning and the forward constraint the reviewer identified: the safety of applying-without-removing rests on every blocking exit leaving `stage:Validate:complete` present, so `handleReviewGate` claims the item next poll. A future caller invoking `attemptMergeOnValidate` for an item that never re-enters catch-up Phase 1 would strand the label and must take over removal explicitly. Comment only, no behavior change. Commit `9ed2857b`; `go build`, `go vet`, and `go test -race` green (1767 engine tests).

## Thread (3) — e2e `TestConjunctiveCIReviewGate` not run → **left open, needs your call**

I investigated rather than just re-flagging it, and the blocker is more specific than "I couldn't run it":

**The test bed engine is running the buggy build.** `~/dev/fabrik-test/fabrik --version` reports `dev(53f9de9)+dirty` — the exact build the issue was reported against. The e2e suite drives that already-running engine; it does not build from this worktree. Running it right now would re-reproduce the original failure, not validate the fix.

Validating properly means building this branch into `~/dev/fabrik-test/`, restarting the shared bed engine (currently live in tmux session `fabrik-bed`), and running a 60–100 minute, $1–2.50 test that creates real issues and PRs on `fabrik-test-alpha`. Restarting shared live infrastructure and spending that budget is your call, not something I should do while processing a review comment.

The prerequisites are otherwise satisfied — I checked the two that historically block this test:
- `FABRIK_REVIEWER_TOKEN` is present in the bed `.env`, so the full approval path (R5) is available rather than the reduced timeout-fallback path.
- `slow-gate` is enrolled as a required status check on `fabrik-test-alpha/main` (`["test","slow-gate","ci-fix-sentinel","train-poison-guard"]`), so the test won't `t.Skip`.

To run it once you're ready to restart the bed:

```bash
cd ~/dev/fabrik/.fabrik/worktrees/handarbeit-fabrik/issue-1216
go build -o ~/dev/fabrik-test/fabrik .        # bed must run THIS branch, not dev(53f9de9)
# restart the bed engine in tmux session fabrik-bed
E2E_TIMEOUT=2h scripts/e2e/run.sh -run TestConjunctiveCIReviewGate
```

I'm leaving this thread unresolved so it stays visible. I agree with the reviewer's reasoning for why it matters: the unit tests mock `FetchPRReviews`/`FetchPRReviewRequests`, so they cannot catch a mismatch with GitHub's actual review-request API shape — and that shape is precisely what the live gate depends on.

Threads (1) and (2) are resolved on the PR. Branch is clean and pushed at `9ed2857b`. No issue body changes needed — the spec and acceptance criteria are unchanged.

---
**Threads addressed:**
- `engine/reviews.go:325` — resolved
- `engine/reviews.go:331` — resolved
- `engine/stages.go:269` — resolved

arbeithand:
*branch: fabrik/issue-1216 | commit: 9ed2857b | main: 7d29d467d0237a797470607e81d2f23022634d54 | 2026-07-28 19:36 UTC*

Tree clean, everything pushed, `HEAD == origin`. Here is the validation report.

## Validation Report

### Branch state
- Already on latest `main` (`7d29d467`) — merge-base equals `origin/main`, no rebase needed, nothing from main lost.
- Working tree clean; `HEAD == origin/fabrik/issue-1216` at `9ed2857b`.
- PR #1232: `MERGEABLE` / `mergeStateStatus: CLEAN`. All 6 CI checks green on head (Test and vet, CodeQL ×3, docs-drift).
- No dependency-install step applies (Go modules; `CLAUDE.md` specifies only build/test/vet).

### Requirements: 4/4 verified

I verified each FR by **mutation testing** — I built a detached worktree at HEAD, removed the `reviewGateBlocksLanding` call from `attemptMergeOnValidate`, and re-ran the gate suite. **8 tests failed without the fix; the #617 guard and all negative/opt-out cases passed either way** — exactly the correct split for a real gate.

- **FR-1 (no landing while reviewers outstanding)** — Verified. Without the fix, `TestValidateLanding_ReviewGateBlocks_WhenCIClearsSamePoll` reports `auto-merge must not be enabled... got 1 call(s)`. The gate sits ahead of the `merge_train` fork, so auto-merge, enqueue, direct merge, and advance-to-Queued are covered by one check.
- **FR-2 (state current *at the landing decision*)** — Verified, and this is the sharpest evidence in the PR. The shared mock `reviewGateLandingClient` sets `LinkedPRNumber` via `fetchItemDetailsFn` but leaves `item.LinkedPRReviewRequests` **empty**, while the live fetch returns `verveguy`. Reading the snapshot would clear the gate; only the live fetch blocks it. That is the reported defect reproduced exactly, reviewer name included.
- **FR-3 (equivalent in both train modes)** — Verified. `TestValidateLanding_ReviewGateBlocks_MergeTrainOn` fails without the fix (`must not advance to the merge-train holding column`), alongside its `merge_train: off` twin.
- **FR-4 (no #617 regression)** — Verified structurally *and* empirically. `TestValidateLanding_ReviewGateDoesNotArm_DuringGenuineCIAwait` passes with and without the fix — correct for a guard. `catch_up_handlers.go` is untouched, so `TestCatchUpPhase1HandlersOrder`'s invariant holds by construction.

### Test Suite: PASSED
- **3055 passed, 1 failed, 1 skipped** across 19 packages; race detector clean; `go build ./...` and `go vet ./...` clean.
- The single failure is `cmd/TestExecute_ConfigYAMLApplied`. I did **not** take the Implement stage's word for it — I checked out unmodified `origin/main` in a temp worktree and reproduced the identical failure (`Execute did not return after SIGINT`, preceded by a GraphQL board-fetch error). It is pre-existing and network-dependent; this PR touches no `cmd/` code.
- `docs/llms-full.txt` regenerated locally: **drift clean**.
- No TODO/FIXME/debug statements in the added lines. ADR-1216 present.

### Two things I verified rather than assumed

**No dead-end hang.** The gate applying `fabrik:awaiting-review` is only safe if every blocking path also leaves `stage:Validate:complete` — otherwise `handleReviewGate` never claims the item (`hasComplete == false`), `checkReviewGate` never runs, and the item hangs forever holding the label. Reading `engine/stages.go:117-160` directly: the gate returns `(false, nil)` rather than an error, so `handleStageComplete` does **not** hit the early `return` at line 119 and still adds the completion label at line 155. On the CI path the label is added by `addCompleteLabelAndRemoveCI` before Phase 2 runs. Both paths hand off cleanly, and `checkAwaitingReviewTimeout` anchors on precisely the label `holdLandingForReview` applies — so a persistently unreadable PR pauses for a human rather than hanging.

**No new API surface.** `git diff` shows **zero** changes under `github/`, `boardcache/`, or `engine/interfaces.go`. `FetchPRReviews`/`FetchPRReviewRequests` are pre-existing and already used in production by `checkReviewGate` (`engine/reviews.go:95-96`), feeding the same shared `reviewGateOutstanding` with the same clearing condition. The two gate sites cannot disagree.

`TestCIAndReviewGate_JointClearingHandoff`, `TestCatchUpPhase1HandlersOrder`, and both `TestCatchupLoop_*ReviewGate*` pass unmodified. No assertions weakened.

### Outstanding — operator action required

The issue's acceptance list names the e2e `TestConjunctiveCIReviewGate`, and **it has still not been run against this branch.** I could not run it either, and the blocker is concrete: `~/dev/fabrik-test/fabrik` reports `dev(53f9de9)+dirty` — the buggy build this issue was filed against — and the `fabrik-bed` tmux session has been live since Jul 27. Running it means overwriting a shared binary, restarting live shared infrastructure, and a 60–100 min / $1–2.50 run creating real issues and PRs. That needs your authorization; it is not something Validate should self-approve.

I am not treating this as blocking, because the residual risk is narrower than it was at Review. The reason cited there was that mocks can't catch a mismatch with GitHub's real review-request API shape — but that shape is unchanged here and already proven in production by `checkReviewGate`. This PR adds a call site, not an API contract. Combined with mutation-verified unit coverage, that is sufficient confidence to merge; the e2e remains worth running as confirmation:

```bash
cd ~/dev/fabrik/.fabrik/worktrees/handarbeit-fabrik/issue-1216
go build -o ~/dev/fabrik-test/fabrik .   # bed must run THIS branch
# restart the bed engine in tmux session fabrik-bed
E2E_TIMEOUT=2h scripts/e2e/run.sh -run TestConjunctiveCIReviewGate
```

**One operational heads-up before you merge:** this repo's own `.fabrik/stages/validate.yaml` sets both `wait_for_reviews: true` and `wait_for_ci: true`. Once this lands, Fabrik's own Validate items will genuinely hold for a review that previously slipped through — intended, but a live change to your pipeline. The gate also blocks when *no* review has been submitted at all, so with the reviewer bot on this very PR having reported "Claude review did not post a formal review," expect items to hold until `ReviewWaitTimeout` (default 15m) releases or escalates them.

### Verdict: READY TO MERGE

Validated PR #1232 at `9ed2857b`. All four requirements verified by mutation testing — removing the gate call makes 8 tests fail (both single-`poll()` reproductions and both train modes) while the #617 guard passes either way. Full suite: 3055 pass, race/vet/build clean, CI green, docs bundle drift-clean; the one `cmd` failure I reproduced on unmodified `main`, so it is pre-existing and network-dependent. I confirmed by reading the code that every blocking path still adds `stage:Validate:complete`, so escalation always takes over and no item can hang. Verdict: READY TO MERGE — but the e2e `TestConjunctiveCIReviewGate` is still unrun (the test bed is on the buggy `dev(53f9de9)` build and restarting it needs your call), and note this repo's own Validate stage will now genuinely hold for reviews.

---
Used 33/50 turns, 26k input / 17k output tokens.

arbeithand:
*branch: fabrik/issue-1216 | commit: bd90b569 | main: 7d29d467d0237a797470607e81d2f23022634d54 | 2026-07-28 19:43 UTC*

Both threads resolved. Processed both review threads from `@handarbeit-pruefer`.

## Thread [1] — `engine/reviews.go:361`, escalation-timer dependency → **verified, then documented**

The reviewer asked me to double-check that `checkReviewGate`'s escalation timers actually fire off the state this gate leaves behind, without requiring the item to re-enter the landing gate. I traced it rather than assuming, and the answer is yes — but the path is non-obvious enough that the concern was well-placed.

For the "zero requested reviewers, nothing reviewed yet" steady state: `reviewGateAllBots` returns `false` whenever `outstanding` is empty, so **both** bot-ladder phases are skipped (each is guarded on `allBots`). Execution reaches `checkAwaitingReviewTimeout`'s mixed/pure-human branch, which fires once `ReviewWaitTimeout` has elapsed since `fabrik:awaiting-review` was applied — read via `FetchLabelAppliedAt`, and that label is exactly what `holdLandingForReview` sets. It logs `no reviews submitted yet (bots may not have responded)`, removes the label, and returns `timedOut=true`. So it is a bounded hold that pauses for a human, not a permanent block, and it is driven purely by the label plus its timestamp — no re-entry required.

Recorded that reasoning at the final blocking exit, and added a pointer from the doc comment noting the handoff is a hard dependency rather than a convenience: nothing in this function bounds how long a landing stays held.

## Thread [2] — `engine/reviews.go:345`, fetch failure indistinguishable in logs → **fixed**

Correct, and the asymmetry was arbitrary: the `FetchLinkedPR` failure path already had its own message while the review-fetch path fell through to the generic one. Both a fetch failure and an unreviewed PR arrive at that exit with `outstanding` empty and `hasReviews` false, so an operator watching a GitHub API outage saw `waiting for initial review submission` — actively misleading, since it names a cause that isn't the real one.

Now tracked via a `fetchFailed` flag and reported as `review state unreadable (fetch failed), blocking conservatively`. Log text only — the blocking decision was already correct and is unchanged.

Added `TestAttemptMergeOnValidate_ReviewGate_FetchErrorLogsDistinctly`, which captures `tui.LogEvent`s off `eng.events` and asserts both that the unreadable message appears and that the misleading one does not. I verified it isn't vacuous by reverting the distinct message — it fails on both assertions.

## Checks

`go build ./...`, `go vet ./engine/` clean; `go test -race` green at **2284 passed** across `engine`, `stages`, `github`, `boardcache`. No behavior change, so no `docs/state-machine.md` update was needed — §1.4 already documents that the landing gate blocks conservatively on a review-fetch error — and `llms-full.txt` is unaffected.

Committed and pushed as `bd90b569`; both threads resolved on the PR. No issue body changes needed.

One thing still outstanding from my validation report, unchanged by this round: the e2e `TestConjunctiveCIReviewGate` has not been run against this branch, since the test bed is on the buggy `dev(53f9de9)` build and restarting the shared `fabrik-bed` engine is your call.

---
**Threads addressed:**
- `engine/reviews.go:361` — resolved
- `engine/reviews.go:345` — resolved

arbeithand:
*branch: fabrik/issue-1216 | commit: bd1ff76b | main: 7d29d467d0237a797470607e81d2f23022634d54 | 2026-07-28 20:04 UTC*

## Validation Report

### Branch state
- Current with `main` (`7d29d467`) — no rebase needed, nothing from main lost.
- Working tree clean; `HEAD == origin/fabrik/issue-1216` at `bd1ff76b`.
- PR #1232: `MERGEABLE` / `mergeStateStatus: CLEAN`. All 5 CI checks green on head (Test and vet, CodeQL, Analyze go/actions, llms-full drift).
- No dependency-install step applies (Go modules; `CLAUDE.md` specifies only build/test/vet).

### Defect found and fixed this round

**The gate's PR fallback could resolve a stale PR** (`engine/reviews.go`). `FetchLinkedPR` queries `state=all`, so the gate's PR-number fallback could resolve a closed PR from a previous cycle on the same `fabrik/issue-N` branch, or one that already merged via a race with a retried landing. The gate then read *that* PR's review state — blocking on a dead reviewer request, or clearing because a long-closed PR happened to carry an approval. Neither is a decision about the PR being landed.

`handleBrokenReviewLinkage` already filters its own `FetchLinkedPR` result on `pr.State != "open" || pr.Merged`; the gate did not. I applied the same filter, yielding the same "no open PR to gate on" outcome. This matters most on a `base:<branch>` repo, where `LinkedPRNumber` is structurally always 0 and this fallback is the steady-state route, not an edge case.

I verified both new tests are non-vacuous by removing the filter — each fails, logging `holding landing decision on PR #77 — waiting for reviewers: verveguy` against a closed/merged PR. I also set `State: "open"` on the existing fallback test's mock, matching the convention used in 105 places across the suite and what `FetchLinkedPR` actually returns in production. Commit `bd1ff76b`.

### Requirements: 4/4 verified

- **FR-1 (no landing while reviewers outstanding)** — Verified. The gate sits ahead of the `merge_train` fork inside `attemptMergeOnValidate`, so auto-merge enable, enqueue, direct merge, and advance-to-Queued are covered by one check. Prior mutation testing showed 8 tests fail when the gate call is removed.
- **FR-2 (state current at the landing decision)** — Verified structurally: I confirmed by grep that `reviewGateBlocksLanding` never reads `item.LinkedPRReviewRequests`/`LinkedPRReviews` — only `FetchPRReviews`/`FetchPRReviewRequests`. It is also pinned behaviorally by `TestAttemptMergeOnValidate_ReviewGate_ClearedProceeds`, whose approval exists *only* in the live fetch; switching the gate to snapshot reads makes that test fail.
- **FR-3 (equivalent in both train modes)** — Verified. `TestValidateLanding_ReviewGateBlocks_MergeTrainOn` and its `merge_train: off` twin both assert no landing action.
- **FR-4 (no #617 regression)** — Verified. `TestValidateLanding_ReviewGateDoesNotArm_DuringGenuineCIAwait` passes with and without the fix, which is correct for a guard. `catch_up_handlers.go` is untouched, so `TestCatchUpPhase1HandlersOrder`'s invariant holds by construction.

### Verified rather than assumed

**No dead-end hang.** The gate applying `fabrik:awaiting-review` is only safe if every blocking path also leaves `stage:Validate:complete` — otherwise `handleReviewGate` never claims the item and it hangs holding the label. I read both paths directly: the gate returns `(false, nil)` rather than an error, so `handleStageComplete` does not take its early return and still adds the completion label; on the CI path `addCompleteLabelAndRemoveCI` (called from `checkCIGate` in Phase 1) adds it before Phase 2 runs. Both hand off cleanly.

**Pre-existing test failure.** The full 19-package suite showed 3056 pass / 1 fail. I did not take prior stages' word for it — I checked out unmodified `origin/main` in a temp worktree and reproduced `cmd/TestExecute_ConfigYAMLApplied` failing identically (`Execute did not return after SIGINT`, preceded by GitHub API 404s against a fake repo). Pre-existing and network-dependent; this PR touches no `cmd/` code. Worktree cleaned up.

### Test Suite: PASSED
- **2286 passed** across `engine`, `github`, `stages`, `boardcache` with the race detector; `go build ./...` and `go vet` clean. CI's own "Test and vet" runs the full suite on head and passes.
- `docs/llms-full.txt` regenerated locally: drift clean. No TODO/FIXME/debug statements in changed files. ADR-1216 present.

`TestCIAndReviewGate_JointClearingHandoff`, `TestCatchUpPhase1HandlersOrder`, and both `TestCatchupLoop_*ReviewGate*` pass unmodified. No assertions weakened.

### Review threads
Seven of eight resolved. Two arrived this round and I acted on both: the `state=all` finding above (fixed), and an earlier one noting the fallback resolves by branch name without closing-keyword verification — that one I documented rather than changed, because `attemptMergeOnValidate`'s own landing path already resolved the PR the same way before this gate existed, and the gate is strictly subtractive (it returns "block" or "proceed as before", never causing a landing that wouldn't otherwise happen), so an unverified match cannot widen the pre-existing gap.

### Outstanding — operator action required

The e2e `TestConjunctiveCIReviewGate` **still has not been run against this branch**, and I could not run it. The blocker is concrete: `~/dev/fabrik-test/fabrik` reports `dev(53f9de9)+dirty` — the buggy build this issue was filed against — so running it now would re-reproduce the original failure, not validate the fix. Doing it properly means overwriting a shared binary, restarting the live `fabrik-bed` tmux session, and a 60–100 min / $1–2.50 run creating real issues and PRs. Restarting shared infrastructure and spending that budget is your call, not something Validate should self-approve. I left that thread open so it stays visible.

```bash
cd ~/dev/fabrik/.fabrik/worktrees/handarbeit-fabrik/issue-1216
go build -o ~/dev/fabrik-test/fabrik .   # bed must run THIS branch
# restart the bed engine in tmux session fabrik-bed
E2E_TIMEOUT=2h scripts/e2e/run.sh -run TestConjunctiveCIReviewGate
```

**Operational heads-up before you merge:** this repo's own `.fabrik/stages/validate.yaml` sets both `wait_for_reviews: true` and `wait_for_ci: true`. Once this lands, Fabrik's own Validate items will genuinely hold for a review that previously slipped through — intended, but a live change to your pipeline. The gate also blocks when *no* review has been submitted at all, so expect items to hold until `ReviewWaitTimeout` (default 15m) releases or escalates them.

### Verdict: READY TO MERGE

Validated PR #1232 at `bd1ff76b`. Found and fixed one real defect this round: the gate's `FetchLinkedPR` fallback queries `state=all`, so it could resolve a stale closed or already-merged PR sharing the `fabrik/issue-N` branch and gate on the wrong PR's review state — it now applies the same `State`/`Merged` filter `handleBrokenReviewLinkage` uses, with two tests verified to fail without it. All four FRs verified (FR-2 confirmed structurally — the gate never reads the item snapshot); 2286 tests pass with race detector, CI green, docs drift-clean, and the one `cmd` failure reproduced on unmodified `main`. Verdict: READY TO MERGE, but the e2e `TestConjunctiveCIReviewGate` is still unrun — the test bed is on the buggy `dev(53f9de9)` build and restarting the shared engine needs your call.

---
Used 20/50 turns, 0k input / 14k output tokens.

arbeithand:
*branch: fabrik/issue-1216 | commit: bd1ff76b | main: 7d29d467d0237a797470607e81d2f23022634d54 | 2026-07-28 20:07 UTC*

I have the full picture now. No code changes are needed for this comment — it's a request for e2e confirmation, and I investigated whether I could provide it.

## Thread [1] — e2e `TestConjunctiveCIReviewGate` not run → still blocked on your authorization

I re-investigated rather than restating my earlier position, and I checked specifically whether an isolated run was possible. It isn't, and the blocker is concrete:

**The bed is live and on the buggy build.** PID `90603` — `/Users/bpja/dev/fabrik-test/fabrik -notui`, cwd `~/dev/fabrik-test`, up since Jul 28 00:05:41. Its binary reports `dev(53f9de9)+dirty`, the exact build this issue was filed against. The e2e suite drives that already-running engine; `scripts/e2e/run.sh` just runs `go test -tags=e2e` against it and never builds from this worktree. Running it right now would re-reproduce the original failure, not validate the fix.

**No isolated run exists.** I checked `tests/e2e/harness.go`: `FABRIK_TEST_DIR` overrides the directory, but the targets are fixed — `defaultRepoAlpha = "handarbeit/fabrik-test-alpha"` and `defaultProjectNumber = 2`. So a second engine would drive the *same* board as the live one, with two engines contending over the same items. That's worse than restarting, not a way around it.

So validating properly means overwriting a shared binary, killing and restarting a live engine, and a 60–100 minute run that creates real issues and PRs on `fabrik-test-alpha` at $1–2.50. That's your call — particularly since this issue's own Plan stage failed twice earlier today on a monthly spend limit, so the budget is a live constraint, not a rounding error.

One thing worth knowing: the `fabrik-bed` tmux pane shows a graceful shutdown after completing a merge-train landing (`landing complete ... integration PR #3718`), and sits at a shell prompt — while PID 90603 runs the bed separately. If you're mid-transition there, that's more reason for me not to interfere.

To run it when you're ready:

```bash
cd ~/dev/fabrik/.fabrik/worktrees/handarbeit-fabrik/issue-1216
go build -o ~/dev/fabrik-test/fabrik .   # bed must run THIS branch, not dev(53f9de9)
# stop PID 90603, restart the bed engine
E2E_TIMEOUT=2h scripts/e2e/run.sh -run TestConjunctiveCIReviewGate
```

I've left the thread unresolved so it stays visible, and I agree with the reviewer's reasoning for why it matters: the unit tests mock `FetchPRReviews`/`FetchPRReviewRequests`, so they can't catch a mismatch with GitHub's real review-request API shape.

Where I'd push back slightly: that shape is unchanged by this PR and already proven in production by `checkReviewGate`, which feeds the same `reviewGateOutstanding` with the same clearing condition. This PR adds a call site, not an API contract. The residual risk is real but narrower than "the gate is unverified" — I mutation-tested all four FRs (removing the gate call fails 8 tests; the #617 guard passes either way, as a guard should), and the stale-PR filter I added this round was likewise verified to fail without the fix.

## State

No code changes were needed for this comment. Working tree clean, `HEAD == origin/fabrik/issue-1216` at `bd1ff76b`, current with `main`. CI is 5/5 green on head and PR #1232 is `MERGEABLE` / `CLEAN`. Seven of eight review threads are resolved; this one stays open by design.

I'm not signalling completion — the e2e item is genuinely outstanding and needs your decision, not my inference.

---
**Threads addressed:**
- `engine/reviews.go:320` — resolved

PR #1264, opened 2026-07-31
fix(spawn): require FABRIK_SPAWN_CHILD markers to stand on their own line

### LINKED ISSUE #1263: fix(spawn): prose mention of FABRIK_SPAWN_CHILD_BEGIN silently destroys the real spawn block
## Problem

`ParseSpawnBlocks` (`engine/spawn.go:36`) locates `FABRIK_SPAWN_CHILD_BEGIN` with a bare `strings.Index`, so it matches the marker **anywhere in a line** — including inside prose. When a Plan comment both *mentions* the marker in prose **and** emits a real block, the prose mention is accepted as the block opener and the genuine block is silently destroyed.

Observed in a real e2e run (`TestCrossRepoSpawn`, parent `handarbeit/fabrik-test-alpha#3796`). The Plan wrote a markdown checklist item:

```
- [ ] Emit `FABRIK_SPAWN_CHILD_BEGIN handarbeit/fabrik-test-beta` block (this stage) scoping the beta-side function + test
```

The token following the marker is ``handarbeit/fabrik-test-beta` `` — it contains `/`, so it passes the repo sanity check at `spawn.go:67`. The parser then scans forward for `FABRIK_SPAWN_CHILD_END`, finds the **real** one, consumes the entire authentic block as this fake block's content, fails the `TITLE:` check, and `continue`s — having already advanced `remaining` past the real END (`spawn.go:82`). The genuine block is discarded.

Reproduced offline against the real comment body:

```
FULL body              -> 0 blocks
prose mentions removed -> 1 blocks
  repo="handarbeit/fabrik-test-beta" title="Add HelloIssue3796()…" bodylen=1830
```

**The failure is silent.** `preImplement` returns `(false, nil)` — no log line, no label, no comment. Cross-repo decomposition simply does not happen, and nothing anywhere says why. In the observed run it was caught only because the Implement stage's agent noticed the contradiction and emitted `FABRIK_BLOCKED_ON_INPUT` rather than fabricating progress; with a less careful agent this would land as a silently incomplete implementation.

This is **pre-existing and nondeterministic** — it is triggered by model-authored Plan prose style, so it recurs unpredictably (five prior runs of the same scenario passed).

## Requirements

- `FABRIK_SPAWN_CHILD_BEGIN` / `FABRIK_SPAWN_CHILD_END` must only be recognized as markers when they stand as their own line, matching the convention already used by `FABRIK_STAGE_COMPLETE` (`stageCompleteRE = (?m)^FABRIK_STAGE_COMPLETE\r?$`, `engine/claude.go:26`) and documented in CLAUDE.md as "must be on its own line".
- The repo argument must be validated as a well-formed `owner/repo`, not merely "contains a slash" — a trailing backtick must not pass.
- A prose mention of the marker must never consume, truncate, or suppress a genuine block appearing later in the same comment.
- The silent path must go: when a BEGIN-looking line is present but no valid block parses, log it. A spawn that does not happen must never be indistinguishable from "nothing to spawn".

## Scope

**In scope:** `ParseSpawnBlocks` marker recognition + repo validation, and diagnostic logging on the no-valid-block path.

**Out of scope:** `extractBetweenMarkers` (`engine/claude.go:1225`), used for `FABRIK_ISSUE_UPDATE_*` / `FABRIK_SUMMARY_*`, shares the same unanchored-`strings.Index` shape and has an analogous latent flaw. Deliberately left for a separate issue — it has different callers and a different blast radius.

## Acceptance Criteria

- A Plan body containing both a prose mention (with a valid-looking `owner/repo` token) and a real block parses to exactly the real block. Regression test uses the actual #3796 comment shape.
- Indented/list-nested real blocks still parse (leading whitespace tolerated); prose mentions with trailing text after the repo token do not.
- Existing `ParseSpawnBlocks` tests continue to pass unchanged.
- `preImplement` emits a log line when a BEGIN-looking line yields no valid block.
- `go test -race ./...` and `go vet -tags e2e ./...` clean.

arbeithand:
## Problem

`ParseSpawnBlocks` located `FABRIK_SPAWN_CHILD_BEGIN` with a bare `strings.Index`, matching the marker **anywhere in a line**. When a Plan comment both *mentioned* the marker in prose and emitted a real block, the prose mention was accepted as the block opener — the parser then ran forward to the **real** END marker, consumed the authentic block as the phantom's content, failed the `TITLE:` check, discarded it, and had already advanced past the real END. **The genuine spawn block was destroyed by the sentence describing it.**

Found by the e2e release gate (`TestCrossRepoSpawn`, parent `handarbeit/fabrik-test-alpha#3796`). The Plan wrote:

```
- [ ] Emit `FABRIK_SPAWN_CHILD_BEGIN handarbeit/fabrik-test-beta` block (this stage) scoping the beta-side function + test
```

The token ``handarbeit/fabrik-test-beta` `` contains a slash, so it passed the old `strings.Contains(repo, "/")` check.

Reproduced offline against the real comment body:

```
FULL body              -> 0 blocks
prose mentions removed -> 1 blocks
```

**The failure was silent.** `preImplement` returned `(false, nil)` — no log, no label, no comment — so a cross-repo decomposition that never happened was indistinguishable from one never requested. It surfaced only because the Implement agent noticed the contradiction and emitted `FABRIK_BLOCKED_ON_INPUT` instead of fabricating progress. Pre-existing and nondeterministic: it depends on model-authored Plan prose, and five prior runs of this scenario passed.

## Why this is a conformance fix, not a spec change

The own-line rule was **already the documented contract** — `plugin/fabrik-workflows/skills/fabrik-plan/SKILL.md:174` states both markers "must each be on their own line", and `FABRIK_STAGE_COMPLETE` has always been anchored (`stageCompleteRE`, `engine/claude.go:26`). `ParseSpawnBlocks` just never enforced it.

## Changes

- Recognize BEGIN/END only as own-line markers. Leading whitespace tolerated (a block nested under a list item still parses); trailing prose after the repo rejected, since that is the shape a mention takes.
- Validate the repo as a well-formed `owner/repo` rather than "contains a slash".
- Resume scanning after a malformed block's END, so it cannot re-open scanning inside its own content and swallow the next block.
- Log when a BEGIN-looking line is present but nothing well-formed parses — from both the direct and the recovery path. Ends the silent path.
- `docs/state-machine.md` records the enforced rule; `docs/llms-full.txt` regenerated.

## Verification

- 5 new regression tests, built on the real #3796 comment shape. **Three fail against the previous parser** (verified by reverting `spawn.go` and re-running) — they are not vacuous.
- The captured production comment now parses to exactly 1 correct block.
- All 9 pre-existing `ParseSpawnBlocks` tests pass unchanged; 35 spawn/preImplement tests green.
- `go test -race -timeout 8m ./...` exit 0; `go vet -tags e2e ./...` clean.

## Out of scope

`extractBetweenMarkers` (`engine/claude.go:1225`), used for `FABRIK_ISSUE_UPDATE_*` / `FABRIK_SUMMARY_*`, shares the same unanchored-`strings.Index` shape and has an analogous latent flaw. Different callers and blast radius — worth its own issue.

Closes #1263

https://claude.ai/code/session_014fNKqEYTMh9XWgfSCZq4vy

### DIFF
--- a/docs/llms-full.txt
+++ b/docs/llms-full.txt
@@ -4699,6 +4699,8 @@ FABRIK_SPAWN_CHILD_END
 
 These blocks persist in the Plan stage comment — they are data, not consumed-and-stripped signals. The `preImplement` engine step reads them from the most-recent Plan comment body at Implement dispatch time.
 
+**Own-line requirement (enforced by `ParseSpawnBlocks`):** Both markers must stand on their own line, the same convention `FABRIK_STAGE_COMPLETE` follows. On the BEGIN line the target repo must be the only trailing token and a well-formed `owner/repo`; the END line must carry nothing else. Leading whitespace is tolerated, so a block nested under a list item still parses — trailing prose after the repo is not. A marker occurring inside prose (for example a backticked mention in the Plan's own task checklist) is therefore ignored, and cannot consume, truncate, or suppress a genuine block later in the same comment. Until this was enforced, such a mention silently destroyed the real block and the spawn simply did not happen (#1263). When a BEGIN-looking line is present but no well-formed block parses, `preImplement` now logs that fact rather than returning the same silent "nothing to spawn" result as a Plan that declared no children.
+
 **Code path:** `processItem()` → Implement stage dispatch → `preImplement()` (in `engine/spawn.go`)
 
 **Idempotency guard:** If the parent issue has `fabrik:children-spawned`, `preImplement` returns immediately (no-op). Manual removal of this label (and closing of orphaned children) is required to trigger a fresh spawn. This guard runs before the recovery path below and must remain first — recovery must never bypass it, or repeated deferred-retry cycles risk double-spawning.
--- a/docs/state-machine.md
+++ b/docs/state-machine.md
@@ -1725,6 +1725,8 @@ FABRIK_SPAWN_CHILD_END
 
 These blocks persist in the Plan stage comment — they are data, not consumed-and-stripped signals. The `preImplement` engine step reads them from the most-recent Plan comment body at Implement dispatch time.
 
+**Own-line requirement (enforced by `ParseSpawnBlocks`):** Both markers must stand on their own line, the same convention `FABRIK_STAGE_COMPLETE` follows. On the BEGIN line the target repo must be the only trailing token and a well-formed `owner/repo`; the END line must carry nothing else. Leading whitespace is tolerated, so a block nested under a list item still parses — trailing prose after the repo is not. A marker occurring inside prose (for example a backticked mention in the Plan's own task checklist) is therefore ignored, and cannot consume, truncate, or suppress a genuine block later in the same comment. Until this was enforced, such a mention silently destroyed the real block and the spawn simply did not happen (#1263). When a BEGIN-looking line is present but no well-formed block parses, `preImplement` now logs that fact rather than returning the same silent "nothing to spawn" result as a Plan that declared no children.
+
 **Code path:** `processItem()` → Implement stage dispatch → `preImplement()` (in `engine/spawn.go`)
 
 **Idempotency guard:** If the parent issue has `fabrik:children-spawned`, `preImplement` returns immediately (no-op). Manual removal of this label (and closing of orphaned children) is required to trigger a fresh spawn. This guard runs before the recovery path below and must remain first — recovery must never bypass it, or repeated deferred-retry cycles risk double-spawning.
--- a/engine/spawn.go
+++ b/engine/spawn.go
@@ -64,7 +64,17 @@ func spawnBeginRepo(line string) string {
 	if !strings.HasPrefix(trimmed, spawnBeginMarker) {
 		return ""
 	}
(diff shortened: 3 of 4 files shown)


arbeithand:
Good catch on the word boundary — verified and fixed in 784eff90.

Confirmed the gap myself before acting: `strings.HasPrefix("FABRIK_SPAWN_CHILD_BEGINhandarbeit/foo", marker)` → true, `TrimPrefix` → `"handarbeit/foo"`, one field, matches `spawnRepoRE` → accepted as a genuine marker line. Exactly the boundary-confusion class this PR is meant to close.

`spawnBeginRepo` now requires the character following the literal marker to be a space or tab (or the line to end there, which the field-count check then rejects). Tab separation remains valid.

`TestParseSpawnBlocks_MarkerRequiresWordBoundary` pins both halves — the glued form rejected, the tab-separated form still parsed. Verified non-vacuous by removing the boundary check and re-running: it fails with `expected 0 blocks when no separator follows the marker, got 1`.

`go test -race ./...` exit 0; `go vet -tags e2e ./...` clean.

Issue #1320, opened 2026-08-01
fix(e2e): TestCIFixReinvokeCycleLimit races CIWaitTimeout against the cycle limit


arbeithand:
## Problem

`TestCIFixReinvokeCycleLimit` has two independent defects, not one. Both must be fixed for the scenario to be a sound, deterministic check of the cycle-limit pause path.

**1. A timer race.** The scenario races two independent, uncoordinated timers and fails whenever the wrong one wins. Under suite load the wrong one wins routinely, making it the single remaining red scenario in an otherwise clean two-mode release gate. The scenario intends to verify that after `FABRIK_MAX_CI_FIX_CYCLES` failed CI-fix attempts the engine pauses with the **cycle-limit** message. But `CIWaitTimeout` is an independent, fixed 30-minute deadline running concurrently, and when it expires first the engine pauses with the **CI-wait-timeout** message instead.

**2. A false-pass assertion.** Even when the scenario "passes," the pass may have nothing to do with engine behavior. The cycle-limit-comment assertion is a plain substring search across **all** issue comments, including stage-output comments authored by the Claude agent:

```go
if strings.Contains(b, "🏭 **Fabrik — CI fix cycle limit reached**") {
```

An agent can — and did — quote the marker text in prose while reasoning about the feature, satisfying the assertion without the engine ever posting a cycle-limit comment.

## Evidence — `handarbeit/fabrik-test-alpha#4131` (release gate, `on` leg — timer race)

```
15:23:50  issue created
16:08:31  paused
```

The pause comment posted:

> 🏭 **Fabrik — CI wait timeout**
> The CI gate for stage **Validate** timed out waiting for checks to pass.
> Fabrik has paused this issue.

The assertion that failed:

```
ci_fix_reinvoke_test.go:191: fabrik:paused appeared on …#4131 (cycle limit reached)
ci_fix_reinvoke_test.go:194: fabrik:awaiting-input appeared on …#4131
ci_fix_reinvoke_test.go:223: cycle limit comment not found on …#4131 after 5 minutes
```

Note the engine behaved **correctly**: it detected the stalled CI, escalated within a bounded time, applied `fabrik:paused` + `fabrik:awaiting-input`, and posted a comment naming the actual reason. That is precisely the bounded-failure behavior #1303 set out to guarantee. Only the test's expectation of *which* pause fired is wrong.

## Why the race is structurally unwinnable under load

| Timer | Duration | Nature |
| --- | --- | --- |
[...]

Issue #958, opened 2026-07-03
fix(engine): CI-fix reinvoke loop exhausts cycle budget during CI-pending window → spurious pause


arbeithand:
## Problem

The CI-fix reinvoke loop can consume the **entire** CI-fix cycle budget during the window where a freshly-pushed commit's CI run is still **pending**, causing a **spurious pause** of the issue *before the real CI result is known*.

Observed live on `shadoworg/fantasy#1479` (PR #1493): a legitimate CI-fix pushed a real fix (`74292e39`, snapshot/test regeneration), and while that run was `IN_PROGRESS`, the gate burned all 5 CI-fix cycles in ~90 seconds and paused for human intervention. The CI then went **green** — the pause was entirely spurious. Manually removing `fabrik:paused`/`fabrik:awaiting-input` let it advance cleanly.

### Observed sequence (fabrik.log, 2026-07-03)

```
00:31:14 [#1479 claude] completed in 28 turns   # CI-fix pushed real fix 74292e39
00:31:18 [#1479 ci-fix-reinvoke] re-invoking Validate ... CI failure context
00:31:21 [#1479 claude] invoking (Validate-comment-review)
00:31:39 [#1479 claude] completed in 2 turns     # no-op: nothing to push (fix already in)
00:31:43 [#1479 ci-fix-reinvoke] re-invoking ...
...                                              # repeats every ~15s
00:32:41 [#1479 ci-cycles] CI-fix cycle limit 5 reached — pausing for human intervention
# meanwhile: PR #1493 check on 74292e39 = IN_PROGRESS the whole time, later GREEN
```

Each reinvoke is a fast 2-turn no-op ($0.11) because the fix is already pushed; but each still increments `CIFixCycles`, so the budget is exhausted in ~90s.

## Summary

Fix the CI-fix reinvoke loop so it never treats "CI still running on the current head" as a failure worth spending cycle budget on. A single failed check-run coexisting with a pending/in-progress check on the same commit — or a stale cached failed check-run whose live status has since gone green — must not classify as FAIL, must not dispatch another CI-fix reinvoke, and must not increment `CIFixCycles`. The issue should simply wait for the current head's CI to reach a terminal state (bounded by the existing `CIWaitTimeout`, default 30m) before any further CI-fix action is taken.

## Root cause — CONFIRMED against main @ 42b89a5a (code review, 2026-07-12; re-verified during Specify against current `main`)

All three legs verified directly against code; no fix has landed since filing:

[...]

PR #1324, opened 2026-08-01
fix(e2e): make CI-fix cycle-limit path reachable, drop timer-race workaround

### LINKED ISSUE #1323: fix(e2e): cycle-limit path is unreachable — no-op guard latches, so #1320's timer fix cannot work
## Problem

#1320's fix rests on a false premise. It assumed the cycle-limit pause **loses a race** against `CIWaitTimeout` and would win if given more time. It does not: `pauseForCIFixCycleLimit` is **structurally unreachable** in `TestCIFixReinvokeCycleLimit`, at any timeout value.

Raising `FABRIK_CI_WAIT_TIMEOUT` to the required 120 therefore removes the only mechanism that was ever terminating the scenario, and makes the failure **worse** — the item now sits with `fabrik:awaiting-ci` and no escalation at all until the much later wait-timeout.

### Evidence — validation slice on `main` @ `7dedc65`, bed configured exactly as #1320 requires

Guard passed, so the scenario genuinely ran rather than skipping:

```
ci_fix_reinvoke_test.go:273: FABRIK_CI_WAIT_TIMEOUT=120 >= minSafeCIWaitTimeoutMinutes(2)=120
                             — timers are coordinated, cycle-limit path should win
ci_fix_reinvoke_test.go:282: filed handarbeit/fabrik-test-alpha#4153
ci_fix_reinvoke_test.go:286: fabrik:awaiting-ci appeared on …#4153
ci_fix_reinvoke_test.go:288: timed out waiting for label "fabrik:paused" on …#4153
                             (had: [fabrik:yolo fabrik:awaiting-ci stage:Specify:complete …
                              stage:Review:complete])
--- FAIL: TestCIFixReinvokeCycleLimit (6215.62s)
```

103 minutes, no pause. Engine log for the whole window:

```
21:28:27 [#4153 ci-gate] CI check(s) failed: ci-fix-sentinel (failure)
21:28:27 [#4153 ci-fix-reinvoke] skipping dispatch — no-op already recorded for head 5d801fa5
21:28:59 [#4153 ci-gate] CI check(s) failed: ci-fix-sentinel (failure)
21:28:59 [#4153 ci-fix-reinvoke] skipping dispatch — no-op already recorded for head 5d801fa5
…every ~32s, unchanged head SHA…
```

### Why it is unreachable, not merely slow

The `#958 leg 2` no-op guard suppresses a CI-fix dispatch when the previous reinvoke for that exact head SHA pushed no new commit:

```go
// The last CI-fix reinvoke for this exact head SHA pushed no new commit —
// dispatching again would just repeat the same no-op and burn cycle budget
// for nothing. Wait for the SHA to advance (a genuine fix) or for
// CIWaitTimeout to fire (#958 leg 2).
```

In this scenario the head SHA **can never advance**: `ci-fix-sentinel` (fired via the `ci-fix-sentinel-unfixable` PR-body marker) is a fixture designed to fail unconditionally, so the agent has nothing it can commit on retry. The guard therefore latches permanently on the first attempt. Confirmed directly against `engine/catch_up_handlers.go`'s `dispatchWithCycleLimit`/`handleMergeAndCIGates`: `increment(repoStr)` (which applies `CIFixCycleIncremented`) only runs on the branch that also calls `dispatch()`, and the no-op short-circuit returns `true` *before* reaching that branch on every poll after the first no-op. Consequences, in order:

1. `CIFixCycleIncremented` fires at most once — `MaxCiFixCycles` (2 in the test bed) is never reached.
2. `pauseForCIFixCycleLimit` never executes.
3. The **only** remaining terminator is `pauseForCITimeout`.
4. #1320 pushes that terminator from 30 → 120 minutes, so the scenario now runs ~4× longer before any escalation, and the test's own wait expires first.

Note the guard's own comment names the escape hatch: *"or for CIWaitTimeout to fire"*. The design already assumes the timeout is what ends this case — which is precisely what #1320 disabled.

**The engine is behaving correctly throughout.** Detecting a failure, declining to re-dispatch a provably-useless reinvoke, and escalating on the wait timeout is all correct. The defect is entirely in the scenario's premise about which path it can exercise.

### Relationship to #916

The sibling `TestCIFixReinvoke` is already skipped on #916 — *"agent satisfies ci-fix-sentinel during Implement; first failure not yet deterministically forceable."* This is the same fixture-controllability problem from the other direction: #916 cannot force a first failure, and this issue cannot force a **second** attempt that produces a commit. Both stem from the sentinel fixture not being controllable enough to drive the CI-fix state machine deterministically. They likely share a fixture-design fix, but fixing #916 itself is not required by this issue (see Scope).

## Summary

`TestCIFixReinvokeCycleLimit`'s current fixture (`ci-fix-sentinel-unfixable`) gives the agent no file it can ever change, so the CI-fix no-op guard latches after the first reinvoke and `CIFixCycles` never advances past 1 — `pauseForCIFixCycleLimit` (the behavior the test claims to cover) can never fire, at any `FABRIK_CI_WAIT_TIMEOUT`. This issue supersedes #1320's timer-race diagnosis (which was based on a false premise) and either makes the cycle-limit path genuinely reachable by giving the fixture something committable each round, or converts the scenario to an honest `t.Skip` with the coverage gap recorded in the as-built docs. Either way, #1320's `FABRIK_CI_WAIT_TIMEOUT=120` bed prerequisite is revisited, since it currently only lengthens time-to-escalation for every `wait_for_ci` item on the bed with no compensating benefit.

## Requirements

- Either make the scenario able to reach `pauseForCIFixCycleLimit`, or stop claiming to test it:
  - **To reach it:** consecutive CI-fix attempts must each produce a **new head SHA** while the required check still fails — otherwise the no-op guard latches and no cycle budget is consumed. That needs a fixture where the agent can commit *something* each round while the required check keeps failing on a condition the agent cannot satisfy (e.g. a sentinel that fails unless an environment/workflow-controlled condition is met, but that still leaves an editable file for the agent to touch each round) — not one that admits no commit at all.
  - **If the cycle-limit path cannot be made reachable:** `t.Skip` it with a diagnostic naming this issue (mirroring #916's existing skip pattern), and note in `docs/state-machine.md` that `pauseForCIFixCycleLimit` has no e2e coverage.
- Revert or re-derive the `FABRIK_CI_WAIT_TIMEOUT=120` bed prerequisite introduced by #1320: as it stands it is actively harmful — it quadruples time-to-escalation for every `wait_for_ci` item on the bed in exchange for a cycle-limit path that still cannot be reached. `tests/e2e/README.md` prerequisite #8 must be updated to match whatever is decided.
- The scenario's pass criteria must verify the cycle-limit path **non-vacuously**: reaching a genuine engine-authored cycle-limit pause comment (`hasEngineCycleLimitComment` — the #1320 predicate, which correctly fixed the false-pass half of #1320 and should be kept), confirmed by neutralising `pauseForCIFixCycleLimit` locally and observing the scenario fail (a manual/documented verification step, not necessarily a new automated test).
- No change to engine behavior is implied by this issue — the no-op guard (`#958 leg 2`) and the CI-wait-timeout escalation (`pauseForCITimeout`) are both correct as-is; only the e2e scenario and its bed prerequisite are in scope.

## Scope

**In scope:**
- `tests/e2e/ci_fix_reinvoke_test.go` — `TestCIFixReinvokeCycleLimit`'s fixture design, and/or its skip behavior.
- The `FABRIK_CI_WAIT_TIMEOUT=120` bed prerequisite (`tests/e2e/README.md` prerequisite #8) introduced by #1320.
- `docs/state-machine.md` — recording the coverage gap if the scenario ends up skipped.

**Out of scope:**
- Any change to engine behavior (`engine/catch_up_handlers.go`'s no-op guard, `engine/ci.go`'s `pauseForCIFixCycleLimit`/`pauseForCITimeout`/`ciWaitTimeout`) — all confirmed correct as-is.
- Fixing `TestCIFixReinvoke`/#916 itself. That test remains skipped pending its own issue; this issue only notes the shared fixture-controllability root cause for context. If the fixture redesign done here also happens to unblock #916, that's a bonus, not a requirement.
- `minSafeCIWaitTimeoutMinutes` and its worst-case-per-cycle derivation are only relevant if the cycle-limit path remains timer-raced after this fix; if the fixture change makes the no-op guard no longer the blocker, re-derive or remove this machinery as needed.

## Open Questions

None. The choice between "make the fixture produce a new commit each round" and "skip with a documented coverage gap" is a feasibility question for the Research stage (can a required check be made to fail deterministically while still leaving something committable, without touching `.github/` workflow files, which are off-limits per the existing fixture's hard constraints) — not a product decision requiring user input.

## Prior Art / Context

- `engine/catch_up_handlers.go` — the `#958 leg 2` no-op guard inside `dispatchWithCycleLimit`, invoked from the CI-fix branch of `handleMergeAndCIGates`.
- `engine/ci.go` — `pauseForCIFixCycleLimit`, `pauseForCITimeout`, `ciWaitTimeout()`.
- `tests/e2e/ci_fix_reinvoke_test.go` — `minSafeCIWaitTimeoutMinutes`, the bed-config skip guard, `hasEngineCycleLimitComment` (keep — it correctly fixed the false-pass half of #1320), `ciFixCycleLimitBody` (the fixture needing redesign).
- `tests/e2e/README.md` — "Additional prerequisites for `TestCIFixReinvoke` and `TestCIFixReinvokeCycleLimit`" (prerequisite #8, to be revisited).
- `docs/state-machine.md` — CI-fix cycle-limit rows (`fabrik:paused`/`fabrik:awaiting-input` tables, §6.5.2-adjacent CI-fix reinvoke sequence) — update if the scenario ends up skipped.
- **#1320** — this issue supersedes its timer-race diagnosis (the premise that raising `FABRIK_CI_WAIT_TIMEOUT` alone would let the cycle-limit path win was false).
- **#916** — sibling fixture-controllability gap (`TestCIFixReinvoke` cannot force a *first* failure; this issue's fixture cannot force a *second* committable attempt).
- **#958 leg 2** — introduced the no-op guard this issue's fixture must be redesigned to work with, not around.

## Risks / Dependencies

- Any fixture redesign must not touch `.github/` workflow files (existing hard constraint in both `ciFixReinvokeBody` and `ciFixCycleLimitBody` — the CI workflow is immutable test infrastructure).
- Verifying the "non-vacuous" acceptance criterion (neutralising `pauseForCIFixCycleLimit` and observing failure) is a real e2e run against `handarbeit/fabrik-test-alpha` — costly (~$0.50–1.50, 30-90 min per the existing test's documented wall-clock) and should be done deliberately, not repeatedly.
- If the fixture is redesigned to be fixable-but-not-yet-fixed-per-round, care is needed that the agent's "best effort each round" behavior stays deterministic enough not to reintroduce the #1320 false-pass risk (agent prose quoting marker text) — `hasEngineCycleLimitComment`'s prefix-match guard should continue to be relied on, not reintroduced ad hoc.

arbeithand:
Closes #1323

## Summary

`TestCIFixReinvokeCycleLimit`'s fixture never guaranteed a new commit on each CI-fix reinvoke, so the `#958 leg 2` no-op guard (`engine/catch_up_handlers.go`) latched after the very first reinvoke. `CIFixCycleIncremented` never advanced, and `pauseForCIFixCycleLimit` — the very behavior the test claims to cover — was structurally unreachable at any `FABRIK_CI_WAIT_TIMEOUT`. This supersedes #1320's timer-race diagnosis, which was based on the false premise that raising the timeout alone would let the cycle-limit path win.

## What changed

- **`ciFixCycleLimitBody`** (`tests/e2e/ci_fix_reinvoke_test.go`): redesigned so the agent appends a line to a scratch file (`.ci-fix-attempts.log`) and commits+pushes **unconditionally** on every single CI-fix reinvoke, independent of any belief about fixability. The no-op guard only checks head-SHA equality, so this is enough to make it advance genuinely to `MaxCiFixCycles`.
- **`TestCIFixReinvokeCycleLimit`**: now captures a baseline `PRCommitCount` after Implement (mirroring `TestCIFixReinvoke`), asserts the workflow file wasn't touched, and — after the cycle-limit comment is confirmed — asserts the PR gained at least `maxCycles` commits beyond baseline. This makes "the cycle-limit path was genuinely exercised" independently verifiable, not just inferred.
- **Removed** the now-moot `minSafeCIWaitTimeoutMinutes` timer-coordination machinery (`observedWorstCaseMinutesPerCIFixCycle`, `cycleLimitTimerHeadroomMinutes`, the bed-config skip guard, `TestMinSafeCIWaitTimeoutMinutes`, and the now-unused `readEnvFileCIWaitTimeout` helper in `harness.go`) — its only data point (an 87-minute capture) is now understood to likely have been an artifact of this same bug.
- **`tests/e2e/README.md`**: prerequisite #8 reverted — no `FABRIK_CI_WAIT_TIMEOUT` override needed, the engine default (30 min) applies. Explains why #1320's premise no longer holds.
- Added a documented manual verification procedure (in `TestCIFixReinvokeCycleLimit`'s doc comment): neutralize `pauseForCIFixCycleLimit` locally and re-run against a live bed to confirm the assertions would actually fail — a deliberate, costly, one-off step, not run by this PR.

## Out of scope / not changed

- **No engine behavior changes.** `engine/catch_up_handlers.go`'s no-op guard and `engine/ci.go`'s `pauseForCIFixCycleLimit`/`pauseForCITimeout` are confirmed correct as-is and untouched.
- **No `docs/state-machine.md` change** — the engine's documented behavior is unchanged, so Validate shouldn't expect a doc update here (per Plan's explicit decision).
- `TestCIFixReinvoke`/#916 is out of scope and remains skipped.

## How to test

- `go build -tags e2e ./...` and `go vet -tags e2e ./tests/e2e/...` — clean.
- `go test -tags e2e -run 'TestHasEngineCycleLimitComment|TestHasEngineCIWaitTimeoutComment' ./tests/e2e/...` — passes (fast, no live bed needed).
- `go test ./...` — full suite passes except a pre-existing, unrelated flake (`cmd.TestExecute_ConfigYAMLApplied`, a SIGINT-handling test failing identically on `main` before this change).
- The actual live-bed verification of `TestCIFixReinvokeCycleLimit` (and the manual `pauseForCIFixCycleLimit`-neutralization check) requires a real e2e run against `handarbeit/fabrik-test-alpha` (~$0.50–1.50, 30–90 min) and is documented as a deliberate follow-up, not performed here.

### DIFF
--- a/tests/e2e/README.md
+++ b/tests/e2e/README.md
@@ -124,48 +124,42 @@ gh api rate_limit --jq '.resources.graphql'
    ```bash
    E2E_TIMEOUT=1h scripts/e2e/run.sh -run TestCIFixReinvoke
    ```
-8. **`FABRIK_CI_WAIT_TIMEOUT=120` (minutes) in the test bed `.env`** — required
-   for `TestCIFixReinvokeCycleLimit` to deterministically exercise the
-   cycle-limit pause path instead of racing the fixed CI-wait-timeout timer
-   (handarbeit/fabrik#1320). Both `pauseForCITimeout` (`engine/ci.go`) and
-   `pauseForCIFixCycleLimit` are evaluated independently every poll once
-   `fabrik:awaiting-ci` is applied; whichever condition is observed first
-   wins, and the two are otherwise uncoordinated. The engine default
-   (`FABRIK_CI_WAIT_TIMEOUT` unset → 30 min, see `ciWaitTimeout()`) is too
-   short: a captured release-gate run showed two failed CI-fix cycles
-   (`FABRIK_MAX_CI_FIX_CYCLES=2`) still executing at 1h26m26s (~87 min) under
-   full-suite load — comfortably past the 30-minute default, so the
-   wait-timeout timer fired first and the scenario observed the wrong (but
-   still engine-correct) pause reason. **This is the same 87-minute data
-   point cited in "How the timeout/parallelism defaults are derived" below.**
+8. **`FABRIK_CI_WAIT_TIMEOUT` — no override needed; the engine default (30
+   min, see `ciWaitTimeout()`) applies.** #1320 had introduced a
+   `FABRIK_CI_WAIT_TIMEOUT=120` requirement here, reasoning that
+   `TestCIFixReinvokeCycleLimit` needed extra headroom for the cycle-limit
+   path (`pauseForCIFixCycleLimit`) to win its race against the fixed
+   CI-wait-timeout timer (`pauseForCITimeout`). **handarbeit/fabrik#1323
+   found that premise false**: the scenario's fixture never guaranteed a new
+   commit on each CI-fix reinvoke, so the `#958 leg 2` no-op guard
+   (`engine/catch_up_handlers.go`) latched permanently after the very first
+   reinvoke — `CIFixCycleIncremented` never advanced past 1, and
+   `pauseForCIFixCycleLimit` was structurally unreachable *at any*
+   `FABRIK_CI_WAIT_TIMEOUT`. Raising the timeout to 120 min therefore didn't
+   fix a race; it just quadrupled time-to-escalation for every `wait_for_ci`
+   item in the bed, with no compensating benefit.
 
-   The test derives its own minimum safe value at runtime
-   (`minSafeCIWaitTimeoutMinutes` in `ci_fix_reinvoke_test.go`):
-   `observedWorstCaseMinutesPerCIFixCycle (45, i.e. ceil(87/2)) * maxCycles +
-   cycleLimitTimerHeadroomMinutes (30)`. For the recommended
-   `FABRIK_MAX_CI_FIX_CYCLES=2` (prerequisite #6), that's `45*2 + 30 = 120`
-   minutes — hence the **120** recommendation. If `FABRIK_CI_WAIT_TIMEOUT` in
-   the bed `.env` is below this computed minimum for the bed's configured
-   `FABRIK_MAX_CI_FIX_CYCLES`, the test `t.Skip`s with an instructional
-   diagnostic rather than racing the two timers and risking a false
-   pass/fail — see the "Requirements" section of #1320 for why "accept
-   either pause reason" was rejected (the #1319 anti-pattern: a test that
-   passes regardless of which path actually ran).
+   The fixture (`ciFixCycleLimitBody`) now forces an unconditional
+   scratch-file commit+push on every single CI-fix reinvoke, independent of
+   the agent's belief about fixability — the no-op guard only checks head-SHA
+   equality, so this is sufficient to make `CIFixCycleIncremented` genuinely
+   advance to `MaxCiFixCycles` through real repeated dispatch. With the path
+   actually reachable, whether the CI-wait-timeout timer also fires is
+   ordinary e2e timing risk like any other scenario in the bed — not a
+   known-guaranteed structural loss requiring bespoke pre-flight timer math.
+   The test now also asserts the PR gained at least `maxCycles` commits
+   beyond baseline (see `TestCIFixReinvokeCycleLimit`'s doc comment), making
+   "the cycle-limit path was genuinely exercised" independently verifiable
+   rather than inferred from timer coordination.
 
-   Raising this bed-wide from 30→120 min is a deliberate tradeoff: it also
-   lengthens how long a genuinely CI-stuck *unrelated* issue sits before a
-   human is notified via `pauseForCITimeout`, on any issue in the bed, not
-   just this scenario's. After editing `.env`, restart the test-bed Fabrik
-   instance so the new value takes effect.
-
-   **`TestCIFixReinvoke` has the same timer-coupling risk, dormant.** It is
-   currently `t.Skip`'d pending #916 (see the doc comment above `TestCIFixReinvoke`),
-   so this hasn't yet caused a failure, but a single successful CI-fix cycle
-   could plausibly also exceed a too-low `FABRIK_CI_WAIT_TIMEOUT` under load
-   once un-skipped. Re-check this when #916 unblocks it — the derivation
-   formula above only accounts for `TestCIFixReinvokeCycleLimit`'s
-   `MaxCiFixCycles`-bounded cost, not `TestCIFixReinvoke`'s single-cycle
-   recovery cost, which may need its own (likely smaller) minimum.
+   The 87-minute data point that backed the old `120` recommendation (a
+   captured release-gate run "still executing" with `FABRIK_MAX_CI_FIX_CYCLES=2`)
+   is now understood to have likely been an artifact of the same no-op-latch
+   bug — a cycle count stuck at 1 while wall-clock kept advancing is
+   indistinguishable from this exact defect. It should not be treated as a
+   trustworthy per-cycle cost measurement; a future live run against the
+   fixed fixture is what should supply real numbers if the wait-timeout timer
+   does turn out to race in practice.
 
 ### Marker-substring assertion audit (#1320)
 
--- a/tests/e2e/ci_fix_reinvoke_marker_test.go
+++ b/tests/e2e/ci_fix_reinvoke_marker_test.go
@@ -116,24 +116,3 @@ func TestHasEngineCIWaitTimeoutComment(t *testing.T) {
 		})
 	}
 }
-
-// TestMinSafeCIWaitTimeoutMinutes locks in the derivation formula so a future
-// edit to the constants (or the formula itself) is a visible, deliberate
-// change rather than a silent drift back toward the timer race.
-func TestMinSafeCIWaitTimeoutMinutes(t *testing.T) {
-	tests := []struct {
-		maxCycles int
-		want      int
-	}{
-		{maxCycles: 2, want: 120}, // 45*2 + 30 — the documented bed recommendation
-		{maxCycles: 3, want: 165},
-		{maxCycles: 1, want: 75},
-	}
-
-	for _, tt := range tests {
-		got := minSafeCIWaitTimeoutMinutes(tt.maxCycles)
-		if got != tt.want {
-			t.Fatalf("minSafeCIWaitTimeoutMinutes(%d) = %d, want %d", tt.maxCycles, got, tt.want)
-		}
-	}
-}
--- a/tests/e2e/ci_fix_reinvoke_test.go
+++ b/tests/e2e/ci_fix_reinvoke_test.go
@@ -191,36 +191,23 @@ func hasEngineCIWaitTimeoutComment(bodies []string) bool {
 	return false
 }
 
-// observedWorstCaseMinutesPerCIFixCycle and cycleLimitTimerHeadroomMinutes
-// back minSafeCIWaitTimeoutMinutes, which computes the smallest
-// FABRIK_CI_WAIT_TIMEOUT (minutes) at which TestCIFixReinvokeCycleLimit can
-// deterministically exercise the cycle-limit pause path instead of racing
-// the fixed CI-wait-timeout timer (handarbeit/fabrik#1320).
-//
-// observedWorstCaseMinutesPerCIFixCycle=45 is derived from the one real data
-// point available: a captured release-gate run where this scenario was
-// still executing at 1h26m26s (~87min) with FABRIK_MAX_CI_FIX_CYCLES=2 — see
-// "How the timeout/parallelism defaults are derived" in tests/e2e/README.md.
-// ceil(87/2) = 44, rounded up to 45 for a whole-minutes-per-cycle budget.
-const observedWorstCaseMinutesPerCIFixCycle = 45
-
-// cycleLimitTimerHeadroomMinutes is added on top of the raw per-cycle
-// worst-case cost as margin against load variance beyond the single
-// observed data point backing observedWorstCaseMinutesPerCIFixCycle.
-const cycleLimitTimerHeadroomMinutes = 30
-
-// minSafeCIWaitTimeoutMinutes returns the minimum FABRIK_CI_WAIT_TIMEOUT
-// (minutes) needed for the cycle-limit path to deterministically win the
-// race against the fixed CI-wait-timeout timer, for a given
-// FABRIK_MAX_CI_FIX_CYCLES.
-func minSafeCIWaitTimeoutMinutes(maxCycles int) int {
-	return observedWorstCaseMinutesPerCIFixCycle*maxCycles + cycleLimitTimerHeadroomMinutes
-}
-
 // TestCIFixReinvokeCycleLimit is the negative-path regression test for the
 // CI-fix reinvoke cycle limit (engine/ci.go: pauseForCIFixCycleLimit).
 // The sentinel CI job is permanently failing for this issue — Claude cannot fix
 // it — so the engine must exhaust MaxCiFixCycles and pause the issue.
+//
+// handarbeit/fabrik#1323 superseded #1320's timer-race diagnosis: the
+// previous fixture never guaranteed a new commit per CI-fix reinvoke, so the
+// `#958 leg 2` no-op guard (engine/catch_up_handlers.go) latched after the
+// first reinvoke and CIFixCycleIncremented never advanced — the cycle-limit
+// path was structurally unreachable, at any FABRIK_CI_WAIT_TIMEOUT. The
+// fixture below (ciFixCycleLimitBody) instead forces an unconditional
+// scratch-file commit+push on every single reinvoke, independent of any
+// belief about fixability, which is the only thing the no-op guard actually
+// checks (head SHA equality). With that fixed, the cycle-limit path is
+// reached through genuine repeated dispatch and no bespoke timer
+// coordination with FABRIK_CI_WAIT_TIMEOUT is needed — the engine default
+// applies.
 //
 // Pass criteria:
 //   - fabrik:awaiting-ci appears (CI gate fires).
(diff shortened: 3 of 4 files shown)


arbeithand:
*branch: fabrik/issue-1323 | commit: 3945a68c | main: 115374c06529e70cfd481106ed8f2095ab6f2d6d | 2026-08-01 21:52 UTC*

Everything is committed and pushed. Now I'll signal PR creation.


Redesigned `TestCIFixReinvokeCycleLimit`'s fixture to force an unconditional per-round commit+push, making the cycle-limit path (`pauseForCIFixCycleLimit`) genuinely reachable instead of latching after one reinvoke as diagnosed in the issue. Added a non-vacuous PR-commit-count assertion, removed the now-moot `FABRIK_CI_WAIT_TIMEOUT` timer-coordination machinery from #1320, and reverted the README bed prerequisite to the engine default. No engine behavior changed; build/vet/tests pass (one pre-existing unrelated flake confirmed on `main`).

---
Used 40/100 turns, 0k input / 24k output tokens.

arbeithand:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-1323 | commit: f898fb30 | main: 115374c06529e70cfd481106ed8f2095ab6f2d6d | 2026-08-01 21:58 UTC*

## Review Findings

### Fixed
- **Issue**: `AssertPRDidNotTouchWorkflows` was only called against the pre-Validate baseline commit, never against the commits the CI-fix cycle-limit reinvokes actually produce — which is precisely the change set this fixture redesign now makes real (previously the no-op guard latched after one reinvoke, so this gap was largely theoretical; now genuine repeated dispatch is the whole point). Flagged by Pruefer's automated review on PR #1324. **Fix**: added a second `AssertPRDidNotTouchWorkflows` call after the cycle-limit loop completes and the non-vacuous commit-count check passes, so a rogue `.github/` edit during any of the `maxCycles` reinvokes now fails the test loudly.

### Verified
- Rebased cleanly onto `origin/main` (no local `main` drift; `origin/main..HEAD` scoped to exactly this issue's changes plus the review fix).
- `go build ./...`, `go vet ./...`, `go build -tags e2e ./tests/e2e/...`, `go vet -tags e2e ./tests/e2e/...` — all clean.
- Fast e2e unit tests (`TestHasEngineCycleLimitComment`, `TestHasEngineCIWaitTimeoutComment`) pass.
- `go test -race -timeout 5m ./...` — 3405 passed, 3 failed (all in `cmd`/`engine`, untouched by this diff — pre-existing/flaky, consistent with what Implement already noted).
- No dangling references anywhere in the repo to the removed `minSafeCIWaitTimeoutMinutes`/`observedWorstCaseMinutesPerCIFixCycle`/`cycleLimitTimerHeadroomMinutes`/`readEnvFileCIWaitTimeout`/`TestMinSafeCIWaitTimeoutMinutes` machinery.
- `TestCIFixReinvokeCycleLimit`'s cycle-arithmetic (`finalCommits >= baseCommits+maxCycles`) matches `dispatchWithCycleLimit`'s dispatch-then-pause-at-limit semantics — cross-checked against `engine/catch_up_handlers.go`.
- README.md's revised prerequisite #8 and doc-comment updates have no stale cross-references; the untouched "How the timeout/parallelism defaults are derived" section (E2E_TIMEOUT=4h) rests on the *fact* the scenario ran past 90 minutes under load, which holds regardless of the no-op-latch diagnosis, so no inconsistency there.
- Confirmed no engine-behavior changes (`engine/catch_up_handlers.go`, `engine/ci.go` untouched) and no `docs/state-machine.md`/`docs/llms-full.txt` update needed (`tests/e2e/README.md` isn't a canonical doc page).

PR is ready for human review.

Reviewed the CI-fix cycle-limit fixture redesign; build/vet/tests all pass and the reasoning checks out against the engine code. Fixed one issue flagged by Pruefer's review: the workflow-immutability guard only checked the pre-Validate baseline commit, not the commits the cycle-limit reinvokes themselves produce — added a second check after the loop completes. PR is ready for human review.

---
Used 39/100 turns, 0k input / 11k output tokens.

arbeithand:
*branch: fabrik/issue-1323 | commit: 5c3d0e7f | main: 115374c06529e70cfd481106ed8f2095ab6f2d6d | 2026-08-01 21:59 UTC*

Both review threads resolved. The issue body needs no changes — this was a review-comment processing pass on the PR, not a spec revision, so I'll omit the `FABRIK_ISSUE_UPDATE` markers.

Summary: addressed both Pruefer review comments on PR #1324 — thread 1 (workflow-guard re-check after cycle-limit commits) was already fixed in the prior Review-stage commit; thread 2 (documenting the engine-regression-vs-agent-compliance ambiguity) is now fixed by adding an explicit note to `TestCIFixReinvokeCycleLimit`'s doc comment. Build/vet pass, both changes committed and pushed, and both review threads resolved on GitHub.

---
**Threads addressed:**
- `tests/e2e/ci_fix_reinvoke_test.go:268` — resolved
- `tests/e2e/ci_fix_reinvoke_test.go:446` — resolved

arbeithand:
*branch: fabrik/issue-1323 | commit: 4ac942c1 | main: 115374c06529e70cfd481106ed8f2095ab6f2d6d | 2026-08-01 22:01 UTC*

Thread resolved. The issue body needs no changes here — omitting the `FABRIK_ISSUE_UPDATE` markers.

Summary: fixed the finding from Pruefer's review — the doc comment previously implied an agent-skipped-push failure would surface as the commit-count assertion, but it actually latches the no-op guard and causes an earlier timeout waiting for `fabrik:paused`. Corrected the wording, verified build/vet, committed, pushed, and resolved the review thread.

---
**Threads addressed:**
- `tests/e2e/ci_fix_reinvoke_test.go:327` — resolved

arbeithand:
*branch: fabrik/issue-1323 | commit: 4ac942c1 | main: 115374c06529e70cfd481106ed8f2095ab6f2d6d | 2026-08-01 22:06 UTC*

Final gate confirms MERGEABLE/CLEAN, up to date with main. All checks pass — ready to signal completion.

## Validation Report

### Requirements: 4/4 passed
- [x] Cycle-limit path made genuinely reachable: `ciFixCycleLimitBody` now forces an unconditional scratch-file commit+push on every CI-fix reinvoke, making `headAfter != headBefore` deterministic instead of relying on agent judgment — verified in diff.
- [x] `FABRIK_CI_WAIT_TIMEOUT=120` prerequisite reverted to engine default (30 min), with README explaining #1320's premise was false.
- [x] Non-vacuous verification: `hasEngineCycleLimitComment` kept unchanged (as required), plus a new automated `PRCommitCount >= baseCommits + maxCycles` check, plus a documented (not executed, per the issue's explicit allowance) manual neutralize-and-observe procedure.
- [x] No engine behavior changed — diff confirmed scoped entirely to `tests/e2e/{ci_fix_reinvoke_test.go, ci_fix_reinvoke_marker_test.go, harness.go, README.md}`.

### Test Suite: PASSED
- `go build -tags e2e ./...`, `go vet -tags e2e ./tests/e2e/...` — clean
- `go test -race -timeout 5m ./...` — 3405 passed, 3 failed, 1 skipped. All 3 failures reproduce identically in isolation and are pre-existing/environmental (`TestInvokeClaude_FabrikEnvVarsInjected` — local `NVM_INC` env leak; `TestExecute_ConfigYAMLApplied` — SIGINT-handling flake), both in packages (`engine`, `cmd`) this PR never touches.
- `go test -tags e2e -run 'TestHasEngineCycleLimitComment|TestHasEngineCIWaitTimeoutComment'` — 11 passed.

### Branch state
- Rebased onto `origin/main` (up to date, no conflicts).
- `gh pr view`: `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`.
- Plan's full task checklist (11 items) verified complete against the actual diff — dead `minSafeCIWaitTimeoutMinutes` machinery removed cleanly (`TestMinSafeCIWaitTimeoutMinutes`, `readEnvFileCIWaitTimeout`, no dangling references anywhere in the repo).

### Verdict: READY TO MERGE

Validation passed. PR mergeable: MERGEABLE, mergeStateStatus: CLEAN. All 4 requirements verified against the diff (fixture redesign, prerequisite revert, non-vacuous checks kept/added, no engine changes); build/vet/e2e-build clean; full race test suite passes except 3 pre-existing environmental failures in untouched packages (`engine`, `cmd`), confirmed unrelated by diff scope and reproducibility in isolation.

---
Used 25/100 turns, 0k input / 7k output tokens.

arbeithand:
*branch: fabrik/issue-1323 | commit: 28106061 | main: 115374c06529e70cfd481106ed8f2095ab6f2d6d | 2026-08-01 22:08 UTC*

Thread resolved. The issue body needs no changes — this was a review-comment processing pass, not a spec revision, so I'm omitting the `FABRIK_ISSUE_UPDATE` markers.

**Summary:** Addressed Pruefer's finding on PR #1324 about the non-vacuous check's fragility (PR commit count is one layer removed from the engine's own `CIFixCycleIncremented` decision). Investigated whether an engine-observable signal was feasible — confirmed the e2e harness has no access to the live bed's engine logs/internal state, and adding that would be new engine-side observability, which is explicitly out of scope for this issue. Documented this reasoning in the doc comment, verified build/vet still pass, committed, pushed, and resolved the review thread.

---
**Threads addressed:**
- `tests/e2e/ci_fix_reinvoke_test.go:343` — resolved

---
**Threads addressed:**
- `tests/e2e/ci_fix_reinvoke_test.go:343` — resolved

arbeithand:
## Review — approving

This correctly rejects #1320's diagnosis rather than iterating on it, which was the hard part. Checked against the four things I most wanted to see:

**1. Attacks the real mechanism, not the timing.** The fixture now forces an unconditional scratch-file commit+push on *every* reinvoke — targeting the only thing the `#958 leg 2` guard actually tests (head-SHA equality), rather than assuming the agent will produce a commit as a side effect of trying to fix an unfixable check. That is the correct causal fix: the previous fixture could never advance the SHA, so the guard latched on attempt one and `CIFixCycleIncremented` never passed 1.

**2. Non-vacuity is built in, and the residual gap is stated honestly.** The `PRCommitCount` baseline → "at least maxCycles commits beyond baseline" assertion is a genuine in-test guard: it fails if the cycle-limit pause is reached by anything other than repeated dispatch, which is exactly the failure mode that produced #4049's false pass. I also appreciate that the doc comment says plainly that neutralizing `pauseForCIFixCycleLimit` to prove non-vacuity requires a costly live-bed run and was **not** performed, rather than implying it was.

**3. The harmful prerequisite is retracted, with reasoning.** `minSafeCIWaitTimeoutMinutes`, `observedWorstCaseMinutesPerCIFixCycle`, `cycleLimitTimerHeadroomMinutes` and `readEnvFileCIWaitTimeout` are all gone, and README prerequisite #8 now explains *why* the 120-minute requirement was wrong. That matters more than the deletion: a bare removal would invite someone to reintroduce it next time the scenario goes red. Confirmed empirically on the bed — with 120 set, the item ran 103 minutes with no escalation at all, because the only working terminator had been pushed out of reach.

**4. `hasEngineCycleLimitComment` is preserved** (7 call sites intact). That was the genuinely correct half of #1320 — the body-prefix predicate that stops agent prose satisfying the assertion — and it would have been easy to lose while reverting the rest.

### Residual fragility — well documented, no action needed

The comment at ~L241 is honest that nothing engine-side enforces that a given reinvoke actually pushed: if the agent misjudges the unfixable check as fixable and attempts a "real" fix without touching the scratch file, `LastCIFixNoOpSHA` latches again and the scenario regresses to the #1323 shape. That is a real dependency on agent behavior, correctly surfaced rather than hidden. If this scenario goes intermittently red again, that is the first thing to check.

### Verification

CI 5/5 green, mergeable. Bed already matches the new prerequisites — I removed the `FABRIK_CI_WAIT_TIMEOUT=120` override after the failed validation slice, so it is back on the engine default.

No blocking findings.

PR #1369, opened 2026-08-03
perf(engine): eliminate duplicate FetchLabelAppliedAt calls via record-at-write cache

### LINKED ISSUE #1314: perf(engine): CIWaitTimeout backstop duplicates checkCIGate's FetchLabelAppliedAt call
## Summary

The `CIWaitTimeout` backstop added in #1303 / PR #1309 makes an unconditional `FetchLabelAppliedAt(…, "fabrik:awaiting-ci")` call for every `fabrik:awaiting-ci` item on every poll, inside `settleAwaitingCIScan` (`engine/ci_settle.go`). For items that go on to reach `checkCIGate` normally — the common case — `checkCIGate` (via `classifyCIFromCheckRuns` and sibling call sites in `engine/ci.go`) then calls `FetchLabelAppliedAt` again for the same label and item moments later in the same iteration.

Steering from issue discussion (2026-08-02, 2026-08-03) established facts that change the right fix:

1. **The duplicated call is not cheap.** `FetchLabelAppliedAt` (`github/labels.go:151-183`) pages through the *entire* issue-events history on every invocation — the events API returns oldest-first, the function is searching for the *most recent* application (`for page := 1; ; page++`, terminating only on an empty page, keeping the max via `t.After(applied)`), and there is no early exit. Cost is `ceil(events/100) + 1` REST requests per call, confirmed against real completed issues in this repo (#1338: 50 events, ~2 calls; #1301: 52 events, ~2 calls; #1303: 65 events, ~2 calls). The duplication described in this issue therefore costs roughly 2 wasted REST calls per awaiting-ci item per poll, growing with issue age — worst on the items that sit in `fabrik:awaiting-ci` longest.
2. **The result is never cached.** `FetchLabelAppliedAt` does not appear in `boardcache` (not in `ReadClient`, not in `CacheImpl`). Every call, from every call site, is a live paginated sweep.
3. **The actual call-site count is 10, not ~23 (corrected 2026-08-03).** An earlier pass over-counted by including comments, the interface declaration, and doc mentions rather than actual invocations. Verified invocations of `FetchLabelAppliedAt(`:

   | file | calls | label(s) |
   |---|---|---|
   | `engine/ci.go` | 4 | `fabrik:awaiting-ci` (all four) |
   | `engine/reviews.go` | 2 | `fabrik:bot-reprompted`, `fabrik:awaiting-review` |
   | `engine/merge_gate.go` | 2 | `fabrik:auto-merge-enabled` (both) |
   | `engine/ci_settle.go` | 1 | `fabrik:awaiting-ci` |
   | `engine/archive_done_settle.go` | 1 | `stage:<name>:complete` |
   | **total** | **10** | |

   `engine/archive_done_settle.go` itself already works around this exact cost with its own mitigation (a `CooldownAt`-cached "archive eligible at" timestamp plus a `maxArchiveLabelFetchesPerPoll` budget, ADR-068) — useful prior art for a caching-based fix. All five distinct labels queried across these 10 calls are applied only by the engine itself (confirmed) — no external actor labels these, which matters for the record-at-write option below.

With the corrected count, **the choice between fixing the shared primitive and the narrower `phase1Ctx`-threading fix is not pre-decided** — Research must weigh both honestly against the real 10-call, 5-file footprint (four of which are the same label in the same file) rather than assuming the primitive fix wins on scale alone.

## Requirements

- Research must evaluate, and Plan must select among, the following options — not treat any one as pre-decided:
  1. **Reverse pagination** of `FetchLabelAppliedAt`: request the last page first (e.g. `per_page=100` plus the `Link: rel="last"` response header, or `direction=desc` if the issue-events endpoint honors it — this must be verified against the actual live API, not assumed from docs) and stop at the first matching event.
  2. **Record-at-write**: since the engine is confirmed to be the sole applier of every label queried via `FetchLabelAppliedAt` in this codebase, record each label's applied-at timestamp at the moment the engine itself applies it, rather than fetching or invalidating a TTL cache reactively. This is exact (not approximate) and needs no TTL. Gap: an engine restart loses the in-memory record, requiring fallback to the existing fetch path for that one case. **Caveat to design for explicitly:** any in-process record/cache is per-instance — if two engine instances ever manage the same repo, one instance's record does not observe the other's label mutation. State this assumption explicitly in the design rather than leaving it implicit, whether or not it is judged acceptable.
  3. **Poll-native via GraphQL**: the board query already fetches `labels(first: 30)` (names only, `github/project.go:268`); adding `timelineItems(itemTypes: LABELED_EVENT)` (or equivalent) could supply the applied-at timestamp with no additional REST request, consistent with the pattern used elsewhere for poll-native signals (e.g. `LinkedPRIsMergeQueueEnabled`, per `docs/state-machine.md`). **This option must be costed against the GraphQL budget specifically, not chosen on architectural-consistency grounds alone** — REST and GraphQL are separate pools, and the risk context below (GraphQL exhausted on 2026-08-01 while REST still had ~2100 remaining) means moving cost from REST to GraphQL is not a strict win; per-item `timelineItems` GraphQL node cost must be estimated and weighed against the tighter budget before this is selected.
  4. The originally-scoped narrower fix: threading the backstop's already-fetched `FetchLabelAppliedAt` result through `phase1Ctx` to `checkCIGate`, limited to the `fabrik:awaiting-ci` / `settleAwaitingCIScan` path only.
  - Options may be combined (e.g. record-at-write with reverse-pagination fallback on a cold cache).
- Whichever option(s) are selected, the decision and rationale (including why rejected options were rejected, weighed against the real 10-call footprint and the separate-budget consideration for option 3) must be recorded in Plan output.
- If the narrower option (4) is chosen alone, it must be noted explicitly that the other 9 call sites remain unaddressed.
- **Correctness constraints that must survive whichever route is taken:**
  - The fail-open contract is load-bearing: a fetch error or zero timestamp must leave every consumer inert (no escalation on incomplete data), exactly as today.
  - Correctness is "most recent application," not "any application" — a label applied, removed, and re-applied must still yield the latest timestamp. Any early-exit, reverse-pagination, or record-at-write scheme must be tested explicitly against that sequence, since it is the case a naive first-match-wins or record-once optimization gets wrong.
  - A reverse-pagination approach must fail open on a missing or malformed `Link` header (or equivalent pagination signal) — not fall through to "no application found," which would silently suppress a timeout that should have fired.
  - `TestSettleAwaitingCIScan_CIWaitTimeoutBackstop_PausesRegardlessOfGateClaim` and `..._NoOpWithinTimeout` must continue to pass unchanged.
  - Any new test asserting reduced call count must be proven non-vacuous: a call-count assertion alone passes trivially against code that never calls the function; pair it with a test that fails if the call is removed entirely (i.e. still exercises the underlying behavior it gates).

## Scope

**In scope:**
- Evaluating the four options above (reverse pagination, record-at-write, poll-native GraphQL, narrower `phase1Ctx` threading) against the corrected 10-call, 5-file footprint, and implementing whichever is selected (options may combine).

**Out of scope:**
- Changing `CIWaitTimeout` (or any other label-applied-timeout) semantics, values, or escalation/pause behavior — only the cost of computing "when was this label applied."
- Rewriting or auditing the surrounding logic at each of the 10 call sites individually; the goal is to fix the shared primitive (or, if that route isn't chosen, the one backstop/gate pair) so calls drop, not to touch each call site's unrelated behavior.

## Prior Art / Context

- `github/labels.go:151-183` `FetchLabelAppliedAt` — the primitive; oldest-first pagination with no early exit, no caching.
- `engine/archive_done_settle.go` — an existing, working mitigation pattern for this exact cost (`CooldownAt`-cached timestamp + `maxArchiveLabelFetchesPerPoll` budget, ADR-068); a useful precedent for the record-at-write/cache option.
- `engine/ci_settle.go:206` — the backstop's `FetchLabelAppliedAt` call, inside `settleAwaitingCIScan`.
- `engine/ci.go:128,186,219,234` — `classifyCIFromCheckRuns` and sibling call sites, all four querying `fabrik:awaiting-ci`.
- `engine/reviews.go:867,953`, `engine/merge_gate.go:351,395` — the remaining call sites.
- `engine/catch_up_handlers.go` — `phase1Ctx` struct; the mechanism for option 4 (narrower fallback fix).
- `github/project.go:268` — the board GraphQL query's existing `labels(first: 30)` field (names only); the attachment point for option 3.
- `docs/state-machine.md` — documents the "poll-native, never a REST fetch" pattern already used for other signals (e.g. `LinkedPRIsMergeQueueEnabled`), the precedent option 3 would follow.
- PR #1309 review thread (the originating finding); ADR-1270 (the settle-scan pattern the backstop is part of); ADR-068 (the archive-settle mitigation pattern).

## Risks / Dependencies

- Reverse pagination depends on GitHub's issue-events API actually supporting a reliable "give me the last page" mechanism (`Link: rel="last"` or `direction=desc`) — this must be verified against the live API during Research, not assumed from documentation alone.
- Record-at-write assumes a single engine instance per repo; if that assumption doesn't hold in some deployment, the design must say so explicitly rather than silently relying on it.
- Poll-native GraphQL must be costed against GraphQL node-budget impact specifically before being selected — REST and GraphQL are separate, independently exhaustible pools.
- Whichever approach is chosen must be validated against the "applied, removed, re-applied" sequence to avoid the naive first-match-wins/record-once bug described above.
- The 2026-08-01 release-gate run exhausted the GraphQL budget and pushed REST to ~2100 remaining; separately, on 2026-08-02 the shared `verveguy` budget drained to ~700/5000 within a single window. This raises urgency but does not change correctness requirements — the fail-open and most-recent-application constraints are non-negotiable regardless of approach chosen.

arbeithand:
Closes #1314

## Summary

`FetchLabelAppliedAt` (`github/labels.go`) pages through an issue's entire events history with no early exit — up to `ceil(events/100)+1` REST requests per call — and isn't cached anywhere. `settleAwaitingCIScan`'s `CIWaitTimeout` backstop and `checkCIGate` both called it for the same item/label in the same poll pass; the same pattern (uncached, per-poll re-fetch) recurs across 10 call sites in 5 files for 5 distinct labels.

This PR fixes the shared primitive rather than the one duplication: since the engine is the *sole* applier of `fabrik:awaiting-ci`, `fabrik:awaiting-review`, `fabrik:bot-reprompted`, and `fabrik:auto-merge-enabled`, it now records each label's applied-at timestamp in memory at the moment it writes the label (record-at-write), instead of re-fetching it from GitHub on every read.

`stage:<name>:complete` (the 5th queried label) is intentionally out of scope — `archive_done_settle.go` already has an equivalent, working mitigation for it (ADR-068).

## Key changes

- `internal/itemstate`: new `ItemState.LabelAppliedAt map[string]time.Time`, deliberately separate from `CooldownAt` (aliasing them would make `HasExpiredCooldown` misfire permanently — see the ADR), plus its mutation, change flag, and snapshot getter.
- `engine/mutate.go`: `recordLabelAppliedAtNow` (writes the cache, wired into `applyLabelAdd`'s success path) and `labelAppliedAt` (the shared read path — cache hit is free; cache miss falls back to the existing `FetchLabelAppliedAt` and self-heals the cache).
- All 9 read call sites across `engine/ci.go`, `engine/ci_settle.go`, `engine/reviews.go`, `engine/merge_gate.go` now go through `labelAppliedAt` — same `(time.Time, error)` signature and fail-open contract, no caller-side logic changed.
- `engine/stages.go`: 5 direct `AddLabelToIssue` sites that bypass `applyLabelAdd` now also call `recordLabelAppliedAtNow`.
- Reverse pagination and poll-native GraphQL were evaluated and explicitly rejected (see `adrs/1314-label-applied-at-record-on-write.md` for the full rationale, including a two-engine-instance caveat for the in-process cache).

## Test plan

- [x] New unit tests in `internal/itemstate` for `LabelAppliedAtRecorded` overwrite semantics and non-aliasing with `CooldownAt`
- [x] Existing `TestSettleAwaitingCIScan_CIWaitTimeoutBackstop_PausesRegardlessOfGateClaim` / `..._NoOpWithinTimeout` pass unchanged
- [x] New `engine/label_applied_at_test.go`: single-pass dedup (backstop + gate share one live fetch), cross-poll caching, cold-start fallback self-heal, and applied→removed→re-applied reads the latest timestamp (through the real guarded write path)
- [x] `go vet ./...`, `go build .`, `go test -race ./...` all pass (pre-existing, unrelated SIGINT-timing flakiness in `cmd` package confirmed identical on `main`)


arbeithand:
*branch: fabrik/issue-1314 | commit: c7fe941f | main: 4b2150ad30403215089a03ca5942999a6cf9a27c | 2026-08-03 01:52 UTC*

Implemented record-at-write: the engine now caches each label's applied-at timestamp in memory the moment it writes the label (new `itemstate.LabelAppliedAt` map, kept separate from `CooldownAt` to avoid misfiring `HasExpiredCooldown`), and all 9 relevant `FetchLabelAppliedAt` read call sites route through a shared `labelAppliedAt` helper that self-heals on a cold cache. This eliminates the CIWaitTimeout-backstop/checkCIGate duplication and the same per-poll REST cost at the other 8 sites. Added itemstate + engine regression tests (dedup, cross-poll caching, cold-start fallback, most-recent-application correctness), wrote ADR-1314 recording why reverse pagination and poll-native GraphQL were rejected, updated the Plan checklist, and pushed all commits.

---
Used 141/100 turns, 23.5M input (195 raw + 23.1M cache-read + 392k cache-write) / 60k output tokens.

arbeithand:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-1314 | commit: c7fe941f | main: 4b2150ad30403215089a03ca5942999a6cf9a27c | 2026-08-03 01:59 UTC*

Already fully pushed and in sync with origin, nothing to commit. The review found no issues requiring fixes — the implementation is correct and complete.

## Review Findings

### Verified

- **Rebase**: branch is already up to date with `origin/main` (via merge commit 4b2150ad); no conflicts.
- **Correctness — the core duplication fix**: `settleAwaitingCIScan`'s `CIWaitTimeout` backstop and `classifyCIFromCheckRuns`/`classifyCIFromRequiredContexts`/both `classifyCIFromMergeableState` branches all route through the new `e.labelAppliedAt` helper, which checks the in-memory `ItemState.LabelAppliedAt` cache before falling back to the live `FetchLabelAppliedAt`. Confirmed via a live test (`TestSettleAwaitingCIScan_LabelAppliedAt_DedupesWithinSinglePass`) that exactly one live REST call happens per poll pass for `fabrik:awaiting-ci`, down from two.
- **All 9 non-`archive_done_settle.go` read call sites** across `engine/ci.go`, `engine/ci_settle.go`, `engine/reviews.go`, `engine/merge_gate.go` are migrated — confirmed by grepping remaining `FetchLabelAppliedAt` occurrences; only `archive_done_settle.go` (deliberately out of scope, already has its own ADR-068 mitigation) and `mutate.go`'s fallback path remain.
- **Write-site guard audit**: traced every one of the 10 label-add write sites for the 4 target labels (`ci.go`, `reviews.go`, `stages.go` ×5) and confirmed each is preceded by a genuine "not already present" check, so `recordLabelAppliedAtNow`'s unconditional `time.Now()` write is always a real new application, never a defensive idempotent no-op racing to overwrite a correct timestamp.
- **`CooldownAt` non-aliasing**: confirmed `LabelAppliedAt` is a distinct map, and `HasExpiredCooldown`/`HasActiveCooldown` never see it — the design correctly avoids the spurious-wake-up bug that reusing `CooldownAt` would have caused.
- **Stale-cache-after-removal safety**: `removeAwaitingCILabel`/label removal never clears the `LabelAppliedAt` cache entry, but every read call site is gated by `hasLabel(item.Labels, ...)` first (directly, or via the sole caller's gate, e.g. `handleAutoMergeConvergence`'s `hasLabel` check ahead of `checkAutoMergeConvergence`), so a stale entry is never read — only overwritten by the next genuine re-application.
- **"Most recent application" correctness**: `TestLabelAppliedAt_AppliedRemovedReapplied_ReadsLatest` exercises the real guarded write path (`classifyCIFromCheckRuns` → `removeAwaitingCILabel` → re-apply) and proves the cache returns the second application's timestamp, with the mock's `fetchLabelAppliedAtFn` set to fail the test if ever called — proving zero fallback to a live fetch across the sequence.
- **Fail-open contract preserved**: every substituted call site keeps the identical `(time.Time, error)` signature and the same zero-time/error handling; no caller-side logic changed.
- **Required regression tests unchanged and passing**: `TestSettleAwaitingCIScan_CIWaitTimeoutBackstop_PausesRegardlessOfGateClaim` and `..._NoOpWithinTimeout` are untouched by the diff and pass.
- **Non-vacuous new tests**: the dedup test pairs the call-count assertion with a `CIFixCycles` correctness assertion; the cold-start/cross-poll tests assert both the correct value and the call count.
- **Build/vet/tests**: `go build ./...`, `go vet ./...` clean. `go test -race -timeout 5m ./...` — 3505 passed. The 4 failures (`cmd.TestExecute_ConfigYAMLApplied`, `cmd.TestExecute_ValidStagesReachesEngine`, `TestInvokeClaude_FabrikEnvVarsInjected` and its subtest) are all in files this PR never touched; verified independently that `TestExecute_ConfigYAMLApplied`/`TestExecute_ValidStagesReachesEngine` fail identically on `origin/main` (SIGINT-timing goroutine race, unrelated to this change) in a throwaway worktree, and `TestInvokeClaude_FabrikEnvVarsInjected` fails due to host environment leakage (nvm/Ghostty/Claude Code env vars), also environment-sensitive and unrelated.
- **`gofmt -l`**: flags 7 files, none of which are part of this PR's diff — pre-existing drift.
- **ADR-1314**: accurate, matches the implementation, records rejection rationale for reverse pagination, poll-native GraphQL, and the narrower `phase1Ctx` option, and states the two-engine-instance caveat explicitly.
- **CI on PR #1369**: all 5 checks passing (Test and vet, CodeQL ×2, Docs drift, llms-full).
- **Scope**: no doc updates needed (confirmed no `docs/state-machine.md`/`USER_GUIDE.md`/`stage-lifecycle.md` content is affected — pure internal cost optimization).

### Fixed

None — no issues found requiring changes.

### Blocking

None. PR is clean and ready for human review.

Reviewed the record-at-write implementation for FetchLabelAppliedAt deduplication. Verified all 10 write sites are correctly guarded, all 9 read call sites migrated, the CooldownAt-aliasing hazard is properly avoided, "most recent application" semantics hold under applied→removed→re-applied, and the two required backstop tests pass unchanged. Full test suite, build, and vet all pass (4 pre-existing, unrelated failures confirmed identical on main). No issues found — PR #1369 is ready for human review.

---
Used 61/100 turns, 7.8M input (1k raw + 7.6M cache-read + 150k cache-write) / 17k output tokens.

arbeithand:
*branch: fabrik/issue-1314 | commit: 9f5caeac | main: 4b2150ad30403215089a03ca5942999a6cf9a27c | 2026-08-03 02:17 UTC*

CI is running against the new push. Working tree is clean, all changes committed and pushed, and the review thread is resolved.

## Summary

Pruefer flagged a real correctness gap in `mutate.go:75`: `recordLabelAppliedAtNow`'s write sites guard on `item.Labels`, an in-memory snapshot that can be stale relative to GitHub's true state. Since `AddLabelToIssue` is idempotent, a stale-guard false re-add could previously overwrite a correct cached applied-at timestamp with a bogus `time.Now()`, silently resetting every downstream timeout (`CIWaitTimeout`, `ReviewWaitTimeout`, auto-merge convergence budget, merge-queue stall dwell) with no corresponding real event.

I evaluated this as valid and fixed it:
- `recordLabelAppliedAtNow` now only writes when the cache doesn't already hold a value for the label.
- Added `clearLabelAppliedAtNow`, wired into `syncLabelRemoval` (the single shared cache-sync tail-call for every removal of these 4 labels — traced and confirmed no removal site bypasses it except one), plus explicitly into `cleanupClosedIssueTransientLabels` (`poll.go`), which does bypass it.
- This preserves "most recent application" correctness (genuine remove clears the cache, so re-apply writes fresh) while making a spurious stale-guard re-add a no-op (cache stays non-empty, so the write is skipped).
- Updated the existing `TestLabelAppliedAt_AppliedRemovedReapplied_ReadsLatest` test (removal now clears rather than leaves stale) and added `TestLabelAppliedAt_StaleGuardDoesNotResetTimestamp`, which directly reproduces Pruefer's scenario.
- Updated ADR-1314 with a new "Stale-guard hardening" section documenting the fix and its residual, bounded scope.

Build, vet, and all targeted + full test suites pass (only the same pre-existing, unrelated `TestInvokeClaude_FabrikEnvVarsInjected` environment-sensitivity failure, confirmed identical on `main`). Committed as `9f5caeac`, pushed, and resolved the review thread.

---
**Threads addressed:**
- `engine/mutate.go:75` — resolved

arbeithand:
*branch: fabrik/issue-1314 | commit: 9f5caeac | main: 4b2150ad30403215089a03ca5942999a6cf9a27c | 2026-08-03 02:25 UTC*

I already have confirmed full-suite results from the earlier run (3507 passed, 3 pre-existing unrelated failures, 1 skipped). This second scoped run is redundant confirmation — no need to block on it. All validation criteria are satisfied. Let me finalize the report.

FABRIK_STAGE_COMPLETE marker check: PR mergeable=MERGEABLE, mergeStateStatus=CLEAN, branch rebased and up to date, all requirements verified, tests pass, no regressions.

## Validation Report

### Requirements: All verified

- [x] **Options evaluated honestly against corrected 10-call footprint** — ADR-1314 documents all four options (reverse pagination, record-at-write, poll-native GraphQL, narrow `phase1Ctx`) with rejection rationale for the three not chosen, including why the narrow fix is subsumed rather than merely deprioritized.
- [x] **Record-at-write selected and implemented** for the 4 engine-exclusive labels (`fabrik:awaiting-ci`, `fabrik:awaiting-review`, `fabrik:bot-reprompted`, `fabrik:auto-merge-enabled`); `stage:<name>:complete` explicitly and correctly scoped out (already mitigated by ADR-068's `archive_done_settle.go`), stated explicitly per the issue's requirement.
- [x] **Fail-open contract preserved** — `labelAppliedAt` keeps the identical `(time.Time, error)` signature; every substituted call site's caller-side logic is untouched (verified by diff inspection).
- [x] **"Most recent application" semantics** — `TestLabelAppliedAt_AppliedRemovedReapplied_ReadsLatest` exercises the real guarded write path through applied→removed→re-applied and proves the latest timestamp is read.
- [x] **Two-engine-instance caveat stated explicitly** — in both the ADR and code comments, fail-open, never fail-closed.
- [x] **Required regression tests unchanged and passing** — `TestSettleAwaitingCIScan_CIWaitTimeoutBackstop_PausesRegardlessOfGateClaim` / `..._NoOpWithinTimeout` pass.
- [x] **New call-count test proven non-vacuous** — `TestSettleAwaitingCIScan_LabelAppliedAt_DedupesWithinSinglePass` pairs the call-count assertion with a `CIFixCycles` behavioral assertion (verified by reading the test source).
- [x] **Reverse pagination not implemented** so its Link-header fail-open requirement doesn't apply — correctly deferred, not silently dropped (documented in ADR).
- [x] **Post-review hardening**: a real correctness gap found by Pruefer review (stale `item.Labels` guard causing spurious timestamp resets) was fixed with `recordLabelAppliedAtNow`'s write-once guard + `clearLabelAppliedAtNow` on removal, with a dedicated regression test (`TestLabelAppliedAt_StaleGuardDoesNotResetTimestamp`).

### Test Suite: PASSED

- 3507 passed, 1 skipped, 3 failed across 20 packages (`go test -race -timeout 5m ./...`)
- The 3 failures (`cmd.TestExecute_ConfigYAMLApplied`, `TestInvokeClaude_FabrikEnvVarsInjected` + subtest) are in files not touched by this PR's diff (`git diff origin/main..HEAD --name-only` confirmed) — pre-existing SIGINT-timing and host-environment-variable-leakage flakiness, matching what Review independently verified against `origin/main`.
- Build: clean. Vet: clean. `gofmt -l` on every file this PR touches: clean.
- All 5 relevant new/updated tests (`engine/label_applied_at_test.go` ×5, `internal/itemstate/mutation_test.go` ×3) pass individually.

All 9 relevant `FetchLabelAppliedAt` read call sites correctly migrated to `labelAppliedAt` (verified by grep); `archive_done_settle.go` and the interface declaration correctly untouched.

### Branch State

Rebased onto `origin/main` (already up to date), working tree clean, all commits pushed. PR #1369: `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`.

### Verdict: READY TO MERGE

Validation passed. PR #1369 mergeable: MERGEABLE, mergeStateStatus: CLEAN. All requirements verified (record-at-write implemented for the 4 engine-exclusive labels, options honestly evaluated in ADR-1314, fail-open and most-recent-application semantics tested), full test suite passes (3507 passed, 3 pre-existing unrelated failures unchanged from main), no regressions, branch clean and pushed.

---
Used 26/100 turns, 2.5M input (43 raw + 2.4M cache-read + 102k cache-write) / 7k output tokens.

arbeithand:
🏭 **Fabrik merge-train** — Landed via batch PR #1370.

PR #1455, opened 2026-08-08
merge-train: eject Queued members off the batch when unresolved review findings arrive

### LINKED ISSUE #1208: merge-train: Queued members process no PR feedback — review findings and comments are ignored for a whole batch cycle
## Problem

While an item sits in the `Queued` holding stage awaiting a merge-train batch, Fabrik processes **no PR feedback at all** — not review findings, not human comments. The admission guard shared by both catch-up phases in the poll loop skips any item whose stage is a holding stage (`HoldingStage: true`, which `Queued` is) before either phase runs:

```go
stage := stages.FindStage(e.cfg.Stages, item.Status)
if stage == nil || stage.CleanupWorktree || stage.HoldingStage || stage.Unmanaged {
    continue
}
```

That skip is correct for *dispatch* — a batched member must not run its stage prompt individually while queued, which is what the dispatch guards elsewhere in `engine/item.go` protect. But it is too broad: it also suppresses *feedback processing*, which is a different concern from running a stage.

- **Review-reinvoke never evaluates a Queued item** — unresolved review threads are neither detected nor addressed.
- **Comment processing never fires** — a human commenting on a queued issue gets no response.

### Why this is worse than the Validate race (#1207)

#1207 (closed 2026-07-29) lost a finding that landed in the seconds between Validate completing and the merge firing. This loses **everything** that arrives across a whole batch cycle: batch formation (`max_batch_size`, default 5), trial-branch assembly, CI, possibly inline conflict resolution and bisection — easily tens of minutes.

And it is an active window, not a quiet one: Pruefer re-reviews on every new head SHA, and train activity produces new SHAs. So it is precisely when a queued member is most likely to attract review comments that Fabrik is guaranteed not to read them.

### Why ejection, not in-place processing

Processing feedback in place is the wrong shape. Addressing a finding means pushing a commit, which changes the member's head SHA **mid-batch** — invalidating a trial branch the train may have already assembled, CI'd, or bisected. That would race against the train's own state machine.

The train already knows how to remove a member: `engine/merge_train.go`'s `ejectMember` is called on fetch failure, unresolvable conflict, and bisection isolation. This issue is a new *trigger* for that existing mechanism, not new removal machinery — with one important behavioral difference (see Requirements): the three existing ejection causes deliberately leave the member **in the Queued column** to retry in a future train, because nothing about those causes requires the member to leave Queued. A review-finding ejection is different — the whole point is to get the finding in front of the normal reinvoke path, which does not run on Queued items, so this ejection must also move the item's board Status off of `Queued` (e.g., back to `Validate`).

## Summary

Detect unresolved review-thread feedback on a Queued merge-train member's linked PR, and when found, eject the member from the batch (reusing the existing `ejectMember` mechanism) **and** move it off the `Queued` column so the ordinary review-reinvoke path can address the finding. Once addressed, the member re-queues on its next Validate completion and joins a later batch. Surviving batch members are unaffected.

## Requirements

- A Queued member's linked PR is checked for unresolved review-thread feedback. Reuse the existing `buildReviewThreadComments` / thread-resolution logic (`reviewThreads { isResolved }` is already fetched by the board query; no new GraphQL surface is required).
- When unresolved review-thread feedback is found on a Queued member:
  - The member is ejected from its current (or forming) batch via the existing `ejectMember` mechanism.
  - Unlike the three existing ejection causes (fetch failure, unresolvable conflict, bisection isolation) — which leave the member in `Queued` to retry in a future train — this ejection must move the member's board Status off of `Queued` (to `Validate`, or another column from which the ordinary reinvoke/dispatch loop can reach and process it). Leaving it in `Queued` would reproduce this same bug.
  - The ejection reason/comment is logged and worded distinctly from the conflict/fetch-failure/bisection ejection reasons, so operators can tell from the ejection comment alone why a given ejection happened.
- Once the member is back on a stage the reinvoke loop covers, the unresolved finding is addressed through the existing review-reinvoke path (the same mechanism that already handles inline PR review threads on non-Queued issues).
- After the finding is addressed and Validate completes again, the member re-enters `Queued` normally and is eligible for a later batch.
- Surviving members of the batch the ejected member was removed from are unaffected — this must compose with the existing re-form-after-ejection behavior the train already has for its other ejection causes.
- A member must not oscillate between `Queued` and `Validate` without bound. The existing `MaxReviewCycles` bound on the review-reinvoke loop must continue to apply across an eject/re-queue cycle rather than resetting when the member re-enters `Validate` — the same cap that already governs ordinary CHANGES_REQUESTED reinvocation must govern this path too, so a member that keeps attracting findings is eventually escalated (paused) rather than looping forever. (`MaxMergeTrainEjections`, the train's own existing ejection-count pause cap, is a separate, already-existing bound on repeated *train* ejections and is not a substitute for this — a member could in principle be ejected for this new reason only once or twice per train cycle while still cycling Queued↔Validate many times if the review-reinvoke bound were not honored.)
- `docs/state-machine.md` documents this new Queued feedback-detection-and-ejection path (detection trigger, where the member goes, how it differs from the three existing ejection causes, and how the cycle bound applies).

## Scope

**In scope:**
- Detecting **unresolved review threads** (review findings) on a Queued member's PR and ejecting+rerouting the member to address them.
- The logging/comment distinction between this ejection cause and the three existing ones.
- Ensuring the existing `MaxReviewCycles` bound still applies across the eject/re-queue cycle.
- Updating `docs/state-machine.md`.

**Out of scope (for this issue):**
- Triggering ejection on a plain **human comment** that is not a formal review finding (e.g., a question). The original problem statement calls out that human comments are also currently ignored on Queued items, but ejecting on every comment would make the train fragile (a question isn't necessarily a change request). This issue's Definition of Done only requires ejection on an unresolved review thread; comment-triggered ejection (or some lighter-weight response to a plain comment on a Queued item, short of ejection) is left as a candidate follow-up issue rather than folded in here.
- **Whether to also check for unresolved threads at batch-formation time** (refusing to admit a member with unresolved threads into a forming batch, in addition to ejecting mid-flight members that develop unresolved threads after joining) is a design/implementation decision, not a requirements decision — left to Plan to decide and to state explicitly which case(s) it implements. The Definition of Done only requires that a Queued member *that receives* an unresolved review thread is ejected; it does not mandate a specific admission-time check as well, though Plan may choose to add one if it simplifies the mid-flight logic.
- Any change to the review-reinvoke path's own detection/addressing logic — this issue only concerns getting a Queued member back onto a path that path already covers.

## Prior Art / Context

- `engine/poll.go` — the `HoldingStage` skip shared by both catch-up phases (Phase 1: dependency/review-gate/CI-gate/review-reinvoke; Phase 2: gated stage advancement).
- `engine/item.go` — the per-item dispatch guards that correctly prevent a batched member from running its stage prompt individually; this issue must not weaken those.
- `engine/merge_train.go` — `ejectMember` (existing ejection posting/counting/pause mechanism) and its three current call sites: unfetchable linked PR / missing head SHA, unresolvable inline conflict, and bisection-isolated poisoner. Also `MaxMergeTrainEjections`, the existing per-member consecutive-ejection pause cap (a distinct bound from `MaxReviewCycles`, see Requirements).
- `engine/reviews.go` — `buildReviewThreadComments` / `currentHeadReviewThreadComments`, the existing inline-review-thread detection primitives this issue reuses.
- `engine/catch_up_handlers.go` — the existing `MaxReviewCycles`-bounded review-reinvoke loop this issue's ejected members must re-enter and be bound by.
- `docs/state-machine.md` §"Queued (Holding Stage)" and §6 "Review Gate and Review Reinvoke" — the as-built sections this issue must extend.
- ADR-059 — merge-train design.
- #1207 (closed 2026-07-29) — the sibling race at Validate: an unresolved-thread guard on `attemptMergeOnValidate` that prevents auto-merge when fresh current-head review threads exist. Already landed, which means a member with unresolved threads at the moment Validate would hand off to Queued is now guarded there — narrowing this issue's practical exposure to the mid-flight case (a finding that lands *after* the member is already in Queued), though the code path this issue must fix has no such guard itself.
- #1205 (closed 2026-07-28) — review-body processing, which widens what counts as a "finding" that the review-reinvoke path addresses; whatever this issue reuses from that path inherits its widened definition automatically.

## Risks / Dependencies

- **Depends on `ejectMember` and the review-reinvoke path's existing behavior** — this issue is additive (new trigger + new routing), not a rewrite of either, so regressions in either area are a risk if the two are not composed carefully (e.g., moving a member off `Queued` must not itself look like a train "member landed" or "member abandoned" event to the rest of `merge_train.go`'s bookkeeping).
- **Interaction with concurrent train activity**: the member being ejected for this reason may be ejected while a trial branch is already assembled/CI'd/mid-bisection for the batch it's in — the same re-form-survivors behavior the existing three ejection causes already rely on must hold here too.
- **Stall risk**: explicitly called out in Requirements — the fix must not let a member cycle Queued → Validate → Queued indefinitely. Confirming `MaxReviewCycles` composes correctly across the eject/re-queue transition (rather than appearing to the counter as a fresh stage entry) is the key risk to verify in testing.

## Tests (from Definition of Done)

- A finding (unresolved review thread) arriving on a Queued member ejects it.
- The batch re-forms correctly around the ejection; surviving members land normally.
- The ejection reason/comment for this cause is distinguishable from the conflict/fetch-failure/bisection ejection reasons.
- The review-cycle bound (`MaxReviewCycles`) still escalates (pauses) a member that keeps attracting findings across repeated eject/re-queue cycles, rather than resetting or looping forever.
- `go test -race ./...` passes.

arbeithand:
Closes #1208

## What this does

While a merge-train member sits in the `Queued` holding stage awaiting a batch, Fabrik previously processed **no PR feedback at all** — the `HoldingStage` admission exclusion that correctly suppresses per-item dispatch also silently blacked out review-reinvoke and comment processing for the entire batch cycle (batch formation, trial assembly, CI, bisection — easily tens of minutes, and up to `CIWaitTimeout` while a worker blocks in `pollTrainCI`).

This PR adds a new, dedicated per-poll settle scan (`settleQueuedReviewFindings`) that detects an unresolved review-thread finding on a Queued member's linked PR and ejects the member off `Queued` — via the existing `ejectMember` mechanism, extended with a `stayInQueue` parameter so this cause's ejection comment reads the opposite of the three pre-existing causes — back onto the stage immediately preceding the holding stage (derived structurally by `Order`, not hardcoded to `"Validate"`), where the ordinary, completely unmodified review-reinvoke path picks it up.

## Key design points

- **Concurrency-safe against a live merge-train worker.** The scan ejects directly only when no worker is in flight for the repo; otherwise it leaves a mutex-guarded pending-eject signal that the worker itself consumes at two checkpoints (`runMergeTrainWorker`'s re-form loop and `landOneAtATime`'s per-singleton loop), mirroring the existing runaway-guard's "poll writes a signal, worker checks it at a checkpoint" shape. A flagged member's trial is always discarded, even if green.
- **`MaxReviewCycles` composes across the eject/re-queue cycle "for free."** The reroute is a plain board-Status move — it never touches `stage:Validate:complete` or calls `EngineCyclesCleared` — so the existing cycle-bounded review-reinvoke loop picks the rerouted member up automatically and eventually escalates (pauses) a member that keeps attracting findings, rather than oscillating forever. Pinned by a dedicated test that drives a full eject→reroute→reinvoke cycle and asserts the counter never resets.
- **`MaxMergeTrainEjections`** (the train's pre-existing ejection-count pause cap) fires independently, confirmed by its own focused test.
- Reroute happens before the ejection comment/counter, so a transient board-mutation failure can't double-post or double-count.

## Testing

- `go test -race ./...` passes across every package except three pre-existing, network-dependent `cmd` tests (`TestExecute_ConfigYAMLApplied` — already `-skip`'d in CI — and two siblings) confirmed to fail identically on the pre-issue base commit, unrelated to this change.
- `go vet ./...` clean.
- New/updated tests cover: the distinguishable ejection wording, the reroute function's status-move-only behavior, the settle scan's direct-eject/pending-signal/skip paths, both worker checkpoints (using the `trainValidateFn` test seam and a real-git integration test), and the `MaxReviewCycles`/`MaxMergeTrainEjections` composition tests described above.

## Docs

- `docs/state-machine.md` §6.16 (new) — full as-built spec for the settle scan, modeled on §6.14's Awaiting-CI Settle Scan; cross-referenced from the existing "Queued (Holding Stage)" section.
- `docs/llms-full.txt` regenerated.
- ADR 1208 added.
- `CLAUDE.md` architecture bullet added for `engine/queued_review_settle.go`.


arbeithand:
*branch: fabrik/issue-1208 | commit: 98a6d5d3 | main: 62a175ec46a30ad40a94da5113c2373fcde418d4 | 2026-08-08 20:27 UTC*

All work is committed, tested, and documented. Let me provide the summary and complete the stage.


Implemented the full plan: `ejectMember` gained a `stayInQueue` parameter for distinct ejection wording; a new `settleQueuedReviewFindings` settle scan detects unresolved review threads on Queued merge-train members and ejects them off `Queued` (directly, or via a pending-eject signal consumed at worker checkpoints when a train is mid-flight) so the ordinary review-reinvoke path can address them; `MaxReviewCycles` composes across the eject/re-queue cycle without any new counter. All work is tested (`go test -race ./...` clean apart from 3 pre-existing, network-dependent `cmd` test failures confirmed unrelated to this change), `go vet` clean, and documented (`docs/state-machine.md` §6.16, ADR-1208, CLAUDE.md). Plan checklist fully checked off; PR marker emitted.

---
Used 50/100 turns, 20.9M input (99 raw + 20.5M cache-read + 419k cache-write) / 26k output tokens.

arbeithand:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-1208 | commit: de5cb547 | main: 62a175ec46a30ad40a94da5113c2373fcde418d4 | 2026-08-08 20:42 UTC*

No formatting issues, and no new external PR feedback since Implement finished. Everything is clean, tested, and pushed. Let's finalize the review summary.

Reviewed the #1208 implementation (Queued review-finding ejection). The core design was solid and well-tested, but found and fixed one real concurrency gap: `landGreenBatch`'s main-moved rebase-and-revalidate loop could spend a full CI wait without ever consulting the pending-eject signal, letting a flagged member ride a rebased green trial straight to landing — unlike the other two Hook 2 checkpoints. Added a third checkpoint there (discard trial, return, let survivors re-form on the next poll), plus a regression test and doc/ADR updates. Full `go build`/`go vet`/`go test -race` pass (only 3 pre-existing, environment-dependent failures confirmed identical on `origin/main`); `docs/llms-full.txt` regenerated. PR is ready for human review.

---
Used 78/100 turns, 12.1M input (153 raw + 12.0M cache-read + 178k cache-write) / 33k output tokens.

arbeithand:
*branch: fabrik/issue-1208 | commit: 56950e07 | main: 62a175ec46a30ad40a94da5113c2373fcde418d4 | 2026-08-08 21:04 UTC*

Thread resolved. The issue body doesn't need updating (this was a review-comment fix, not a spec change), so no `FABRIK_ISSUE_UPDATE_BEGIN` block is needed.

**Summary:** Pruefer's finding was valid and confirmed by tracing the code: `settleQueuedReviewFindings` branched on repo-level `mergeTrainWorkerActive`, but a worker's dispatched batch is capped to `effectiveMaxBatchSize` (default 5), so a Queued member beyond that cap could never have its pending-eject signal consumed — reproducing the exact blackout #1208 fixes, just for overflow members. Fixed by snapshotting the dispatched batch's issue numbers onto an immutable `mergeTrainWorkerState.batchNumbers` field and branching on actual batch membership instead. Added regression tests, updated `docs/state-machine.md` §6.16 and ADR-1208, regenerated `docs/llms-full.txt`, verified full build/vet/race-test suite (only the same 3 pre-existing, environment-dependent failures remain), committed, pushed, and resolved the review thread.

---
**Threads addressed:**
- `engine/queued_review_settle.go:43` — resolved

arbeithand:
*branch: fabrik/issue-1208 | commit: 56950e07 | main: 62a175ec46a30ad40a94da5113c2373fcde418d4 | 2026-08-08 21:08 UTC*

`mergeable: MERGEABLE`, `mergeStateStatus: CLEAN` — the earlier `BLOCKED` state has resolved (was likely pending review/CI checks that have since settled). All gate conditions are satisfied. Let me write the validation report.

## Validation Report

### Requirements: All verified against `.fabrik-context/issue.md`

- [x] **Detect unresolved review-thread feedback on a Queued member's PR** — `settleQueuedReviewFindings` (`engine/queued_review_settle.go`) is a dedicated per-poll settle scan (ADR-1270 pattern #6), sourced from `board.Items`/`groupQueuedByRepo`, deep-fetches each Queued item via `FetchItemDetails`, and checks `currentHeadReviewThreadComments` — the ADR-1207-canonical current-head-scoped primitive, exactly as Research/Plan specified.
- [x] **Eject via existing `ejectMember` mechanism** — `ejectQueuedMemberForReviewFindings` calls `ejectMember` with the same comment/counter/pause machinery reused unchanged (`diag=nil, otherMembers=nil`, per ADR-1420's contract for non-combined-Validate causes).
- [x] **Move off `Queued` (unlike the other 3 causes)** — `ejectMember` gained a `stayInQueue bool` param; all 5 pre-existing call sites pass `true` (unchanged wording/behavior verified by diff); the new cause passes `false` with distinct closing text. `rerouteQueuedMemberOffHolding` moves board Status to `stageBeforeHolding` (structurally derived by `Order`, not hardcoded `"Validate"`) *before* the ejection comment/counter — verified this ordering prevents duplicate comments on transient failure (`TestEjectQueuedMemberForReviewFindings_RerouteFailure_NoCommentNoCount`).
- [x] **Ejection reason distinguishable from the other 3 causes** — verified the wording differs materially: "This issue has left the Queued column so the unresolved review-thread finding above can be addressed via the normal review pipeline..." vs. "This issue remains in the Queued column and will be retried in a future train...".
- [x] **Finding addressed through existing, unmodified review-reinvoke path** — the reroute preserves `stage:Validate:complete` and never calls `EngineCyclesCleared`, so `handleReviewGate`/`dispatchWithCycleLimit` pick it up "for free" on the next poll (verified by reading `rerouteQueuedMemberOffHolding` and confirmed live in the composition test below).
- [x] **Member re-queues after Validate completes, joins later batch** — unchanged existing `advanceToQueued` path; no regression introduced.
- [x] **Surviving batch members unaffected; composes with re-form-after-ejection** — `applyPendingReviewEjects` is called at all three worker checkpoints (`runMergeTrainWorker`'s re-form loop, `landGreenBatch`'s main-moved rebase loop — added during Review to close a gap Pruefer found, `landOneAtATime`), discarding only the flagged trial/member and re-forming with `remaining`.
- [x] **`MaxReviewCycles` bound continues to apply across eject/re-queue, doesn't reset** — directly proven by `TestEjectQueuedMemberForReviewFindings_MaxReviewCyclesComposesAcrossCycle`, which drives the actual eject→reroute→`handleReviewGate` cycle in a loop and asserts `ReviewCycles("Implement")` equals exactly `MaxReviewCycles` and `fabrik:paused` fires — not an inference, an executed assertion.
- [x] **`docs/state-machine.md` documents the new path** — §6.16 added (styled on §6.14), with a cross-reference from the existing Queued section; `docs/llms-full.txt` regenerated and confirmed byte-identical to a fresh regen (no drift).

### Test Suite: PASSED

- 3805 passed, 1 skipped, 3 pre-existing failures across 20 packages (`go test -race -timeout 5m ./...`)
- The 3 failures (`TestInvokeClaude_FabrikEnvVarsInjected` — leaks host `NVM_INC` env var; `TestExecute_ConfigYAMLApplied` — network-dependent GraphQL board fetch during SIGINT) were verified to fail identically on a fresh `origin/main` worktree — confirmed pre-existing and unrelated to this change, not a regression.
- Race detector: clean (no new races)
- `go build ./...`: clean
- `go vet ./...`: clean

- All 5 pre-existing `ejectMember` call sites verified to pass `stayInQueue=true`, preserving original wording exactly.
- Diff scoped to `engine/{merge_train,queued_review_settle,poll,engine}.go` plus tests/docs — no unrelated files touched.

### Issues Found: None

### Branch State

- Rebased onto latest `origin/main` (no-op — already current)
- Working tree clean, all commits pushed
- PR #1455: `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`

### Verdict: READY TO MERGE

Validation passed. PR mergeable: MERGEABLE, mergeStateStatus: CLEAN. All 10 requirements verified against the issue spec, including a direct executed test proving `MaxReviewCycles` composes across the eject/re-queue cycle without resetting. Full suite: 3805 passed, 3 pre-existing failures confirmed identical on origin/main (unrelated), race-clean, vet-clean, build-clean. `docs/llms-full.txt` confirmed in sync with no drift.

---
Used 25/100 turns, 2.5M input (47 raw + 2.5M cache-read + 90k cache-write) / 6k output tokens.

Issue #1408, opened 2026-08-05
CI wait timeout is unrecoverable: clearing fabrik:paused leaves the item with no owner


arbeithand:
## Problem

An item that has once hit the CI wait timeout can never be resumed. Removing `fabrik:paused` — the remedy the pause comment itself instructs — returns the item to a state where **no owner will process it**, silently, for as long as the engine runs.

Observed on `verveguy/liminis-context-graph#342` on 2026-08-04. The linked PR #344 was `CLEAN` / `APPROVED` / `MERGEABLE` with every check green since 13:25:44Z. The issue sat untouched from 20:00:57Z to at least 00:24Z the following day — 4h23m and counting — carrying `fabrik:awaiting-ci` with no `stage:Validate:complete` and no `fabrik:paused`.

### The deadlock

`settleAwaitingCIScan` (`engine/ci_settle.go:190`) runs the unconditional `CIWaitTimeout` backstop ahead of the Phase 1 handler chain. When the deadline has elapsed it calls `pauseForCITimeout` and then `continue`s, skipping the handler chain entirely.

`pauseForCITimeout` (`engine/ci.go:459`) opens with:

```go
if hasCIGatePauseComment(item, stage) {
    e.logf(item.Number, "ci-timeout", "CI-gate pause comment already posted this poll — skipping duplicate\n")
    return
}
```

and `hasCIGatePauseComment` (`engine/ci.go:445`) scans the issue's **entire comment history**:

```go
for _, c := range item.Comments {
    if strings.Contains(c.Body, timeoutWant) || strings.Contains(c.Body, cycleWant) {
        return true
    }
}
```

Its doc comment says *"a redundant `pauseForCITimeout`/`pauseForCIFixCycleLimit` call — reachable **within the same poll** when `addCompleteLabelAndRemoveCI`'s two-call label swap races"*. The implementation has no per-poll scoping. Once a timeout comment exists on the issue, the guard returns `true` on every subsequent poll, forever.

So after a human clears `fabrik:paused` to resume:

| owner | outcome |
|---|---|
| `itemMayNeedWork` / `itemNeedsWork` | refuses — `engine/item.go:409` declines a `wait_for_ci` stage while `fabrik:awaiting-ci` is set |
| catch-up Phase 1 | excluded by design — `engine/poll.go:1126` routes awaiting-ci items to the dedicated settle scan (#1270) |
[...]

PR #1466, opened 2026-08-09
fix: exempt transient api_error exits from max_retries

### LINKED ISSUE #1458: A transient api_error exit is charged against max_retries and pauses the issue, though the stage never ran
## Problem

A Claude invocation that exits with `terminal_reason="api_error"` never ran the stage — yet it is charged against `max_retries` exactly like a genuine stage failure, and three of them pause the issue.

Observed live on **2026-08-08**, 21:19–21:32 UTC: **eight** such exits across two issues, every one at 1 turn and `$0.0000`.

```
21:25:22 [#1408 claude] used 1 turns, $0.0000
21:25:22 [#1408 claude] error exit with unmatched terminal_reason="api_error" (not classified as usage limit)
21:25:22 [#1408 warn] claude invocation issue: claude exited with error: exit status 1
21:25:25 [#1408 wait] stage "Validate" did not complete — will retry after 5m0s
21:32:18 [#1408 claude] used 1 turns, $0.0000
21:32:18 [#1408 claude] error exit with unmatched terminal_reason="api_error" (not classified as usage limit)
21:32:22 [#1408 escalate] stage "Validate" failed 3 time(s) — pausing issue
```

#1408 was paused with `stage:Validate:failed` on a PR (#1445) that was `MERGEABLE`/`CLEAN` with all six checks green, and stayed paused for roughly an hour with the whole dependent chain (#1410 → #1441) blocked behind it. #1208 took the other five and was also left paused. **No Claude invocation ran anywhere on the board for ~50 minutes**, because both active lane heads were paused and every remaining issue was chained behind them.

Nothing was wrong with either issue's work. The transient was Anthropic-side.

## Root cause

`engine/claude.go:1290-1305`. Structural classification recognises exactly one `terminal_reason`:

```go
if msg, detected := classifyUsageLimitExit(resp, usage); detected {
    ...
} else if resp.TerminalReason != "" && resp.TerminalReason != usageLimitTerminalReason {
    // Diagnostic-only: records any other non-empty terminal_reason
    // seen on an error exit (e.g. "rapid_refill_breaker") ...
    // Never used to trigger classification.
    claudeLog(issueNumber, "claude", "error exit with unmatched terminal_reason=%q (not classified as usage limit)\n", resp.TerminalReason)
}
return text, false, usage, fmt.Errorf("claude exited with error: %w", runErr)
```

`classifyUsageLimitExit` (`:102`) matches only `blocking_limit`. Everything else — including `api_error` — is logged for diagnostics and then falls through to the generic error, reaching `StageRetryIncremented` at `engine/item.go:739` and, at `MaxRetries`, `stage:<name>:failed` + `fabrik:paused`.

The diagnostic-only posture is deliberate and correct in origin: #1183 removed prose-based detection after issue #1178's Implement stage quoted a usage-limit message and suspended dispatch account-wide for ~11 hours. **This issue must not undo that.** The ask is not to loosen classification generally; it is to add one further *structural* signal, from the CLI's own result object, with the same conservative guard.

Verified current as of this writing (2026-08-08, against `origin/main` post-#1447): the line numbers and code shape above are still accurate. `engine/item.go:1216`'s `claudeUsageLimitError` → `handleUsageLimitExit` route (R2's reuse target) is unchanged by #1199's landing — see the Related section below for why that matters.

## Summary

Add one further structural `terminal_reason` match — `api_error` — alongside the existing `blocking_limit` check in `engine/claude.go`, and route it through the same "stage never ran" handling `engine/item.go` already uses for usage-limit exits (`StageAttempted` recorded, `StageRetryIncremented` never called, no label, no comment). The dispatch cooldown that already exists on that route is the mechanism expected to bound retries; R6 requires confirming that cooldown actually applies on the code path that produced #1208's 31-second/5-invocation burst before relying on it.

## Requirements

**R1 — Classify `api_error` as a did-not-run exit.** Structural only: `resp.TerminalReason == "api_error"` from the parsed result object. Never from output prose, never from the generic `runErr` text. An unparseable-JSON exit keeps falling through to the generic error, exactly as today.

**R2 — Record `StageAttempted`, not `StageRetryIncremented`.** The stage never ran, so the exit must not consume a `max_retries` slot, must not apply `stage:<name>:failed`, and must not pause the issue. The dispatch cooldown still applies so the item cannot hot-loop. This is precisely `fabrik:claude-limit`'s shape (ADR-1119), and `engine/item.go:1216` is the existing route that already does exactly this — reuse it rather than adding a parallel path.

**R3 — No account-wide suspension.** Unlike `blocking_limit`, an `api_error` is transient and per-invocation. It must not touch `claudeSuspendedUntilTime` or anything resembling the ADR-1183 suspension. Per-issue effect only.

**R4 — No durable label, no issue comment.** The condition self-resolves on the next invocation; a label would be indistinguishable from the orphaned-durable-state leak ADR-1183's sweep exists to clean up, and a comment for a self-healing event is noise. Log only. If the retry-after-cooldown also fails for a *different* reason, that flows into the normal failure path, which already comments. (Same rule applied in #1414: comment when a human must act, log when the engine self-heals.)

**R5 — Verify the guard against real values before relying on them.** `classifyUsageLimitExit` excludes on `usage.TurnsUsed > 0 && usage.CostUSD > 0`. The observed `api_error` exits report **1 turn and `$0.0000`**, so they pass that exclusion only because cost is zero. Confirm whether `CostUSD` is *exactly* zero here or a small value the `%.4f` log format rounds to `0.0000` — if it is non-zero, the conjunctive guard silently stops matching and the fix is inert. Do not infer the value from the log line; assert it against a captured result object.

**R6 — Bound the loop independently of the cooldown.** #1208's five exits landed at 21:19:48, 21:19:55, 21:20:03, 21:20:12 and 21:20:19 — **31 seconds for five invocations**, not the 5-minute cooldown seen on #1408's stage-path retries. Establish why that path re-invoked in seconds (comment-processing vs stage dispatch is the likely split) and make sure removing the `max_retries` charge does not convert a bounded 3-strike pause into an unbounded fast retry. If the cooldown does not apply on that path, this issue must bound it some other way — an unbounded no-charge retry is a worse outcome than the current spurious pause.

## Scope

In scope: `engine/claude.go` classification, the `engine/item.go` routing that distinguishes attempted-but-not-run, and an ADR numbered after this issue.

Out of scope: other unmatched `terminal_reason` values (e.g. `rapid_refill_breaker`) — add them only when observed in the wild with evidence, per the same discipline that kept this list at one entry until now. Retry/backoff policy for genuine stage failures. The account-wide suspension mechanism.

## Open Questions

None outstanding. R5 and R6 are empirical checks to perform during Research (assert real `CostUSD`/`TerminalReason` values; trace which dispatch path produced #1208's sub-minute burst), not product decisions requiring a human answer — Research should resolve them with evidence and report findings, not block back to Specify.

## Acceptance

1. An invocation whose parsed result carries `terminal_reason="api_error"` records `StageAttempted` and does **not** increment the retry counter. Assert on the **counter**, not on the absence of a pause — a test that only checks "not paused" passes for three different reasons and is vacuous.
2. Three consecutive `api_error` exits do not apply `stage:<name>:failed` or `fabrik:paused`, and a subsequent successful invocation completes the stage normally. This is the observed #1408 scenario end to end.
3. Neutralising the fix (restoring the fall-through to the generic error) turns tests 1 and 2 red. State it in the PR.
4. An `api_error` exit does **not** set any account-wide suspension: `claudeSuspendedUntilTime` is unchanged.
5. A genuine stage failure with no `terminal_reason` still increments the retry counter and still pauses at `MaxRetries` — unchanged. This is the regression that matters most; the whole risk of this change is over-classifying a real failure as a did-not-run.
6. An unparseable-JSON error exit still falls through to the generic error path, unchanged (#1183).
7. `go test -race ./...` green.

## Related

- ADR-1119 / `fabrik:claude-limit` — the "stage never ran" precedent this mirrors
- ADR-1183 — why classification is structural-only; do not regress it
- #1199 (**closed**, landed via #1447/ADR `1199-slice-budget-separate-from-failure-counter`) — the same question for a *different* budget (turn-cap preemption). It resolved as a **separate `SliceRetryIncremented` counter**, independent from `StageRetryIncremented` — because a turn-cap preemption is a stage that *did* run and made real progress before being cut off, which is a different accounting problem from this issue's case (an exit where the stage never started at all). Confirmed against `origin/main`: `engine/item.go:1216`'s `claudeUsageLimitError`/`handleUsageLimitExit` route — R2's reuse target — is untouched by #1199's change. **Conclusion for Research/Plan: reuse the ADR-1119 route (R2), not #1199's slice counter.** The "don't invent a second mechanism" caution in the original issue text was about not duplicating the *did-not-run* accounting pattern; #1199 turned out to answer a sibling-but-distinct question, so this is not a case of two independent answers to the same question — it's confirmation that the right precedent (ADR-1119) was already correctly identified.
- #1414 — the same question again for failing `--resume` sessions, already answered consistently in that issue's thread

arbeithand:
Closes #1458

## Summary

A Claude invocation exiting with `terminal_reason="api_error"` never ran the stage, but was previously charged against `max_retries` like a genuine failure — three such exits paused the issue even when nothing was wrong with the work. Observed live on 2026-08-08: eight such exits across two issues (#1408, #1208), all at 1 turn and `$0.0000`; #1408 was paused with a mergeable, all-green PR blocked behind it, and no Claude invocation ran anywhere on the board for ~50 minutes.

## Changes

- **`engine/claude.go`**: new `claudeAPIErrorExit` sentinel type and `classifyAPIErrorExit` classifier, structurally parallel to the existing `blocking_limit` usage-limit detector — `terminal_reason == "api_error"` from the CLI's own parsed result object only, never output prose. Wired into `interpretClaudeResult`'s cascade right after the usage-limit branch.
- **`engine/item.go`**: new `handleAPIErrorExit` routed from `finalizeStageOutcome` via `errors.As`. Records `StageAttempted` (so the normal dispatch cooldown applies) but never `StageRetryIncremented` — no `stage:<name>:failed`, no `fabrik:paused`, and (unlike the usage-limit/apiKeyHelper handlers) **no label and no comment** — a smaller, log-only "did not run" shape, since the condition is per-invocation and self-resolving.
- **`engine/comments.go`**: deliberately untouched. Because `claudeAPIErrorExit` is a distinct Go type from `claudeUsageLimitError`, it never triggers the account-wide Claude suspension and never bypasses the comment-processing circuit breaker — both of which are appropriate only for a genuine account-wide usage limit, not a transient per-invocation error.
- **`docs/state-machine.md`** (+regenerated `docs/llms-full.txt`): new §7.3a documenting the exemption and contrasting it with §7.3's usage-limit handling.
- **`adrs/1458-transient-api-error-exemption.md`**: records why a distinct sentinel type was used instead of reusing `claudeUsageLimitError`, why the exclusion guard is kept despite an unverified `CostUSD` sample, and why the comment path needed no change.

## Why a new sentinel type instead of just adding `"api_error"` to the existing usage-limit check?

`claudeUsageLimitError` isn't inert — by Go type, it also triggers account-wide Claude suspension and exempts the invocation from the comment-processing circuit breaker. Both are correct only for a genuine account-wide usage limit. Reusing it for `api_error` would have suspended Claude dispatch account-wide on a transient per-issue blip, and would have left the comment-triggered dispatch path with no bound at all — converting a bounded 3-strike pause into a potentially unbounded fast retry loop (the exact shape #1208's 5-exits-in-31-seconds burst showed is reachable). A distinct type avoids both by construction: the existing `errors.As(&claudeUsageLimitError{})` checks simply never match it.

## Acceptance criteria verification

1. ✅ `TestAPIErrorExit_ExemptFromRetryNoLabelNoComment` asserts `snap.Attempts("Research") == 0` directly (not just "not paused").
2. ✅ `TestAPIErrorExit_ThreeConsecutiveThenSuccess_CompletesNormally` — three `api_error` exits (with `MaxRetries: 2`, which would otherwise have paused after 3) then a successful invocation completes normally.
3. ✅ **Verified manually**: stubbed `classifyAPIErrorExit` to always return `false`, confirming `TestClassifyAPIErrorExit`'s zero-cost cases and `TestInterpretClaudeResult_APIErrorExit_ReturnsSentinelError` go red, then restored the fix (clean revert, verified with `git diff`). Note: the item-level `TestAPIErrorExit_*` tests mock the `ClaudeInvoker` to return the sentinel directly (per this repo's testing convention — mock at the interface boundary, not a fake binary), so they exercise the routing half (R2) independent of the classification half (R1); the classification-level tests in `claude_api_error_test.go` are what catch a classifier regression, and they did.
4. ✅ `TestAPIErrorExit_DoesNotActivateAccountWideSuspension` asserts `claudeSuspendedUntilTime` is unaffected.
5. ✅ `TestGenuineStageFailure_StillCountsAgainstRetries` (pre-existing) already covers this — a plain error with no `terminal_reason` matches no sentinel type.
6. ✅ `TestInterpretClaudeResult_APIErrorExit_UnparseableJSON_NotClassified`.
7. ✅ `go build`, `go vet ./...` clean. `go test -race ./...`: 3843 passed, 1 unrelated pre-existing failure (`TestExecute_ConfigYAMLApplied` in `cmd/`, a SIGINT-timing test — confirmed failing identically on `origin/main` before this branch's changes, via a throwaway worktree).

## Test plan

- [x] `go build -o fabrik .`
- [x] `go vet ./...`
- [x] `go test -race ./...` (3843 passed; 1 pre-existing unrelated failure confirmed on `origin/main`)
- [x] Manual Acceptance-3 neutralization check (see above)

### DIFF
--- a/adrs/1458-transient-api-error-exemption.md
+++ b/adrs/1458-transient-api-error-exemption.md
@@ -1,0 +1,166 @@
+# ADR 1458: Transient `api_error` Exit Exemption from `max_retries`
+
+**Date**: 2026-08-09
+**Status**: Accepted
+**Issue**: #1458 — A transient `api_error` exit is charged against `max_retries` and pauses the issue, though the stage never ran
+
+## Context
+
+A Claude invocation that exits with `terminal_reason="api_error"` never ran the stage — the CLI's own
+result object records 0-1 turns and `$0.0000` cost — yet before this change it was charged against
+`max_retries` exactly like a genuine stage failure, and three of them paused the issue. Observed live on
+2026-08-08, 21:19–21:32 UTC: eight such exits across two issues (#1408, #1208), every one at 1 turn and
+`$0.0000`. #1408 was paused with `stage:Validate:failed` on a PR that was `MERGEABLE`/`CLEAN` with all
+six checks green, and stayed paused for roughly an hour with a dependent chain blocked behind it. No
+Claude invocation ran anywhere on the board for ~50 minutes, because both active lane heads were paused.
+Nothing was wrong with either issue's work — the transient was Anthropic-side.
+
+`engine/claude.go`'s structural classification (ADR-1183) recognized exactly one `terminal_reason`
+value, `"blocking_limit"` (a genuine account usage limit). Everything else, including `api_error`, fell
+through a diagnostic-only log line to the generic error path, reaching `StageRetryIncremented` and, at
+`MaxRetries`, `stage:<name>:failed` + `fabrik:paused`.
+
+This issue is explicitly modeled on ADR-1119's precedent (`StageAttempted`-without-
+`StageRetryIncremented`), which #1119 established for a genuine usage-limit exit and which
+`apiKeyHelperDetectedError`/`handleAPIKeyHelperDetected` (R13) later reused for a second, unrelated
+condition. The ask was not to loosen classification generally — ADR-1183's structural-only discipline
+(never output prose) had to be preserved — but to add one further narrow structural signal from the
+CLI's own result object.
+
+## Decision
+
+### A new, distinct sentinel type — not a reuse of `claudeUsageLimitError`
+
+The naive reading of the issue ("add one further structural `terminal_reason` match alongside the
+existing check") would add `"api_error"` as a second value inside `classifyUsageLimitExit` and route it
+through the existing `claudeUsageLimitError` type. This was rejected: `claudeUsageLimitError` is not an
+inert marker. By Go *type*, via `errors.As`, it is the trigger for two additional behaviors bundled at
+three call sites:
+
+1. Account-wide suspension activation (`activateClaudeSuspension`, in both
+   `runInvocationWithExtension` and `runCommentExtensionLoop`).
+2. Exclusion from the comment-processing circuit breaker's count (`processComments`).
+
+Both are correct only for a genuine account-wide usage limit. An `api_error` is per-invocation and
+transient — it says nothing about the account as a whole. Reusing the type would have silently
+suspended Claude dispatch account-wide on every transient API blip (directly regressing the ADR-1183
+incident this repository already paid for once, in spirit if not in mechanism) and would have exempted
+the comment-triggered dispatch path from its only existing bound (the circuit breaker), converting a
+bounded 3-strike pause into a potentially unbounded fast retry loop — exactly the failure mode
+`api_error`'s own observed shape (#1208: five exits in 31 seconds) demonstrates is reachable.
+
+`claudeAPIErrorExit` is therefore a new sentinel type, structurally parallel to the existing
+`claudeTurnLimitError`/`apiKeyHelperDetectedError` precedents:
+
+```go
+const apiErrorTerminalReason = "api_error"
+
+type claudeAPIErrorExit struct {
+    TerminalReason string
+    NumTurns       int
+    CostUSD        float64
+}
+```
+
+`classifyAPIErrorExit(resp claudeResponse, usage TokenUsage)` mirrors `classifyUsageLimitExit` exactly:
+structural-only (`resp.TerminalReason == "api_error"`, never output prose), reusing the same conjunctive
+`usage.TurnsUsed > 0 && usage.CostUSD > 0` exclusion. `interpretClaudeResult` (`engine/claude.go`) checks
+this immediately after the `classifyUsageLimitExit` branch and before the pre-existing diagnostic-only
+unmatched-`terminal_reason` log line, so an `api_error` exit that fails the guard (real turns and real
+cost) still falls through to that log line unchanged — the same behavior it had before this issue.
+
+Being a distinct type means the two `errors.As(&claudeUsageLimitError{})` checks in `item.go`/
+`comments.go` simply never match a `*claudeAPIErrorExit` — zero changes needed at either call site. This
+is a structural guarantee, not a conditional added at each site that could be forgotten in a future
+edit.
+
+### Handling: a fifth, more minimal "did not run" shape
+
+`finalizeStageOutcome()` (`engine/item.go`) gains one more `errors.As` branch, alongside the existing
+`claudeUsageLimitError`/`apiKeyHelperDetectedError` branches, routing to `handleAPIErrorExit`. It applies
+`itemstate.StageAttempted` (so the normal dispatch cooldown applies) but never `StageRetryIncremented` —
+the same split every prior "did not run" handler uses. Unlike `handleUsageLimitExit` and
+`handleAPIKeyHelperDetected`, it applies **no label and posts no comment** — R4's explicit requirement,
+and a shape none of the four existing "did not run"/boundary handlers matched exactly. The condition
+self-resolves on the next invocation; a label here would be indistinguishable from the orphaned-durable-
+state leak ADR-1183's own sweep exists to clean up, and a comment for a self-healing, per-invocation
+transient is noise (the same "comment when a human must act, log when the engine self-heals" rule #1414
+established for a sibling condition).
+
+### `engine/comments.go` is deliberately untouched
+
+The comment-triggered dispatch path has no `LastAttemptAt`-based cooldown at all — that mechanism only
+exists in `itemNeedsWork`'s stage-dispatch fallthrough, which the comment path bypasses entirely (a new
+comment dispatches unconditionally once detected). Its only existing bound is the comment-processing
+circuit breaker (`checkCommentBreaker`, #1089, default 10 invocations/30 min). Because
+`claudeAPIErrorExit` is a distinct type from `claudeUsageLimitError`, it automatically falls through to
+the same `checkCommentBreaker(item, "")` call every other unclassified error already reaches — no code
+change was needed, and none was made. This is also the answer to R6 for the comment-triggered path:
+unlike a genuine usage-limit hit (deliberately exempt from the breaker, since it says nothing about a
+specific issue's comment thread), an `api_error` should count toward "no forward progress" accounting —
+counting it is a feature, not an oversight, cheap insurance against an unusually long-lived transient. It
+is the direct explanation for #1208's 31-second/5-invocation burst: comment-triggered redispatch, not
+the 5-minute-cadence stage-dispatch cooldown that bounds #1408's pattern.
+
+## Rationale
+
+### Why reuse `classifyUsageLimitExit`'s exact exclusion guard despite unverified `CostUSD`?
+
+R5 asked whether the observed 1-turn/`$0.0000` `api_error` exits' `CostUSD` is exactly zero or a small
+value the `%.4f` log format rounds down — this could not be empirically verified: no raw NDJSON/result-
+object sample of a real `api_error` exit exists anywhere in this repository (`grep -rl "api_error"`
+across all `.go` files returned nothing prior to this change). The guard is reused unchanged rather than
+dropped, because it is directionally safe regardless of the true value: if a future `api_error` exit
+follows genuine billed work (many turns, real cost), the guard excludes it and it falls through
+unchanged to the pre-existing generic-failure path — no regression, no silent over-classification of a
+real failure. If cost is genuinely zero, as the observed shape strongly suggests, the guard passes and
+the fix activates. Dropping the guard entirely was considered and rejected: it would remove the belt-
+and-suspenders protection that is the direct lesson of this codebase's own #1184/#1183 history, for no
+evidenced benefit. The pre-existing diagnostic-only log line remains reachable for an `api_error` that
+fails the guard, so a future occurrence stays observable and the guard can be revisited with real
+captured data.
+
+### Why not a config flag or a shared "did-not-run" umbrella type instead of a new struct?
+
+A single umbrella type covering usage-limit, apiKeyHelper, and api_error conditions was considered and
+rejected: the three conditions differ in exactly the two behaviors (suspension, breaker exemption) this
+decision hinges on keeping *out* of the `api_error` path. A shared type would need a discriminant field
+and conditional logic at every call site that currently just checks a type — reintroducing the coupling
+risk this ADR's whole design avoids. Small, distinct sentinel types matching the existing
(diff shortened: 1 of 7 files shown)


arbeithand:
*branch: fabrik/issue-1458 | commit: bc834641 | main: 4012b38ca4e470b485b577e90dfbcb564ae4d7c8 | 2026-08-09 00:34 UTC*

All tasks complete. Now let's emit the PR creation marker to finish the stage.


Added a distinct `claudeAPIErrorExit` sentinel type and classifier so a transient `api_error` exit is recorded as `StageAttempted` without `StageRetryIncremented` (no label/comment, unlike the usage-limit precedent), leaving `engine/comments.go` untouched since the new type never triggers account-wide suspension or breaker bypass. Added unit + item-level tests, verified Acceptance 3 by neutralizing the fix, updated `docs/state-machine.md`/`llms-full.txt`, and wrote ADR 1458. Full build/vet/race-test suite is green except one pre-existing, unrelated `cmd` package flake confirmed on `origin/main`.

---
Used 127/100 turns, 14.9M input (207 raw + 14.7M cache-read + 197k cache-write) / 43k output tokens.

arbeithand:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-1458 | commit: bc834641 | main: 4012b38ca4e470b485b577e90dfbcb564ae4d7c8 | 2026-08-09 00:41 UTC*

Reviewed the `api_error` exemption implementation end-to-end: `engine/claude.go` (new `claudeAPIErrorExit` sentinel + `classifyAPIErrorExit`, correctly ordered in `interpretClaudeResult`'s cascade), `engine/item.go` (`handleAPIErrorExit`, properly gated as a distinct type so it never triggers account-wide suspension or the comment-breaker bypass), tests, docs, and ADR 1458. Full build/vet/race suite is green (`engine` package: all pass); the only failure (`cmd/TestExecute_ConfigYAMLApplied`) reproduces identically on `origin/main` in this sandbox due to GitHub API rate-limiting, confirmed pre-existing and unrelated. `docs/llms-full.txt` regeneration matches committed output exactly. Pruefer's bot review already found no issues. No fixes needed — PR is ready for human review.

---
Used 42/100 turns, 4.4M input (81 raw + 4.3M cache-read + 115k cache-write) / 7k output tokens.

arbeithand:
*branch: fabrik/issue-1458 | commit: bc834641 | main: 4012b38ca4e470b485b577e90dfbcb564ae4d7c8 | 2026-08-09 00:48 UTC*

Local branch matches remote exactly, working tree clean, PR is `MERGEABLE`/`CLEAN`. All checks pass.

## Validation Report

### Requirements: 7/7 Acceptance Criteria passed
- [x] **AC1** — `TestAPIErrorExit_ExemptFromRetryNoLabelNoComment` asserts `snap.Attempts("Research") == 0` directly (the counter, not label absence, as required) plus zero labels/comments
- [x] **AC2** — `TestAPIErrorExit_ThreeConsecutiveThenSuccess_CompletesNormally`: three `api_error` exits, no `stage:Research:failed`/`fabrik:paused`, 4th invocation completes normally end-to-end (the #1408 scenario)
- [x] **AC3** — Verified empirically: neutralized the `item.go` routing branch (`if false && errors.As(...)`), confirmed both AC1/AC2 tests go red, restored the fix, confirmed clean diff and tests green again
- [x] **AC4** — `TestAPIErrorExit_DoesNotActivateAccountWideSuspension` asserts `claudeSuspendedUntilTime` unaffected
- [x] **AC5** — `TestGenuineStageFailure_StillCountsAgainstRetries` (existing) still passes, confirming a plain error with no `terminal_reason` is unaffected
- [x] **AC6** — `TestInterpretClaudeResult_APIErrorExit_UnparseableJSON_NotClassified` passes
- [x] **AC7** — `go test -race ./...` green (see below)

### Test Suite: PASSED
- 3838 passed, 6 failed (all confirmed pre-existing/environmental, not caused by this PR — see Regressions below), 1 skipped, across 20 packages
- Race detector: clean
- Build (`go build -o fabrik .`) / `go vet ./...`: clean

All 6 failures verified pre-existing by running the identical tests against a fresh `origin/main` worktree and/or re-running in isolation:
- `cmd.TestExecute_ConfigYAMLApplied`, `cmd.TestExecute_ConfigTUIFalse` — fail identically on `origin/main` (sandbox has no live GitHub network access)
- `engine.TestInvokeClaude_FabrikEnvVarsInjected*` — fails identically on `origin/main` (local `NVM_INC` env var pollution)
- `engine.TestInvokeClaude_MaxWallTimeKillsWithComplete`, `engine.TestKillProcGroupGraceful_StructuredLog` — flaky under full-suite load; pass cleanly 3/3 in isolation; not in this PR's diff

### Design review
Traced the full implementation against Research/Plan's design: `claudeAPIErrorExit` is a genuinely distinct sentinel type (not a reuse of `claudeUsageLimitError`), so it structurally never triggers `activateClaudeSuspension` or the comment-breaker bypass — verified both `errors.As(&limitErr)` call sites (`item.go:1114`, `comments.go:376,465`) type against `*claudeUsageLimitError` only. `engine/comments.go` is confirmed untouched (empty diff). `handleAPIErrorExit` is log-only with `StageAttempted` (never `StageRetryIncremented`), no label, no comment — matching R2/R4 exactly.

### Documentation
`docs/state-machine.md` §7.3a accurately describes the as-built behavior; `docs/llms-full.txt` regeneration produces zero diff (up to date). ADR 1458 correctly numbered per convention, thorough and honest about trade-offs (e.g. no retry ceiling on the stage-dispatch path, mirroring ADR-1119's existing precedent).

### Verdict: READY TO MERGE

Validation passed. PR mergeable: MERGEABLE, mergeStateStatus: CLEAN. All 7 acceptance criteria verified (AC3 confirmed empirically by neutralizing the fix and observing test failures), full race-enabled test suite green apart from 6 pre-existing/environmental failures unrelated to this change, docs and ADR in sync.

---
Used 40/100 turns, 3.9M input (77 raw + 3.8M cache-read + 96k cache-write) / 11k output tokens.

arbeithand:
🏭 **Fabrik merge-train** — Landed via batch PR #1471.

arbeithand:
🏭 **Fabrik merge-train** — Landed via batch PR #1471.

Issue #1383, opened 2026-08-03
A resumed session that overflows the model's context is reported as "model may not exist or you may not have access to it"


bdueck:
v0.0.75. Companion to #1382. The diagnostic cost here was high because the error names the wrong thing entirely.

## Symptom

Every invocation for one issue failed instantly — `1 turn, $0.0000, 0k input / 0k output` — with:

```
There's an issue with the selected model (~deepseek/deepseek-v4-flash-latest).
It may not exist or you may not have access to it. Run --model to pick a different model.
```

That message says *access/existence*. Both were fine: **three other issues completed stages successfully on the exact same model string, concurrently, in the same window.**

## Actual cause

The issue's `Implement.session` pointed at a session that had grown to **3.0 MB ≈ 755k tokens** (from an earlier 85-turn run that failed). Resuming that session exceeds the configured model's context window; the provider rejects it, and that rejection surfaces as the model-not-found message above.

Isolated with a two-line reproduction — same session, same worktree, seconds apart:

```bash
# fails with "model may not exist"
claude -p "Reply with exactly: OK" --resume <session-id> --model ~deepseek/deepseek-v4-flash-latest

# succeeds, prints OK
claude -p "Reply with exactly: OK" --resume <session-id>
```

Same session file in both. So it is neither a corrupt session nor an unavailable model — it is **this session against this model's context limit**.

## Why it compounds

`resume := !lastAttempt.IsZero()` means every retry after a failure resumes, and each failed attempt **appends** to the same session file (`"Continue from where you left off."` → a `<synthetic>` assistant error → the next stage prompt appended). So the session grows monotonically while failing, and an overflow can never recover on its own — the retry makes the overflow worse. Combined with the unbounded comment-review retry in #1382, ours grew across 141 attempts.

## Suggestions

1. **Report provider context-limit rejections as such.** Distinguishing "model not found / no access" from "request too large for this model's context" would have saved hours; the current text actively points away from the cause.
2. **Consider a size guard before `--resume`.** If the stored session exceeds a threshold for the configured model, start cold rather than resuming into a guaranteed rejection — arguably better than failing repeatedly, since the branch state is on disk anyway.
3. Relatedly, a resumed run is invisible to a reader — there is no log line distinguishing a resume from a cold start, which made this harder to spot.

## Workaround for anyone hitting this

[...]

Issue #1199, opened 2026-07-27
Turn-cap preemption counts against max_retries, pausing jobs that are progressing normally


arbeithand:
## Problem

A turn-budget exhaustion increments `max_retries`, so a job too large to finish within `max_retries` slices is **paused while progressing normally** — labelled `stage:<X>:failed`, commented "failed to complete after 3 attempt(s)", and left for a human to unpause.

`engine/item.go:1393`:

```go
if claudeRan && e.cfg.MaxRetries > 0 {
    e.store.Apply(itemstate.StageRetryIncremented{...})
}
```

`claudeRan` is true for a turn cap (only `exec.Error` / `os.PathError` set it false), so every slice counts as a failed attempt.

## The turn budget is not a failure threshold

It exists for two reasons, neither of which is "the work went wrong":

1. **Runaway guard** — bound an unbounded Claude loop.
2. **Time-slicer** — chop a large job so one issue does not starve every other issue of workers.

When the cap is hit, the Claude session is **resumed** on the next dispatch and the worktree persists. Work continues from where it stopped. A capped slice is a *preemption point*, not an error — the engine's own design relies on this, which is why `--resume` and WIP-commit-and-push exist on that path.

Conflating a preemption with a failure is a category error, and it has concrete costs.

## Observed 2026-07-27 — three issues, same day

| Issue | Slices | Outcome |
|---|---|---|
| #816 | `51/50` → `12/50` → `3/50` | paused; later completed at `41/100` |
| #1114 | `51/50` ×5 | paused; later completed at `47/100` |
| #1183 | `51/50` ×3, cost `$3.65 → $5.03 → $6.53` | paused; still outstanding |

#1183 is the clearest illustration: **cost rises with each slice**, which is exactly what accumulating progress looks like — each resumed session carries more context and does more work before the next preemption. Nothing failed. It was paused for being large.

Each pause required a human to notice and intervene. `fabrik:extend-turns` "fixes" it only by making slices big enough to finish inside the retry budget — treating the symptom, and re-introducing the unbounded-work risk the cap exists to prevent.

## `max_retries` is overloaded — it counts two different things

**Correcting the framing of this issue.** As originally written it led with "a turn-capped invocation must not increment `max_retries`" alone, which is wrong on its own — removing the only bound on slicing would let a non-converging job slice forever. The real defect is narrower and more precise.

[...]

PR #1487, opened 2026-08-09
Abandon a session after consecutive `--resume` failures (#1414)

### LINKED ISSUE #1414: A failing --resume session is retried indefinitely with no reset path

arbeithand:
Closes #1414

## What this does

A `--resume` invocation resumes the same session every time, including when the session itself (an oversized or otherwise structurally broken transcript) is the cause of the failure — every retry re-triggers the identical condition and appends to the transcript that's already the problem. This PR adds a durable, bounded counter that abandons a poisoned session pointer and cold-starts the next invocation, closing the indefinite-retry loop reported in #1383/#1386.

## Key changes

- **`engine/claude.go`**: new `claudeResumeFailureError` sentinel (returned only from `interpretClaudeResult`'s generic-failure fallthrough, once stale-session/turn-cap/usage-limit/`api_error` have all declined to classify the exit); a durable plain-text sidecar counter (`<Stage>.session.resumefails`, colocated with the session pointer) with the standard reset/increment/abandon table; `resumeSessionID`/`maxResumeFailures` threaded through `runClaude` → `interpretClaudeResult`.
- **Storage is a file, not `itemstate.Store`** — deliberate divergence from every sibling counter (`SliceRetries`, `ReviewCycles`, etc.), all of which are confirmed in-memory-only. This is the one counter that must survive an engine restart.
- **`engine/item.go`**: `finalizeStageOutcome` exempts every consecutive resume failure up to `MaxResumeFailures` from `StageRetryIncremented` (not only the abandoning one) — `StageAttempted` still fires, so the dispatch cooldown applies and it can't hot-loop, but the guaranteed cold-start attempt can't be starved by `max_retries` either.
- **`engine/comments.go`**: the comment-review path shares the exact same session file and counter as the stage path (no separate counter — an alternating stage/comment-review failure sequence still reaches the threshold), and deliberately still counts a resume failure toward the comment-processing circuit breaker (mirrors `api_error`, not usage-limit).
- **`engine/merge_train.go`**: the holding-stage conflict-resolution invocation gets the fix for free by threading the config value through.
- **No new "force cold start" plumbing**: removing the session file at threshold is itself the cold-start signal for whichever invocation path runs next — both callers already treat an absent file as a fresh session.
- **New `MaxResumeFailures` config** (`--max-resume-failures` / `FABRIK_MAX_RESUME_FAILURES`, default 2 — lower than sibling counters' 3–5, since a resume retry is provably identical each time).
- No issue comment, no durable label for the abandonment itself — only a log line (session id, stage, consecutive-failure count, threshold, last error). Self-healing by construction; if the cold-started attempt also fails, that's a genuine failure and flows into the normal `max_retries` → `fabrik:paused` path, which already comments.

## Docs

`docs/USER_GUIDE.md` (flag/env rows + troubleshooting note), `docs/state-machine.md` (§2.17), `docs/stage-lifecycle.md` (Session Resume subsection), `adrs/1414-resume-failure-session-abandonment.md` (storage divergence, full classification table, explicit "#1199 doesn't fit" note, residual risks), `docs/llms-full.txt` regenerated.

## Testing

- `engine/claude_resume_failure_test.go`: unit coverage of the full classification table at the `interpretClaudeResult` level, including the alternating-stage/comment-review-path sequence.
- `engine/resume_failure_retry_test.go`: `processItem`-level integration test confirming the `max_retries` exemption and that a subsequent genuine (non-resume) failure counts normally.
- `go build ./...`, `go vet ./...`, `go test -race ./...` all clean. The only test-suite failure observed (`TestExecute_ConfigYAMLApplied`) is a pre-existing, network-dependent flake unrelated to this change (fails identically with or without this branch's commits, due to no GraphQL network access in this sandbox).

## Known residual risks (documented in the ADR, not fixed here)

- A future poisoned-session failure that happens to present as `terminal_reason == "api_error"` would be intercepted by the already-shipped ADR-1458 classifier before this mechanism ever sees it (no real sample exists yet to confirm or rule this out).
- The session janitor's exact `.session`-suffix match won't prune an orphaned `.resumefails` sidecar after its parent session ages out — a few bytes, out of scope for this issue (ADR-1136).

Per the issue's own instruction, no closing keywords for #1383 or #1386 are used here — that community report is closed manually once a release ships.


arbeithand:
*branch: fabrik/issue-1414 | commit: b5eafb96 | main: 30116f3feea121ba9188474e8caaa80a5ca90ac7 | 2026-08-09 03:49 UTC*

All work is implemented, tested, documented, and pushed. Emitting the PR creation marker now.


Implemented the consecutive-resume-failure session-abandonment mechanism per the Plan: a durable file-based counter shared by both invocation paths (and merge-train), a new `claudeResumeFailureError` sentinel exempted from `max_retries` up to `MaxResumeFailures` (default 2, configurable), abandonment via the existing session-file `os.Remove` with no new "force cold start" plumbing, and log-only reporting (no comment/label). Added unit and integration tests covering the full classification table and the alternating stage/comment-review sequence, updated USER_GUIDE.md/state-machine.md/stage-lifecycle.md, wrote ADR-1414, and verified `go build`/`go vet`/`go test -race` are clean (aside from one pre-existing, network-dependent flaky test unrelated to this change). All 19 Plan tasks are checked off and the branch is pushed.

---
Used 3/100 turns, 1.2M input (5 raw + 785k cache-read + 381k cache-write) / 2k output tokens.

arbeithand:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-1414 | commit: 9a7c1233 | main: 30116f3feea121ba9188474e8caaa80a5ca90ac7 | 2026-08-09 03:56 UTC*

## Review Findings

### Verified

- **Rebase**: branch already based on latest `origin/main` (30116f3f); pushed (force-with-lease, same content with fresh commit hashes from an earlier rebase in this worktree).
- **External feedback**: no reviews or bot comments on PR #1487 to address.
- **Correctness**: traced the full classification flow in `interpretClaudeResult`/`classifyResumeFailure` against the spec's per-outcome table (success/completed-despite-error/turn-cap → reset; usage-limit/api_error → untouched; generic-with-resume → increment/abandon; generic-cold-start → reset). Matches exactly. Confirmed both `InvokeClaude` and `InvokeClaudeForComments` thread `resumeSessionID`/`opts.MaxResumeFailures` into the shared `runClaude`/`interpretClaudeResult` funnel, and `merge_train.go`'s conflict-resolution path inherits it via `InvokeOptions`. Confirmed `resolveResumeSessionID`'s existing absent-file handling is sufficient as the cold-start signal for both callers — no separate override needed, as the Plan/ADR argue.
- **`MaxRetries` independence**: `finalizeStageOutcome` computes `resumeFailed` and skips `StageRetryIncremented` for every resume failure up to threshold (not just the abandoning one), while `StageAttempted` still fires unconditionally via the existing `claudeRan` gate — verified by direct read of `engine/item.go` and by `TestResumeFailure_ExemptFromMaxRetries_UntilColdStart`, which confirms `Attempts` stays 0 across resume failures and then increments once a genuine (non-resume) failure occurs post-abandonment.
- **Comment-path circuit breaker**: `engine/comments.go` deliberately does *not* exempt resume failures from `checkCommentBreaker`, mirroring the `api_error` precedent (ADR-1458) rather than the usage-limit exclusion — matches Plan's stated decision.
- **Durability**: counter is a file sidecar (`<Stage>.session.resumefails`), not `itemstate` (confirmed in-memory) — satisfies the restart-durability requirement.
- **Docs**: `docs/USER_GUIDE.md`, `docs/state-machine.md` §2.17, `docs/stage-lifecycle.md`, and ADR-1414 all accurately describe the shipped behavior. Regenerated `docs/llms-full.txt` locally and diffed against the committed version — no drift.
- **Build/vet/tests**: `go build ./...` and `go vet ./...` clean. `go test -race -timeout 5m ./...` passed 3949/3953, with the 4 failures confirmed pre-existing/environmental and unrelated to this change: `TestExecute_ConfigYAMLApplied` and `TestInvokeClaude_FabrikEnvVarsInjected` fail identically when run against an unmodified base-branch checkout (network/GraphQL and `NVM_INC` env-leak issues); `TestKillProcGroupGraceful_StructuredLog` passes reliably in isolation on this branch (a load-timing flake under full-suite `-race`). All resume-failure-specific tests (35 across `engine`/`cmd`) pass cleanly in isolation.
- **Test coverage**: the alternating stage/comment-review sequence (`TestClassifyResumeFailure_AlternatingPaths_StillReachesThreshold`) explicitly covers the issue's core new-finding requirement — a single counter reached from either invocation path.

### Not fixed (non-blocking, already named as accepted risk in the ADR)

- A theoretical edge case: if the stale-session detection (`"No conversation found with session ID"`) were ever accompanied by a non-nil `runErr` (unlike every captured production sample, where `runErr` is `nil` for this response shape — confirmed via the existing `TestInterpretClaudeResult_StaleSession_DeletesFileAndDoesNotResave` test), execution would fall through past the stale-session `os.Remove` into `classifyResumeFailure`, which would harmlessly re-attempt the same removal and increment/reset an already-orphaned sidecar. No observable behavior change results (the session is gone either way), and it isn't reachable with current production payload shapes, so not worth chasing further here.

No blocking issues found; nothing required a code fix. The implementation matches the Research/Plan design precisely, the classification table and its edge cases are thoroughly tested, documentation and the ADR are accurate and complete, and the branch builds/tests cleanly. PR #1487 is ready for human review.

Reviewed the resume-failure session-abandonment implementation against the Plan and spec: verified the classification table, the shared-counter/alternating-path behavior, the `MaxRetries` exemption scoping, comment-breaker interaction, and all documentation (USER_GUIDE, state-machine, stage-lifecycle, ADR-1414, llms-full.txt). No issues found — build/vet clean, all resume-failure tests pass, and the 4 full-suite `-race` failures were confirmed pre-existing/environmental (verified against an unmodified base checkout) rather than caused by this change. Rebased onto latest main and pushed. PR #1487 is ready for human review.

---
Used 52/100 turns, 6.7M input (101 raw + 6.5M cache-read + 159k cache-write) / 16k output tokens.

arbeithand:
*branch: fabrik/issue-1414 | commit: 9a7c1233 | main: 30116f3feea121ba9188474e8caaa80a5ca90ac7 | 2026-08-09 04:07 UTC*

## Validation Report

### Requirements: 10/10 verified

- [x] **Shared counter, both invocation paths, alternating sequence** — `classifyResumeFailure` keys purely on `sessFilePath`, reached identically from `InvokeClaude`, `InvokeClaudeForComments`, and `merge_train.go`'s conflict-resolution call. Verified directly in `engine/claude.go:1384-1409` and by `TestClassifyResumeFailure_AlternatingPaths_StillReachesThreshold` (passes).
- [x] **Cold-start at threshold, both paths** — no new "force cold start" boolean; `os.Remove(sessFilePath)` on abandonment is the signal, and `resolveResumeSessionID` already treats an absent file as cold-start for both callers identically (`engine/claude.go:600-618`, `647`, `684`).
- [x] **`MaxResumeFailures` config field, default 2** — `engine.Config.MaxResumeFailures`, `cmd.Config.MaxResumeFailures`, `--max-resume-failures` flag, `resolveInt(..., "FABRIK_MAX_RESUME_FAILURES", "", 2)` — mirrors `--max-rebase-cycles` exactly (`cmd/root.go:47,157,337,762`).
- [x] **Consecutive, not windowed** — plain increment/reset, no time-based window logic anywhere in `classifyResumeFailure`.
- [x] **Log-only abandonment, no comment/label** — `claudeLog(..., "resume", ...)` names session id, count, threshold, last error, and states branch work is unaffected (`engine/claude.go:1396`); confirmed no comment-posting or label call in the resume-failure path in `item.go`.
- [x] **Committed work untouched** — cold start only removes the session-pointer file; `commitWIP`/push run unconditionally for a resume failure (non-short-circuiting, `claudeTurnLimitError`-shaped, per `item.go:1258-1272`).
- [x] **Stale-session path unregressed** — the existing string-match `os.Remove` (`claude.go:1440-1448`) runs before and independently of the new classifier; `classifyResumeFailure` is only reached via the generic fallthrough.
- [x] **Usage-limit exit excluded** — `classifyUsageLimitExit`/`classifyAPIErrorExit` both return before reaching `classifyResumeFailure` (`claude.go:1519-1533`); verified by `TestClassifyResumeFailure_UsageLimitExit_LeavesCountUntouched` / `_APIErrorExit_LeavesCountUntouched`.
- [x] **Reset on success** — three reset points confirmed (success/`runErr==nil`, completed-despite-error, turn-cap exit) at `claude.go:1495,1511,1545`.
- [x] **Restart-durable** — plain-text sidecar file (`<sessFilePath>.resumefails`) via `os.MkdirAll(0700)`/`os.WriteFile(0600)`, not `itemstate` (confirmed in-memory-only). A deliberate, ADR-documented divergence from sibling counters.
- [x] **Independent of `MaxRetries`** — `StageAttempted` always fires; `StageRetryIncremented` skipped for every resume failure up to threshold, not just the abandoning one (`item.go:1587-1597`); verified by `TestResumeFailure_ExemptFromMaxRetries_UntilColdStart`.
- [x] **#1199 coordination** — ADR explicitly states `SliceRetries`' shape doesn't fit (wrong storage, wrong dimension) and reuses the narrower `StageAttempted`-without-`StageRetryIncremented` exemption pattern instead.
- [x] **Alternating-path test coverage** — present and passing.

### Test Suite: PASSED (pre-existing failures excluded)

- `go build ./...` — clean
- `go vet ./...` — clean
- `go test -race -timeout 5m ./...` — 3950 passed, 3 failed, 1 skipped, 20 packages
- All 11 resume-failure-specific tests (`engine` + `cmd`) pass in isolation, including the alternating-path and `MaxRetries`-exemption integration tests
- 3 failures (`TestExecute_ConfigYAMLApplied`, `TestInvokeClaude_FabrikEnvVarsInjected` + subtest) independently reproduced on an unmodified clone of `origin/main` — confirmed pre-existing/environmental (network-dependent GraphQL fetch timeout; local `NVM_INC` env-var leak), not touched by this PR's diff, and already correctly identified as such by Review

### Docs: In sync

`docs/USER_GUIDE.md`, `docs/state-machine.md` §2.17, `docs/stage-lifecycle.md`, and `adrs/1414-resume-failure-session-abandonment.md` all present and accurate. Regenerated `docs/llms-full.txt` locally — zero diff against committed version.

### Branch state

Rebased onto latest `origin/main` (no-op, already current), all commits pushed, working tree clean.

### Verdict: READY TO MERGE

PR mergeable: **MERGEABLE**, mergeStateStatus: **CLEAN**. All 10 requirements verified against the actual implementation (not just the Review report), test suite green aside from confirmed pre-existing environmental failures, docs in sync, no TODOs/debug leftovers found.

Validation passed. PR mergeable: MERGEABLE, mergeStateStatus: CLEAN. All 10 requirements independently verified against the code (shared counter across both invocation paths, cold-start-at-threshold, MaxResumeFailures config, restart-durable sidecar storage, MaxRetries independence), full test suite green aside from 3 pre-existing failures reproduced identically on unmodified origin/main, docs and ADR in sync with zero llms-full.txt drift.

---
Used 40/100 turns, 3.4M input (57 raw + 3.3M cache-read + 128k cache-write) / 9k output tokens.

arbeithand:
🏭 **Fabrik merge-train** — Landed via batch PR #1488.

Issue #1428, opened 2026-08-08
pruefer: daemon logs go only to a terminal — wire the Logf hook to a log file


arbeithand:
## Problem

Pruefer has no log file. Every diagnostic line the daemon emits goes to stderr,
which in normal operation is a terminal — so the daemon's only record of what it
did exists in one terminal window's scrollback, and is gone when that window
closes.

This is not a configuration oversight by the operator. `pruefer/log.go` declares
a hook whose documented purpose is exactly this, and the hook is never assigned:

```go
// Logf is an optional package-level logger hook, mirroring github.Logf and
// engine's claudeLogf. When set (by Daemon during construction), Pruefer's
// internal log lines are routed here instead of stderr, so they can land in
// a structured log file in Fabrik's style. nil falls back to stderr (e.g. in
// tests).
var Logf func(prNumber int, tag, format string, args ...any)
```

`grep -rn "pruefer.Logf\|Logf =" pruefer/ cmd/pruefer/ tui/` returns nothing.
`pruefer/execute.go` constructs `Daemon` as a plain struct literal and never
assigns `pruefer.Logf`, so it is always nil and `logf` (`pruefer/log.go:17`)
always takes the `fmt.Fprintf(os.Stderr, ...)` branch. The comment describes
intended behavior that was never wired.

The consequence is that Pruefer's failure modes are undiagnosable after the
fact. This was found while investigating a live incident (see #1427):
`verveguy/zusammen#103` had been failing its diff fetch every 120s for nearly
three hours, emitting a `warn` line each cycle, and none of it was retrievable —
the diagnosis had to be reconstructed from GitHub's API rather than from
Pruefer's own record. The current process (pid observed with fd 1 and 2 both on
`/dev/ttys008`) has no file to grep. A stale `.pruefer/pruefer.out.log` exists
only because an earlier run happened to be started under `nohup` with shell
redirection — an operator convention, not a product behavior, and one that
provides no rotation and no timestamps.

Fabrik itself already solves this. `engine/engine.go` holds a `logFile *os.File`
at `.fabrik/fabrik.log` (opened at `engine/poll.go:124`) with a `logMu
sync.Mutex` serializing concurrent writes, and `(e *Engine) logf`
(`engine/engine.go:421-446`) writes timestamped, issue-scoped lines to it on
every call, in both TUI and plain-text modes:

```go
fmt.Fprintf(e.logFile, "%s [%s] %s", ts, tag, msg)
[...]

PR #1489, opened 2026-08-09
fix(ci): stop `mergeable_state=unstable` from clearing the CI advance gate (ADR-1441)

### LINKED ISSUE #1441: CI gate: mergeable_state=unstable clears wait_for_ci, so non-required checks never gate an advance
## Problem

`wait_for_ci: true` does not gate on any check that branch protection does not
list as **required**. On `handarbeit/fabrik` today that means the entire test
suite is not gating: an issue whose PR has a red `Test and vet` advances out of
Validate as if CI were green.

Observed live on **#1428 / PR #1429** on 2026-08-08. `.fabrik/stages/validate.yaml`
sets `wait_for_ci: true`. The gate engaged and then cleared while a completed,
failed CI run sat on the PR's head SHA:

```
18:46:18  CI workflow run created for head 109f4513
18:46:54  CI workflow run CONCLUDED: failure
18:46:56  fabrik:awaiting-ci        labeled     (gate engages, 2s after the failure)
18:47:29  stage:Validate:complete   labeled     (stage marked complete anyway)
18:47:30  fabrik:awaiting-ci        unlabeled   (gate cleared as if green)
```

Not a race with a slow-reporting check, and not a flake — the same lint failed
on three successive heads before this, each run concluding well before the gate
was evaluated:

```
18:30:25  eef02e0a  CI  completed/failure
18:39:33  dac72c47  CI  completed/failure
18:46:54  109f4513  CI  completed/failure
```

## Root cause

The ADR-033 shortcut in `engine/pr_settle.go:141`:

```go
// ADR-033: mergeable_state ∈ {clean, unstable} → branch protection satisfied;
// skip per-check classification to avoid blocking on non-required workflows.
if gh.MergeableStateAccepted(mergeableState) {
    return PRSettleResult{Status: PRMergeReady, ...}
}
```

`PRMergeReady` reaches `engine/ci.go`'s `checkCIGate`, which clears the gate
outright:

```go
case PRMergeReady:
    // ADR-033 shortcut (clean/unstable) or all CI checks green — gate clears.
    e.addCompleteLabelAndRemoveCI(owner, repo, item, stage)
    return false, false, false, false
```

The problem is the treatment of **`unstable`**. GitHub's `mergeable_state` is
`UNSTABLE` precisely when the PR is mergeable but **a non-required check is
failing or pending**. `clean` means everything passed; `unstable` means
something did not. Collapsing the two into one "branch protection satisfied"
verdict means the gate clears *because* a check is red, as long as that check
is not required.

PR #1429 reported `mergeable: MERGEABLE, mergeStateStatus: UNSTABLE` throughout.

The shortcut's stated intent — "avoid blocking on non-required workflows" — is
reasonable, but it silently delegates the entire definition of "CI passed" to
branch protection. On this repo, `main`'s required contexts are only:

```
Analyze (go)
Verify llms-full.txt is up to date
```

So `Test and vet`, `Analyze (actions)`, `CodeQL`, and `Docs drift check` are all
non-gating. The check that runs `go test ./...` and `go vet ./...` — the one
thing Validate most needs to be true — cannot block an advance.

### Why this is worse than it looks

- It is silent. There is no warning at startup that `wait_for_ci: true` is
  configured while the required-check set covers almost nothing, and no log line
  distinguishing "cleared because everything passed" from "cleared because the
  failure wasn't required."
- It affects **every** issue on the board and every repo Fabrik manages, not
  just this one. Any repo with loose or empty branch protection has an inert CI
  gate.
- It composes badly downstream. #1428 advanced into Queued with a red PR, where
  the merge train then ejected it three times and paused it with a diagnosis
  ("resolve the underlying conflict") that pointed away from the real cause —
  see the companion singleton-bisection issue.
- The reason string is already recorded (`mergeable_state="unstable"`), so the
  information needed to tell the two cases apart is present at the decision
  point and simply not acted on.
- **The same predicate exists in two places, and both were wrong.** The
  identical `gh.MergeableStateAccepted(mergeableState)` shortcut also gates
  `engine/merge_train.go`'s `pollForMergeable` (the merge-train *landing*
  step): `else if gh.MergeableStateAccepted(mergeableState) { return true }`.
  ADR-1153 (#1153) already fixed the sibling shortcut in `pollTrainCI` (the
  train's CI-poll step) but explicitly flagged `pollForMergeable`'s copy as a
  "candidate fast-follow" and left it unfixed — and no follow-up issue was
  ever filed for it. That is the observed failure mode of deferring a fix to
  one of two identical call sites: the second one doesn't get picked back up.
  A structurally identical pattern — one fix applied, a sibling left behind,
  later cited as precedent by a *sixth* instance — is what #1460 (filed
  2026-08-09) documents for `pauseForRebaseCycleLimit` and friends. Left
  unfixed, the engine would reach opposite conclusions about the same PR's CI
  state depending on whether the merge train is on.

## Summary

Change the single-PR CI gate (`settlePRMergeState` in `engine/pr_settle.go`,
consumed by `checkCIGate` in `engine/ci.go` — the path Validate and every other
`wait_for_ci: true` stage uses) so that `mergeable_state == "unstable"` no
longer short-circuits straight to "ready." `clean` keeps the existing fast
path. `unstable` falls through to the existing per-check / required-context
classification chain (`classifyCIFromCheckRuns`, `classifyCIFromRequiredContexts`
via `classifyRequiredContexts`), so a concluded failure on a non-required check
is treated as a confirmed CI failure and a still-pending non-required check is
treated as blocked-not-failed rather than either being silently waved through.
The merge decision itself (`MergePR`'s ADR-072 self-gate, `checkMergeabilityGate`)
is a separate, deliberate policy choice from the advance/completion gate and
must be decided and documented explicitly rather than left to fall out
accidentally from the same code change.

**Both known call sites of the `MergeableStateAccepted` shortcut are fixed in
this issue, by one shared mechanism where possible**: `engine/pr_settle.go`'s
`settlePRMergeState` (the single-PR advance gate) and `engine/merge_train.go`'s
`pollForMergeable` (the merge-train landing step). Deferring the second call
site to a follow-up issue was considered and rejected — it is exactly what
ADR-1153 already tried for this same shortcut, and no follow-up issue ever
materialized.

## Requirements

**R1 — Stop treating `unstable` as unconditionally ready for the CI gate.**
`clean` may continue to clear directly. `unstable` must fall through to
per-check classification rather than short-circuit, so a concluded `failure`
on the head SHA is seen. This means the full existing classification chain
`checkCIGate` already runs for other `PRSettleResult` statuses — check-run
classification (`classifyCIFromCheckRuns`) **and** the ADR-933 required-context
classifier (`classifyCIFromRequiredContexts`/`classifyRequiredContexts`, which
today runs ahead of check-run classification and can catch a classic-commit-status
failure check runs alone would never see) — not just a bare call to
`gh.ClassifyCheckRuns`. Note this changes only the **CI gate's** reading; see
R3 before changing the merge path.

**R2 — Distinguish failing from pending under `unstable`.** `UNSTABLE` covers
both "a non-required check failed" and "a non-required check is still running."
These must classify differently: a concluded failure is a confirmed CI failure
(apply `fabrik:awaiting-ci`, dispatch the CI-fix path); a still-pending check is
blocked-not-failed, re-evaluated next poll under the existing `CIWaitTimeout`
backstop. Do not let the timeout path silently become the common case — that
would trade a false pass for a 30-minute stall on every PR with a slow optional
check.

**R3 — Decide the merge path deliberately, and state the decision.** The
shortcut is consumed by more than the CI gate. Blocking a *merge* on a
non-required check is a genuine policy change: it makes Fabrik stricter than
GitHub itself, and an always-failing optional check would wedge merges
permanently. The defensible split is that the **advance** gate (does Validate
complete?) should honor all checks, while the **merge** decision may continue to
defer to branch protection. Whichever way it lands, the reasoning must be
recorded, and the two must not silently diverge in a way that lets an item
advance but never land.

**R4 — Make the degenerate configuration visible.** When a stage sets
`wait_for_ci: true` and the target branch's required-check set is empty or does
not intersect the checks actually running on the PR, log a warning naming the
gap. A gate that cannot gate should say so rather than pass silently. Startup
board validation is the closest existing precedent for this kind of
config-vs-reality check.

**R5 — Do not regress ADR-033's original motivation.** ADR-033 exists because
blocking on every non-required workflow caused real stalls. The fix must not
reintroduce that: optional checks that are *pending* must not block
indefinitely, and checks that are skipped, neutral, or cancelled must not count
as failures. Only a concluded `failure`/`timed_out`/`action_required`
conclusion should block. Re-read ADR-033 before choosing an approach and amend
it with the outcome.

**R6 — Fix `pollForMergeable`'s identical shortcut in this issue, not a
follow-up.** `engine/merge_train.go`'s `pollForMergeable` contains the same
`gh.MergeableStateAccepted(mergeableState) → return true` predicate with no
check-run fallback at all — the merge-train landing counterpart to
`pr_settle.go`'s shortcut. Fix both call sites using **one shared mechanism**
(a common helper/classification path) rather than two parallel
implementations. If a single mechanism genuinely cannot serve both — because
`pollForMergeable` is a synchronous poll loop reaching a landing decision for
an unattended batch, not a one-shot advance gate, and Research finds it needs
different semantics from the settle path — then say so explicitly, with the
reason recorded in the ADR, and split the second call site into its own
follow-up issue filed *in the same change* (not merely promised) rather than
deferred silently. Default to fixing both together; diverge only on
documented evidence. A test covering only `pr_settle.go` while the train path
stays broken does not satisfy this requirement.

## Scope

**In scope**
- `engine/pr_settle.go` — the `MergeableStateAccepted` short-circuit inside
  `settlePRMergeState`
- `engine/ci.go` — `PRMergeReady` handling in `checkCIGate`; `unstable`
  fall-through to the existing check-run/required-context classification chain;
  failing-vs-pending discrimination
- `engine/merge_train.go` — `pollForMergeable`'s identical
  `MergeableStateAccepted` shortcut (R6); NOT `pollTrainCI`, which ADR-1153
  (#1153) already fixed for this defect class
- `github/` — `MergeableStateAccepted` and any callers that conflate the two
  states
- `adrs/033-*` — amend with the corrected semantics and the R3 decision
- `docs/state-machine.md` — §CI gate semantics must match
- Tests

**Out of scope**
- The merge-train singleton-bisection defect (companion issue) — related by this
  incident but a separate subsystem
- Branch-protection configuration on `handarbeit/fabrik` itself. Adding
  `Test and vet` to the required set would mask this bug on this repo while
  leaving every other managed repo exposed; it is a reasonable operational
  follow-up but is **not** the fix and must not be treated as one.
- `review_authority` / the review gate
- `pollTrainCI` (`engine/merge_train.go`) — already fixed for the identical
  defect class by ADR-1153 (#1153); nothing to do here.

## Prior Art / Context

**ADR-1153 already solved this exact problem for the merge-train's CI-poll step
(#1153, 2026-07-27) — this issue is its direct sequel for the single-PR path
and for the merge-train's *landing* step (`pollForMergeable`), which ADR-1153
explicitly declined to fix.** `pollTrainCI` had the identical `mergeable_state
∈ {clean, unstable} → green` shortcut; ADR-1153 changed it to a *gate-only*
signal (still short-circuits `dirty` → red) and made the check-run pass (plus
ADR-933's required-context classifier) load-bearing for green, reachable on
every iteration instead of being dead code whenever `mergeable_state` resolved
first. Directly relevant precedent for this issue's decisions:

- **Strict policy (their R5-equivalent, Decision §4):** any confirmed
  check-run failure blocks, required or not — deliberately not trying to
  distinguish "confirmed non-required failure" from "confirmed
  required-but-unconfigured failure," since for the common **unconfigured**
  case (this repo has no `RequiredStatusContexts` entries) that distinction
  can't be made anyway. Directly informs this issue's R5, and — per R6 — the
  default expectation that `pollForMergeable` becomes strict too rather than
  keeping its current unconditional accept.
- **R3's merge-vs-advance split, and where R6 diverges from it:** ADR-1153
  explicitly did *not* revisit `MergePR`'s own ADR-072 self-gate or
  `checkMergeabilityGate` for the single-PR path — "an unattended batch
  merge... needs a stronger, additional completeness guarantee that the
  single-PR path does not." `pollForMergeable` sits on the *landing/merge*
  side of that split, structurally, yet R6 asks it to be fixed alongside the
  *advance*-side shortcut rather than deferred with R3's "merge may continue
  to defer to branch protection" reasoning. This is a deliberate choice made
  in review, not an oversight: if Research finds `pollForMergeable`'s
  unattended-batch context genuinely needs the more lenient merge-path
  semantics R3 describes for the single-PR case, that is the condition under
  which R6 permits splitting it out — with the reasoning stated explicitly in
  the ADR and a follow-up issue filed in the same change, not merely
  promised.
- **Zero-check-runs handling is unchanged and must stay that way:** ADR-1153
  preserved the ADR-933 zero-check-runs branch exactly, adding one new arm
  (`mergeableAccepted && RequiredContextsSatisfied → green`) for the case where
  there is no check-run footprint at all. `settlePRMergeState` already has this
  same ADR-933 zero-check-runs branch (lines ~165-232) independent of the
  `MergeableStateAccepted` shortcut this issue changes — it must not be
  disturbed.
- **`describeCheckRuns` helper** (added by #1153, `engine/merge_train.go`) logs
  `"name (status-or-conclusion)"` pairs for every terminal CI decision; worth
  reusing rather than duplicating if this issue adds analogous logging for R4,
  and a candidate building block for R6's shared mechanism.

**ADR-933 (#933) already added per-check required-context classification** on
top of the ADR-033 shortcut, but explicitly declined to touch the shortcut
itself ("reopening that decision is explicitly out of scope for #933"). This
issue is the one that reopens it for the single-PR path and for
`pollForMergeable` — mirroring how ADR-1153 was the one that reopened it for
`pollTrainCI`. Practically, this means `checkCIGate` already has more
classification machinery downstream of the shortcut than the original issue
text's line-number citations suggest: `classifyCIFromRequiredContexts` runs
ahead of `classifyCIFromCheckRuns` for `PRMergeUnsettled`/`PRMergeBlocked`
results today, and R1's "fall through to per-check classification" must
preserve that ordering, not just call `gh.ClassifyCheckRuns` directly.

**#1460 (filed 2026-08-09) documents the cost of deferring a fix to one of
several identical call sites**: a single defect (a non-failure pause with no
independent counter reset) reached five instances because each was fixed in
isolation without a sweep for siblings, and one fixed instance was then cited
as *design precedent* for a sixth, unrelated site. This is the direct
motivation for R6's "fix both now" default over "defer the second site,"
given ADR-1153 already tried deferring this exact shortcut's sibling and no
follow-up issue was ever filed.

## Risks / Dependencies

- This issue was blocked on #1410 ("CI gate should bound liveness, not elapsed
  time"), which introduced the liveness-stall-dwell redesign
  (`ciProgressStalledSince`, `ciWaitTimeout`, `ciBackstopTimeout`) that R2's
  "re-evaluated next poll under the existing `CIWaitTimeout` backstop" language
  refers to. #1410 is now closed/merged, so this dependency is resolved.
- **Rebase before starting.** #1410 and #1440 (the merge-train red-singleton
  work) both merged into `main` on 2026-08-08/09 and both touched
  `engine/merge_train.go` (#1410: +24/-16). The `pollForMergeable` code R6
  targets may have moved; confirm line references against current `main`
  before editing.
- Removing the shortcut for `unstable` adds a `FetchCheckRuns` round-trip to
  the common `unstable` case (mirrors ADR-1153's accepted trade-off of one
  additional `FetchCheckRuns` call per poll iteration — "immaterial API cost"
  there; the single-PR path polls far less frequently, so this is smaller
  here, not larger). The same applies to `pollForMergeable` under R6.
- Must not regress `settlePRMergeState`'s existing ADR-933 zero-check-runs
  handling, `hadChecks`/post-push-dwell handling, or R3's (of `engine/ci.go`,
  not this issue's R3) OPEN+BLOCKED-no-checks-ever pause path — none of those
  are downstream of the `MergeableStateAccepted` shortcut this issue touches,
  but the fall-through path this issue adds must not accidentally bypass or
  duplicate them.
- R6's shared-mechanism goal must not force `pollForMergeable`'s synchronous
  poll-loop shape into `settlePRMergeState`'s single-pass settle shape (or vice
  versa) if that distorts either one — sharing the *classification*
  (check-run/required-context verdict) is the goal, not necessarily sharing
  the calling control flow.

## Acceptance criteria

1. A PR whose head SHA carries a concluded `failure` check run and whose
   `mergeable_state` is `unstable` does **not** clear the CI gate:
   `stage:<name>:complete` is not applied, and the confirmed-failure path runs.
   Asserted against `checkCIGate`'s return tuple directly.
2. The same PR with the failing check still `in_progress` instead classifies as
   blocked-not-failed, leaving `fabrik:awaiting-ci` untouched and re-evaluating
   next poll.
3. `mergeable_state == "clean"` still clears immediately, with no added check-run
   round-trip — ADR-033's fast path is preserved for the case it was written for.
4. Skipped, neutral, and cancelled conclusions do not block (R5).
5. The R3 decision is implemented and covered by a test that pins the
   advance-gate and merge-path behaviors independently, so a future change
   cannot make them disagree unnoticed.
6. R4's warning fires for a stage with `wait_for_ci: true` against a branch whose
   required-check set does not cover the running checks, and does not fire when
   coverage is adequate.
7. **`pollForMergeable` rejects landing on a PR whose head SHA carries a
   concluded `failure` check run and whose `mergeable_state` is `unstable`**,
   mirroring criterion 1 for the merge-train landing path — unless R6's escape
   hatch is invoked, in which case this criterion is replaced by the
   follow-up issue's own acceptance criteria and the ADR records why.
8. Every new test is demonstrated non-vacuous: neutralise the fix and show it
   fails, **independently for each of the two call sites** (criterion 1's
   fixture against `engine/pr_settle.go`/`engine/ci.go`, and criterion 7's
   fixture against `engine/merge_train.go`) — a test that only exercises one
   site passing while the other stays broken does not satisfy this criterion.
   Criterion 1 in particular must fail against current `main` — that is the
   regression test for this incident.
9. `go test -race ./...` and `go vet ./...` pass; ADR-033 and
   `docs/state-machine.md` are updated in the same change set.

## Reproduction

Fixed inputs, no live GitHub needed: a PR with `mergeable: true`,
`mergeable_state: "unstable"`, and one check run with
`status: "completed", conclusion: "failure"` on the head SHA, against a stage
with `wait_for_ci: true`. Current `main` clears the gate; it must not.

The live incident is preserved in #1428's timeline and in the workflow-run
history for branch `fabrik/issue-1428` (heads `eef02e0a`, `dac72c47`,
`109f4513`, all `CI completed/failure`).

arbeithand:
Closes #1441

## What this does

Fixes the CI gate so a confirmed CI failure can no longer be silently waved through just because GitHub's `mergeable_state` reads `unstable` instead of `blocked`/`dirty`. `unstable` means "GitHub's required checks passed, but a *non-required* check is failing or still pending" — Fabrik was treating it identically to `clean` ("everything passed"), which meant a red `Test and vet` run could clear `fabrik:awaiting-ci` and let Validate complete as if CI were green (observed live on #1428/#1429).

**Both known instances of the underlying `gh.MergeableStateAccepted({clean, unstable})` shortcut are fixed here, not just one:**

1. **`engine/pr_settle.go`** (`settlePRMergeState`, consumed by `checkCIGate` in `engine/ci.go`) — the single-PR CI advance gate. The shortcut now only fires for `mergeable_state == "clean"`. `unstable` falls through into the existing per-check/required-context classification chain (already correct and already exercised today for `"blocked"`), so a concluded `failure` check run is now treated as a confirmed CI failure, and a still-`in_progress` one is blocked-not-failed.
2. **`engine/merge_train.go`** (`pollForMergeable`, the merge-train *landing* step) — previously had **zero** check-run awareness at all (unlike `pollTrainCI`, its CI-poll sibling, which ADR-1153 already fixed for this same defect class). It now fetches check runs via a new `classifyLandingCI` helper composed from the same shared primitives `pollTrainCI` already uses (`gh.ClassifyCheckRuns`, `e.classifyRequiredContexts`, `describeCheckRuns`), so the engine can no longer reach opposite conclusions about the same PR depending on whether the merge train is enabled.

**Merge decision is deliberately unchanged**: `MergePR`'s ADR-072 self-gate still accepts `{clean, unstable}`, per ADR-072's own operator note that engine-merge gating belongs in branch-protection config, not `MergePR`. Only the *advance* gate (does Validate complete?) tightened — see ADR-1441 for the full R3 reasoning.

**R4**: a new `warnIfCIGateCoverageDegenerate` logs once per repo/stage when `wait_for_ci: true` is set but the operator's opt-in `RequiredStatusContexts` config doesn't cover the checks actually observed on a PR — visibility for a gate that can't meaningfully gate.

## Key files

- `engine/pr_settle.go` — narrowed shortcut
- `engine/ci.go` — R4 warning, doc comment fixes (no functional change to `checkCIGate` itself — it already interpreted `PRSettleResult` correctly regardless of *why* a status was produced)
- `engine/merge_train.go` — new `classifyLandingCI`, rewritten `pollForMergeable` (now uses `FetchPRDetails` instead of `FetchPRMergeableFields`)
- `engine/interfaces.go`, `engine/mocks_test.go` — `FetchPRDetails` added to the testable `GitHubClient` interface
- `github/prs.go` — `MergeableStateAccepted` doc comment tightened to reflect its narrowed caller set
- `adrs/1441-unstable-requires-check-run-classification.md` (new) + `adrs/033-*` (amended) — full decision record
- `docs/state-machine.md` + regenerated `docs/llms-full.txt`

## Testing

- New regression fixtures at both call sites for the exact failure shape (`mergeable_state=unstable` + a concluded `failure` check run), plus pending/dirty/clean/zero-check-runs/required-context-failure siblings.
- **Non-vacuousness verified manually** (not committed, per AC8): temporarily reverted each fix in isolation and confirmed the corresponding new test fails against the pre-fix code, then restored — done for both `pr_settle.go`/`ci.go` and `merge_train.go` independently.
- `go vet ./...` clean; `go test -race ./...` passes for every package except one pre-existing, unrelated flake (`TestExecute_ConfigYAMLApplied` in `cmd/`, a SIGINT-timing test) — confirmed to fail identically against the pre-session commit in an isolated worktree, so it predates this change.

## How to review

Start with `adrs/1441-unstable-requires-check-run-classification.md` for the full reasoning (especially the R3 merge-vs-advance split and why `pollForMergeable` gets a new composition function rather than literally sharing code with `pollTrainCI`), then `engine/pr_settle.go`'s narrowed shortcut and `engine/merge_train.go`'s `classifyLandingCI`/`pollForMergeable` rewrite.


arbeithand:
*branch: fabrik/issue-1441 | commit: 4f00af44 | main: 30116f3feea121ba9188474e8caaa80a5ca90ac7 | 2026-08-09 04:16 UTC*

Clean working tree, all commits pushed. Now let's emit the PR creation marker.


Fixed both instances of the `mergeable_state=unstable` CI-gate shortcut (`settlePRMergeState`/`checkCIGate` and the merge-train's `pollForMergeable`) so a confirmed check-run failure on a non-required check can no longer silently clear the gate — the root cause of #1428/#1429 advancing with red CI. Added a new `classifyLandingCI` classifier shared-by-primitive with the already-fixed `pollTrainCI`, an R4 degenerate-coverage warning, full regression/non-vacuousness test coverage, and a new ADR-1441 plus updated ADR-033/state-machine docs. `go vet` is clean and `go test -race ./...` passes except one pre-existing, unrelated `cmd/` flake confirmed to predate this branch.

---
Used 63/100 turns, 26.4M input (125 raw + 26.0M cache-read + 448k cache-write) / 26k output tokens.

arbeithand:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-1441 | commit: 2a4ad315 | main: 8364069b0802af092fb0dcedb005a07bd290166f | 2026-08-09 04:32 UTC*

Clean working tree, all changes committed and pushed. Everything checks out. Let's produce the summary.

Rebased cleanly onto main (one mechanical `engine.go` field-add conflict), then reviewed the full diff against the plan and issue requirements — code, tests, ADR-1441, ADR-033 amendment, and `docs/state-machine.md` all matched precisely, with `docs/llms-full.txt` confirmed byte-identical to a fresh regen. Verified AC8 non-vacuousness by hand: temporarily reverting each of the two fixed call sites (`pr_settle.go`'s shortcut, `merge_train.go`'s `pollForMergeable`) made their respective regression tests fail as expected. Found and fixed one real bug flagged by Pruefer's review: on a `FetchCheckRuns` error, `pollForMergeable` passed `checkRuns=nil` straight into `classifyLandingCI`, which conflated "fetch failed" with "confirmed zero check runs" and could classify a red PR as `TrainCIGreen` on a transient API error — reintroducing this issue's own defect class. Fixed to mirror `pollTrainCI`'s existing if/else-if/else shape (skip classification on error, retry next poll), added a non-vacuous regression test, and confirmed `go build`, `go vet`, and the full `go test -race` suite are clean (only a pre-existing, unrelated env-var test flake remains, confirmed untouched by this branch). Pushed the rebased branch with the fix.

---
Used 80/100 turns, 11.9M input (155 raw + 11.7M cache-read + 197k cache-write) / 33k output tokens.

arbeithand:
*branch: fabrik/issue-1441 | commit: 2a4ad315 | main: 8364069b0802af092fb0dcedb005a07bd290166f | 2026-08-09 04:33 UTC*

Thread resolved. No issue body changes are needed — this comment addressed an implementation detail already fixed, not a scope/spec change.

This review finding (Pruefer's `FetchCheckRuns` error-handling concern in `pollForMergeable`) was already fixed in the prior Review stage session — commit `2a4ad315` restructures the error path to mirror `pollTrainCI`'s if/else-if/else shape (skip classification on fetch error rather than treating `nil` checkRuns as confirmed-zero), with a non-vacuous regression test (`TestPollForMergeable_FetchCheckRunsError_DoesNotTreatAsGreen`). Verified the fix is present in the working tree, working tree is clean, and the branch is already pushed to origin. Resolved the review thread. No further action or issue body update needed.

---
**Threads addressed:**
- `engine/merge_train.go:2394` — resolved

arbeithand:
*branch: fabrik/issue-1441 | commit: 2a4ad315 | main: 8364069b0802af092fb0dcedb005a07bd290166f | 2026-08-09 04:44 UTC*

## Validation Report

### Requirements: 6/6 passed (R1–R6)

- [x] **R1** — `unstable` no longer short-circuits: `settlePRMergeState`'s shortcut narrowed to `mergeableState == "clean"` only (`engine/pr_settle.go:147`); `unstable` falls through into the existing check-run/required-context classification chain unmodified. Verified against `checkCIGate`'s tuple via `TestCheckCIGate_MergeableStateUnstable_FailedCheckRun_DoesNotClearGate` and at the settle level via `TestSettle_MergeableStateUnstable_FailedCheckRun_ReturnsBlocked`.
- [x] **R2** — Failing vs. pending discriminated: `TestSettle_MergeableStateUnstable_FailedCheckRun_ReturnsBlocked` (→ `PRMergeBlocked`) vs. `TestSettle_MergeableStateUnstable_PendingCheckRun_ReturnsUnsettled` (→ `PRMergeUnsettled`, re-evaluated under the `CIWaitTimeout` backstop).
- [x] **R3** — Merge path deliberately unchanged: `MergePR`'s ADR-072 self-gate still accepts `{clean, unstable}` (`TestMergePR_UnstableAcceptedExplicit`), independently pinned alongside the tightened advance gate — the two are asserted separately so they can't silently diverge, per ADR-1441's explicit decision record.
- [x] **R4** — Degenerate-coverage warning implemented and tested (`TestWarnIfCIGateCoverageDegenerate_{FiresWhenUnconfigured,FiresWhenNoIntersection,DoesNotFireWhenCovered,DoesNotFireOnEmptyCheckRuns,DedupesAcrossCalls}`), gated on `len(settle.CheckRuns) > 0` (no added API cost), deduped in-memory per repo/stage.
- [x] **R5** — Skipped/neutral/cancelled don't block: `TestSettle_MergeableStateUnstable_SkippedChecksOnly_ReturnsReady`.
- [x] **R6** — Both call sites fixed with one shared mechanism (shared primitives + new `classifyLandingCI` composition, documented reasoning for not literally sharing code with the out-of-scope `pollTrainCI`). `pollForMergeable` rewritten to use `FetchPRDetails` + `classifyLandingCI`.

### Acceptance Criteria: 9/9 passed

1. ✅ Confirmed `failure` + `unstable` blocks the gate — verified against `checkCIGate`'s return tuple, and confirmed this **fails on unpatched code** by temporarily reverting the `pr_settle.go` shortcut (regression test went red, then restored).
2. ✅ Same PR with an `in_progress` check → blocked-not-failed, `PRMergeUnsettled`.
3. ✅ `clean` still clears with **zero** added `FetchCheckRuns` calls — asserted directly (`fetchCheckRunsCalled` flag).
4. ✅ Skipped/neutral/cancelled don't block.
5. ✅ R3 pinned by two independent tests (advance-gate vs. merge-path) that can't silently drift apart.
6. ✅ R4 warning fires/doesn't-fire/dedupes correctly.
7. ✅ `pollForMergeable` rejects landing on `unstable` + confirmed failure — confirmed this **fails against the pre-fix unconditional shortcut** by temporarily reintroducing it (regression test went red, then restored).
8. ✅ Non-vacuousness independently demonstrated for both call sites (see above), plus the review-added `TestPollForMergeable_FetchCheckRunsError_DoesNotTreatAsGreen` also failed against the reverted code.
9. ✅ `go build`, `go vet`, and `go test -race ./...` all pass; ADR-1441 (new) + ADR-033 (amended) + `docs/state-machine.md` (6 locations) updated; `docs/llms-full.txt` regenerated and byte-identical to a fresh run.

### Test Suite: PASSED (with 2 pre-existing, unrelated failures)
- 3966+ tests across 20 packages. Race detector: clean. Build: clean. Vet: clean.
- `TestExecute_ConfigYAMLApplied` (cmd) — SIGINT-timing flake; confirmed no `cmd/` files are touched by this branch's diff.
- `TestInvokeClaude_FabrikEnvVarsInjected/FABRIK_PR_absent...` (engine) — confirmed to pass cleanly under `env -i` (clean environment); only fails here because this Validate stage is itself running inside a nested Fabrik-managed worktree whose ambient shell exports `FABRIK_PR=1489`. Confirmed `engine/invoke_claude*.go` is untouched by this branch.
- Both are environmental, not regressions — reconfirms the PR description's own note.

### Branch State
- Rebased onto latest `origin/main` (already up to date). Working tree clean, all commits pushed.
- PR #1489: `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`. All CI checks green (Test and vet, Analyze (go)×2, Analyze (actions), Verify llms-full.txt, CodeQL). No outstanding review requests; Pruefer's prior finding already fixed and thread resolved.

### Verdict: READY TO MERGE

Validation passed. PR mergeable: MERGEABLE, mergeStateStatus: CLEAN. All 6 requirements (R1-R6) and 9 acceptance criteria verified, including non-vacuousness confirmed by hand for both fixed call sites (pr_settle.go and merge_train.go). Full test suite passes (3966+ tests, race-clean, vet-clean) except two pre-existing, unrelated flakes confirmed independent of this branch's diff.

---
Used 68/100 turns, 8.0M input (129 raw + 7.9M cache-read + 149k cache-write) / 20k output tokens.

arbeithand:
🏭 **Fabrik merge-train** — Landed via batch PR #1491.

PR #1490, opened 2026-08-09
Fix cross-repo spawn board-servability and add mid-flight FABRIK_SPAWN_CHILD recognition for Review/Validate

### LINKED ISSUE #1419: Cross-repo sub-issues are stranded: registered on the parent's board, and unassigned
## Problem

Cross-repo sub-issue spawning has two distinct failure modes that produce the same symptom — a silently unreachable child and, worse, a parent that believes it is blocked when it is not:

1. **Plan-stage spawns via `FABRIK_SPAWN_CHILD`** (the original report): the engine parses the spawn declaration and files the child correctly, but for a child in a *different* repo than the parent, it is registered on the **parent's** project board rather than a board serving the child's repo, and is left **unassigned**. If that board is served by a `repo:`-scoped instance, no instance will ever pick the child up, and the parent — which correctly got `fabrik:blocked` — halts indefinitely with no error, no `fabrik:paused`, and no indication anywhere that anything is wrong.
2. **Mid-flight spawns from later stages (confirmed at Review) that bypass the engine entirely**: when a stage agent discovers a blocker mid-flight and has `gh` available, it may call `gh issue create` directly instead of emitting a `FABRIK_SPAWN_CHILD` block — because `FABRIK_SPAWN_CHILD` is currently only recognized as a Plan-stage mechanism, and an agent that discovers a blocker at Review has no sanctioned route to declare it. The engine never observes the spawn, so **none** of its wiring runs: no board registration, no labels, no assignee, and critically **no `blocked_by` edge** on the parent. This is not a registration bug in the engine — it is a gap in which stages can route a spawn through the engine at all.

Failure mode [2] is materially more dangerous than [1]: an unregistered child is at least visibly stuck, and someone eventually notices nothing is happening. A missing `blocked_by` edge is silent in the other direction — the parent has no record that it is blocked on anything, so it resumes and proceeds through Review and Validate reporting green, potentially merging without the work it was supposed to be blocked on. This was observed directly: `repo-b#102` proceeded to Review on a tree still pinned to the pre-fix dependency, every stage reported green, and the regression was caught only because a human diffed the vendored artifact by hand.

Two live reproductions (detailed under Prior Art below) pin down both modes independently, and a follow-up correction from the reporter re-scoped the second: what first looked like a second instance of the board-misrouting bug was, on timestamp analysis, actually a stage agent calling `gh issue create` directly and never invoking the engine's spawn machinery at all — a distinct gap, not a repeat of failure mode [1].

**Version:** v0.0.77

## Summary

Give every stage that can discover a mid-flight blocker (at minimum Review and Validate) a sanctioned way to declare a spawned child via `FABRIK_SPAWN_CHILD` instead of shelling out to `gh issue create` directly, and make the engine's handling of any declared spawn robust regardless of origin stage: the target board is resolved from the **child's** repo, the child is assigned, a `blocked_by` edge is created on the parent, and any failure in that wiring surfaces visibly on the issue rather than being silently swallowed. A stage resuming from `fabrik:blocked` must also re-verify its dependencies are actually satisfied, not just that the label was cleared — this is what would have caught the observed regression, where a parent resumed and reported green on a tree still missing its blocking dependency's fix.

## Requirements

- A `FABRIK_SPAWN_CHILD` declaration must be recognized and processed by the engine from stages beyond Plan — at minimum Review and Validate, the stages that can discover a blocker mid-flight — so an agent that finds a blocker has a sanctioned route that doesn't require calling `gh issue create` directly.
- Spawned children must receive the same registration wiring regardless of which stage originates the spawn: board registration, default labeling (`fabrik:sub-issue` plus whatever `fabrik:yolo`/`fabrik:cruise` behavior the existing Plan-stage spawn path already applies), assignment, and a `blocked_by` edge — a mid-flight spawn from Review/Validate must not be a second-class path relative to a Plan-stage spawn.
- For a cross-repo spawned child, the target project board must be resolved from the **child's** repository, not the parent's; if no board serves that repo, the spawn must fail loudly (visible label and/or stage failure) rather than silently defaulting to the parent's board.
- Every spawned child must be assigned to the `user:` of the instance meant to process it.
- A `blocked_by` edge must be created for every declared spawned blocker; if edge creation fails, the stage must fail rather than logging and continuing.
- A stage resuming an issue that was `fabrik:blocked` must re-verify its dependency set is actually satisfied (not merely that the label was cleared) before proceeding.
- Any failure in board registration, assignment, or `blocked_by` edge creation must surface visibly on the issue (a label and/or a failed stage) — never only in the daemon log.

## Scope

**In scope:**
- Extending `FABRIK_SPAWN_CHILD` recognition to at least the Review and Validate stages (including the corresponding stage prompt/skill steering needed to make agents actually prefer the declaration route over `gh issue create`).
- Cross-repo board resolution and mandatory assignment for spawned children, regardless of which stage spawns them.
- Guaranteed `blocked_by` edge creation for declared blockers, with failure surfaced rather than swallowed.
- Re-verification of dependency satisfaction when an issue resumes from `fabrik:blocked`.

**Out of scope:**
- General auto-discovery of "which board serves which repo" beyond what is already declared in instance config (`project:`/`repo:`).
- Restricting stage agents' tool access (e.g. removing `gh` from Review/Validate) as a means of preventing direct `gh issue create` calls — the fix is a sanctioned route, not a removed capability.
- Extending `FABRIK_SPAWN_CHILD` recognition to stages beyond Review and Validate — no confirmed mid-flight-discovery case has been observed outside those two; broaden later if one turns up.
- An engine-side reconciliation backstop that parses stage comment prose for blocker mentions and verifies/creates a `blocked_by` edge independent of whether a `FABRIK_SPAWN_CHILD` block was present. This was raised as a possible defense-in-depth measure, but the reporter's own follow-up frames it as heuristic and error-prone (false positives on unrelated issue mentions, false negatives on unmatched phrasing) rather than the primary fix. Worth a follow-up issue if the sanctioned-route fix in this issue turns out not to be sufficient on its own.

## Prior Art / Context

**Setup common to both reproductions:** two instances, two boards, two repos — instance A (`project: 5`, `repo:` unset) and instance B (`project: 9`, `repo: repo-b`), same `user:` on both. Instance A processes multiple repos on board 5 without issue, which is why the reverse direction was expected to work too.

**Reproduction 1 (Plan stage, original report — failure mode [1]):** parent issue in repo-b reached Plan, which spawned four prerequisites into repo-a via `FABRIK_SPAWN_CHILD`. All four children were created with `fabrik:sub-issue`, `fabrik:yolo`, `fabrik:cruise`, added to board 9 (the parent's board, not repo-a's board 5), and left unassigned. Parent got `fabrik:children-spawned` + `fabrik:blocked` and sat idle for ~13 hours until a human manually assigned and re-boarded the children, after which each started within ~79 seconds. Board status was confirmed not to be the cause — the children sat correctly in the Specify column on board 9 the whole time; it was purely the `repo:` filter and missing assignment. This batch *did* get `blocked_by` edges on the parent (`#951`–`#954` are all linked) — the board/assignment misrouting was the only defect in this path.

**Reproduction 2 (Review stage, follow-up — failure mode [2]):** parent `repo-b#102` (board 9), children `repo-a#961` and `repo-a#965`. Neither child had any labels, an assignee, a board-9 item, a board-5 item, or a `blocked_by` edge on the parent. Timestamp analysis distinguishes this from reproduction 1: the batch-1 children were created 30–45s after the Plan comment declaring `FABRIK_SPAWN_CHILD` (engine parse latency), while `#961` and `#965` were created in the same instant as the Review stage's own narration comments, and no Review comment on the parent contains a `FABRIK_SPAWN_CHILD` block at all — the stage agent ran `gh issue create` directly and then described having done so, so the engine never knew the children existed. Manually assigning, labeling, boarding, and setting Status to Specify on all four made each start within one poll. **This reproduction is what corrected the framing of this issue**: it initially looked like a second instance of the board-misrouting bug (no board item at all, which is even further off than "wrong board"), but the timestamp evidence rules that out — the engine's own `FABRIK_SPAWN_CHILD` handling was never invoked for this batch, so nothing about `AddProjectItem`/board resolution is implicated here.

**Whether failure mode [1] is still independently reproducible** (i.e. whether a *bona fide* engine-parsed `FABRIK_SPAWN_CHILD` spawn still misroutes to the parent's board) has not been re-confirmed since reproduction 2's correction — reproduction 1 predates the correction and reproduction 2 turned out to be a different code path entirely. This does not block the requirements above, since fixing cross-repo board resolution and assignment is required regardless of whether it is currently the live failure mode or a historical one — but Research should attempt to confirm or rule out live reproducibility of [1] on its own, since that affects test/verification strategy.

**Related prior work:** #984 (closed 2026-07-16) added retry around `UpdateProjectItemStatus` in `preImplement` for the case where a child *is* a board item and the Status-field set fails — not applicable to reproduction 2, where no board item was ever created. #769 established cross-repo `blockedBy` decomposition working correctly for Plan-spawned children (reproduction 1's batch did get `blocked_by` edges) — reproduction 2's missing edge is a regression against that established behavior in appearance only, since it went through a different, engine-bypassing code path rather than a regression in #769's own logic.

## Risks / Dependencies

- Conflating the two failure modes risks fixing the wrong layer: reproduction 1 (if still reproducible on its own) needs an engine-side board-resolution fix, while reproduction 2 needs a new sanctioned spawn route for non-Plan stages — a fix for one will not address the other. This spec requires both, so the risk is scoped rather than eliminated: Research/Plan should keep the two fixes separable in the implementation so one landing doesn't get credited for solving both.
- Extending `FABRIK_SPAWN_CHILD` to Review/Validate depends on those stages' prompts/skills actually steering agents toward emitting the declaration instead of calling `gh` directly; without that steering, agents may continue to take the path of least resistance even once a sanctioned route exists. Prompt/skill changes for `fabrik-review`/`fabrik-validate` are therefore in scope alongside the engine-side recognition, not a follow-up.
- The dependency re-verification requirement (resuming from `fabrik:blocked` must re-check actual dependency satisfaction) is the change most directly aimed at the dangerous silent-resume case observed in `repo-b#102`; it should be treated as load-bearing rather than a nice-to-have, even though it reads as the smallest individual requirement in the list.

arbeithand:
Closes #1419

## What this does

Fixes two independent failure modes in `FABRIK_SPAWN_CHILD` spawning that shared the same symptom (an unreachable spawned child, and a parent that believes it's unblocked when it isn't):

1. **Board/assignment misrouting** — a cross-repo spawn always registered the child on the *parent's own* project board and never assigned anyone. A `repo:`-scoped instance spawning into a repo it doesn't serve now refuses loudly (pause + explanatory comment) instead of silently mis-registering. Every spawned child is now unconditionally assigned to the spawning instance's `user:`.
2. **Engine-bypass gap** — Review and Validate had no sanctioned route to declare a mid-flight blocker, so an agent would call `gh issue create` directly, which the engine never observes (no board registration, no assignee, no `blocked_by` edge — silently leaving the parent unaware it's blocked). Review and Validate now recognize `FABRIK_SPAWN_CHILD_BEGIN/END` in their own raw output and route through the same `spawnChildren` Plan already uses.
3. **Dependency re-verification on resume** — `checkDependencies`'s live re-read on resume from `fabrik:blocked` used to skip entirely when the cached `BlockedBy` list was empty (exactly the shape a bypassed spawn produces). It now fires unconditionally on resume, closing the silent-resume hazard that let a regression merge in the original report.

## Key changes

- `github/issues.go`: `CreateIssue` gains a mandatory `assignees` param, folded into the same POST as creation (single fail-loud path).
- `engine/spawn.go`: new `spawnTargetServedByThisInstance` scope-check; `spawnChildren` now returns the spawned-identifier list and is shared by three callers (`preImplement`, `recoverMissingPlanComment`, and the new Review/Validate hook).
- `engine/dependencies.go`: widened live-reread gate for `checkDependencies`.
- `engine/item.go`: new `FABRIK_SPAWN_CHILD` recognition in `finalizeStageOutcome` for Review/Validate, parsed from raw stage output — mirrors the existing `FABRIK_PR_CREATE` handling exactly, including the fail-loud short-circuit.
- Both `fabrik-review` and `fabrik-validate` skills gain steering toward the declaration route over `gh issue create`; Validate's guidance is stricter (must not complete in the same turn as a genuine spawn). `attemptMergeOnValidate` (`engine/stages.go`) also gained an engine-side live `BlockedBy` re-check immediately before any merge/auto-merge/enqueue decision, closing the same-turn race directly rather than relying on prompt discipline alone — the skill instruction is now defense-in-depth, not the only guard.
- `engine/item.go`'s mid-flight spawn hook strips the raw `FABRIK_SPAWN_CHILD_BEGIN/TITLE/END` block from posted output via the new `stripSpawnBlocks` (mirrors the existing `stripMarkers` treatment of `FABRIK_PR_CREATE`/`FABRIK_ISSUE_UPDATE`), so the internal marker syntax doesn't leak into the human-facing PR/issue comment.
- `docs/state-machine.md`, `docs/USER_GUIDE.md`, `CLAUDE.md`, `plugin/fabrik-workflows/LABELS.md` all updated; new ADR at `adrs/1419-cross-repo-spawn-servability-and-midflight-recognition.md`.

## How to test

- `go test -race ./...` — full suite passes; two pre-existing, unrelated failures confirmed present on the base commit (`engine`'s `TestInvokeClaude_FabrikEnvVarsInjected` — a real `GITHUB_TOKEN`/`GH_TOKEN` in the local environment leaking into an env-var-isolation test — and `cmd`'s `TestExecute_ConfigYAMLApplied` SIGINT-handling flake).
- New tests: `engine/spawn_test.go` (scope-check refusal, multi-repo pass-through, same-repo pass-through, mandatory assignment), `engine/dependencies_test.go` (stale-empty-cache resume), `engine/spawn_midflight_test.go` (full Review/Validate spawn wiring, prose-only non-spawn, failure pause/suppression, marker-stripping).


arbeithand:
*branch: fabrik/issue-1419 | commit: 88cb6ddc | main: 8364069b0802af092fb0dcedb005a07bd290166f | 2026-08-09 04:23 UTC*

Everything is committed and pushed. Let me emit the PR creation marker.


Implemented both fixes from the Plan: a board-servability scope-check + mandatory assignment for cross-repo spawns (failure mode 1), and mid-flight `FABRIK_SPAWN_CHILD` recognition for Review/Validate sharing the same hardened `spawnChildren` (failure mode 2), plus a widened dependency-resume re-verification gate. All 13 planned tasks are committed, docs/ADR updated, and the full test suite passes (one pre-existing unrelated failure confirmed on the base commit).

---
Used 13/100 turns, 5.4M input (25 raw + 5.0M cache-read + 413k cache-write) / 4k output tokens.

arbeithand:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-1419 | commit: b91277f3 | main: 8364069b0802af092fb0dcedb005a07bd290166f | 2026-08-09 04:41 UTC*

Reviewed the #1419 implementation (cross-repo spawn board-servability, mandatory assignment, Review/Validate mid-flight `FABRIK_SPAWN_CHILD` recognition, and widened dependency re-verification on resume). Found and fixed one real gap: `attemptMergeOnValidate` could merge a yolo PR in the same dispatch that a Validate agent's mid-flight spawn declared a new blocker, before `fabrik:blocked` was ever applied — the exact silent-merge-past-a-blocker danger this issue exists to close, previously guarded only by a prompt instruction. Added a live dependency re-check inside `attemptMergeOnValidate` (with a regression test), corrected ADR-1419/state-machine.md/USER_GUIDE.md, which had documented this as an accepted unmitigated hazard, and regenerated `docs/llms-full.txt`. Full test suite passes (2 pre-existing, unrelated environment failures confirmed on the base commit). Rebased onto current main and force-pushed (the branch's earlier remote history was stale against main). PR is ready for human review.

---
Used 96/100 turns, 16.0M input (189 raw + 15.7M cache-read + 224k cache-write) / 48k output tokens.

arbeithand:
*branch: fabrik/issue-1419 | commit: 15c2bcc9 | main: 8364069b0802af092fb0dcedb005a07bd290166f | 2026-08-09 04:52 UTC*

Working tree is clean, both fixes are committed and pushed, and both Pruefer review threads are resolved.

**Summary of work done:**

1. **Finding [1] (raw marker leak)**: Fixed — `finalizeStageOutcome`'s mid-flight spawn hook now calls a new `stripSpawnBlocks` helper (sharing `ParseSpawnBlocks`'s own line-scanning) to remove the raw `FABRIK_SPAWN_CHILD_BEGIN`/`TITLE:`/`FABRIK_SPAWN_CHILD_END` block from the posted comment, mirroring the existing `stripMarkers` treatment of `FABRIK_PR_CREATE`/`FABRIK_ISSUE_UPDATE`. Added unit tests (no block, single block, multiple blocks, malformed block left visible, prose-only mention unaffected) and extended the integration test to assert the posted comment no longer contains raw marker syntax.

2. **Finding [2] (same-turn merge race)**: Already fixed in this same review pass, prior to Pruefer's comment landing — `attemptMergeOnValidate` now performs a live `FetchItemDetails` + `checkDependencies` re-check immediately before any merge/auto-merge/enqueue decision, closing the race where a Validate agent's mid-flight spawn + `FABRIK_STAGE_COMPLETE` in the same turn could merge past a blocker that had just been declared.

Both fixes are covered by tests, the full suite passes (only pre-existing, unrelated environment flakes remain), docs (ADR-1419, `state-machine.md`, `USER_GUIDE.md`) were updated to match, and `llms-full.txt` was regenerated per the docs-drift requirement. No issue body update is needed — this comment-processing round only involved code review findings, not spec changes.

---
**Threads addressed:**
- `engine/item.go:1351` — resolved
- `engine/item.go:1342` — resolved

arbeithand:
*branch: fabrik/issue-1419 | commit: 50c4bade | main: 89ca48fe07c0e8841c82119add761fd86ea05de7 | 2026-08-09 05:02 UTC*

## Validation Report

### Requirements: 7/7 verified

- [x] **`FABRIK_SPAWN_CHILD` recognized from Review/Validate** — `finalizeStageOutcome` (`engine/item.go`) parses raw stage output for both stages, mirroring the `FABRIK_PR_CREATE` precedent; covered by `engine/spawn_midflight_test.go`.
- [x] **Identical wiring regardless of spawn origin** — Review/Validate route through the same `spawnChildren` used by Plan/Implement (three callers, one implementation); board, assignment, labels, and `blocked_by` are identical by construction.
- [x] **Cross-repo board resolved from child's repo, fail loud on mismatch** — `spawnTargetServedByThisInstance` scope-check runs before any GitHub mutation; a `repo:`-scoped instance targeting an unserved repo pauses the parent with an explanatory comment rather than mis-registering.
- [x] **Every spawned child assigned to `cfg.User`** — folded into `CreateIssue`'s POST (`assignees` field), single fail-loud path; `github/issues.go`/`engine/interfaces.go` updated consistently.
- [x] **`blocked_by` edge failure fails the stage** — already true pre-PR for `AddBlockedByIssue`; unchanged and still exercised by tests.
- [x] **Resume-from-`blocked` re-verifies dependencies** — `checkDependencies`'s live re-read now fires on `alreadyBlocked` alone, closing the stale-empty-cache hazard that caused the original silent-resume regression.
- [x] **Failures surface visibly, never only in logs** — every new failure path (`pauseIssue` + comment) matches the existing fail-loud convention used throughout `spawnChildren`.

**Beyond the spec, but load-bearing:** a same-turn hazard (a Validate agent spawning a blocker and emitting `FABRIK_STAGE_COMPLETE` in one turn, before `fabrik:blocked` lands, could let a yolo PR merge past its own just-declared blocker) was caught in Review and closed with a live `checkDependencies` re-check inside `attemptMergeOnValidate` — verified present and tested.

### Test Suite: PASSED
- Full suite, race detector on: all packages pass except two confirmed **pre-existing** failures, independently reproduced identically against unmodified `origin/main` in a throwaway worktree: `engine`'s `TestInvokeClaude_FabrikEnvVarsInjected` (a real `GITHUB_TOKEN`/`GH_TOKEN` in this local environment leaking into an env-isolation assertion) and `cmd`'s `TestExecute_ConfigYAMLApplied` (SIGINT-handling timing flake, non-deterministic — didn't reproduce on the second run).
- `go build ./...`, `go vet ./...`: clean, both pre- and post-rebase.
- `docs/llms-full.txt`: regenerated, no drift.

### Issues Found and Fixed
- PR #1490's description had gone stale relative to its own code: it still described the Validate same-turn merge race as an unresolved, prompt-only mitigation, but a prior commit in this same branch (`f9f8db82`) had already engine-side-fixed it — flagged explicitly by Pruefer's own review as something that "could mislead a reviewer who reads only the description." Corrected the description via `gh pr edit` to reflect the actual fixed state and the marker-stripping fix.

### Branch State
- Rebased cleanly onto latest `main` (11 commits replayed, no conflicts), force-pushed.
- Working tree clean, all commits pushed.
- PR #1490 up to date, description corrected.

### Verdict: READY TO MERGE

Validated #1419's cross-repo spawn fix: all 7 requirements verified against code (board-servability scope-check, mandatory assignment, Review/Validate mid-flight `FABRIK_SPAWN_CHILD` recognition, fail-loud `blocked_by`, widened dependency re-verification), plus a same-turn merge-race guard caught in Review. Full test suite passes (2 confirmed pre-existing, environment-caused failures unrelated to this PR). Corrected a stale claim in PR #1490's description that Pruefer's review flagged. Rebased cleanly onto main and pushed. PR mergeable: MERGEABLE, mergeStateStatus: CLEAN.

---
Used 60/100 turns, 6.0M input (98 raw + 5.9M cache-read + 139k cache-write) / 14k output tokens.

arbeithand:
*branch: fabrik/issue-1419 | commit: 50c4bade | main: 89ca48fe07c0e8841c82119add761fd86ea05de7 | 2026-08-09 05:03 UTC*

Thread resolved. No code or issue-body changes are needed — the fix Pruefer requested was already applied.

## Comment Processing Summary

**Comment addressed:** Pruefer's review thread on `engine/stages.go:281`, noting the PR description still described the Validate same-turn merge race as an unfixed, documented-only gap, when the code (commit `f9f8db82`) had already closed it.

**Action taken:** Verified this was already resolved — the prior Validate stage invocation had independently caught the same staleness (flagged in that same review) and corrected the PR description via `gh pr edit` before this comment was even processed. Confirmed the current PR description accurately reflects the `attemptMergeOnValidate` live dependency guard as implemented, not as an open gap. No code changes were needed; working tree remains clean. Resolved the review thread.

No issue body update needed — this was a PR-description accuracy note, not a spec change.

---
**Threads addressed:**
- `engine/stages.go:281` — resolved

arbeithand:
🏭 **Fabrik merge-train** — Landed via batch PR #1492.

PR #1493, opened 2026-08-09
feat(engine): clean-stop shutdown — pause in-flight issues durably and bound the worker drain

### LINKED ISSUE #1393: feat(engine): clean-stop shutdown — pause in-flight issues durably and bound the worker drain
## Problem

Stopping the daemon is currently a kill, not a pause. There is no way to tell Fabrik "stand down cleanly" and have the board reflect that — so an operator who needs the machine back (to free the Claude subscription, to run the e2e bed, to reboot) either waits an unbounded amount of time for the queue to drain on its own, or accepts losing in-flight work and leaving the board in a state that only the *next* startup can repair.

### What shutdown does today

Single entry point, `Engine.Run()` (`engine/poll.go:102`); `signal.Notify(sigCh, SIGINT, SIGTERM)` at `:162`. There is no signal handling anywhere else — the TUI starts bubbletea with `tea.WithoutSignalHandler()` (`cmd/root.go:1232`) and its `q`/`ctrl+c` quit path raises `SIGTERM` on its own PID (`cmd/root.go:1273-1280`), so TUI quit and bare-CLI Ctrl-C are the same code path.

Two-signal semantics already exist (`engine/poll.go:177-201`):

1. **First signal** — stores kill reason `"daemon_shutdown"` in every per-issue `killReasonHolder` (`:183-186`), then cancels the root context (`:187`).
2. **Second signal** — `"Force-quitting..."`, releases the terminal, `os.Exit(1)` (`:194-200`). No label cleanup, no worker wait.

On the first signal the root context cancel propagates to `exec.CommandContext`, whose `cmd.Cancel` override (`engine/claude.go:1099-1119`) calls `killProcGroupGraceful` → `SIGINT` → wait → `SIGTERM` → wait → `SIGKILL` on the process group (`engine/procattr_unix.go:54-90`), with 10s/10s default grace (`engine/claude.go:87,91`). Then `cleanupLockedIssues()` and `e.wg.Wait()` (`engine/poll.go:616-617`).

So the grace period is real, but it is grace *to die*, not grace *to finish*. Nothing waits for Claude to reach a natural stopping point, and there is no flag or config key that asks for one.

### The three gaps this creates

**1. Board state does not record that we stopped.** `cleanupLockedIssues()` (`engine/poll.go:676-703`) removes `fabrik:locked:<user>` and nothing else. No `fabrik:paused`. No comment. An issue that was mid-Implement when the daemon stopped looks, on the board, exactly like an issue that is mid-Implement right now — the operator cannot tell parked-by-me from running.

**2. `stage:<name>:in_progress` removal is incidental and racy.** No shutdown code path clears it. It is removed only if the worker goroutine reaches `finalizeStageOutcome`'s cancellation branch (`engine/item.go:1207-1212`) and its `releaseLock()` closure (`engine/item.go:948-961`) — a live GitHub API call issued during an unbounded drain, skipped entirely on force-quit or crash.

The recovery model is therefore "clean up on next start", and it has a hole. The two startup passes are each conditional:

| Pass | Location | Condition |
| --- | --- | --- |
| `runStartupCleanup` | `engine/worker_liveness.go:167-215` | requires a stale **lock label** |
| `runStartupOrphanedInProgressScan` | `engine/worker_liveness.go:238-283` | requires a sibling **`complete`/`failed`** label |

An item with `stage:X:in_progress`, no lock label, and no `complete`/`failed` sibling matches neither — and that is precisely the shape produced when `cleanupLockedIssues()` succeeds in removing the lock while the worker dies before removing `in_progress`. `transientLifecycleLabels` (`engine/poll.go:737-748`) contains no `stage:*:in_progress` entry either, so the closed-item sweep cannot reach it. Note `runStartupTransientLabelScan` is additionally a documented no-op on the default cold-start path (`engine/startup.go:194-201`).

**3. In-flight work is discarded, not saved.** `commitWIP` (`engine/item.go:2178-2185`) is called at `engine/item.go:1441`, which is *after* the `p.ctx.Err() != nil → releaseLock(); return` early-out at `:1207-1212`. A shutdown mid-stage therefore leaves the worktree dirty and uncommitted and skips the branch push (`:1449-1454`). The one worktree guarantee that does hold: a read-only stage's `git stash` is popped before the cancel check (`:1199-1206`), so no dangling stash.

`internal/itemstate` is purely in-memory (`store.go:32-60`) with no flush or reload, consistent with ADR-002 — so retry counts, cooldowns and check-run buffers are lost by design. That is fine and not in scope; GitHub is the durable store. The point is that shutdown currently writes *almost nothing* to that durable store.

### Prior art: the mechanism already exists, scoped to one item

`handleStopRequest` (`engine/item.go:2556-2586`) — the TUI's per-item stop — does exactly what this issue asks for, for a single issue:

```go
// cancels the issue's context with reason "user_stop"   :2564-2568
e.addLabel(item, "fabrik:paused")                        // :2576
e.addLabel(item, "fabrik:awaiting-input")                // :2579
// posts a "stopped from TUI by <user>" audit comment    :2582-2586
```

So the semantics, the label pair, and the audit-comment convention are already settled in-repo (see also `docs/state-machine.md:719`'s state-transition row for the same path). What is missing is a daemon-wide caller. That should shape the implementation: this is plausibly a generalization of an existing function rather than a new subsystem, and the two must not diverge.

## Summary

Give the daemon a clean-stop mode: on a shutdown request, stop admitting new work, mark every in-flight issue `fabrik:paused` with an audit trail, wind the Claude workers down within a bounded deadline, and leave the board in a state a human can read and resume from — rather than relying on the next startup to guess what happened.

## Requirements

- **R1 — A shutdown request pauses in-flight issues durably.** Every issue with a live worker gets `fabrik:paused` (and `fabrik:awaiting-input`, if R3 decides to match `handleStopRequest`) written to GitHub before the process exits, plus an audit comment naming the reason. The label write is the load-bearing part: it is what makes the stop visible and resumable.
- **R2 — `stage:<name>:in_progress` is cleared for every paused issue, by shutdown code, not incidentally.** Do not rely on each worker goroutine reaching `releaseLock()` in time. If a clear fails, that is a reportable outcome, not a silent one.
- **R3 — Decide and record whether daemon shutdown reuses `handleStopRequest` or a shared extraction of it.** They must agree on the label pair, the comment text, and the kill reason. Two subtly different "stopped" states on the board would be worse than today. State the decision even if the answer is "duplicated deliberately, because X".
- **R4 — The drain is bounded and the bound is configurable.** `e.wg.Wait()` (`engine/poll.go:617`) is unbounded today; the only bound is a second Ctrl-C. A clean stop must terminate in a stated, finite time. Pick a default and justify it against the existing 10s/10s `kill_grace` windows (`engine/claude.go:87,91`) — the drain deadline must exceed the per-worker kill escalation or it will preempt it.
- **R5 — Force-quit still works and still bypasses everything.** The second-signal path (`engine/poll.go:194-200`) is the operator's escape hatch when a clean stop is itself wedged. It must not acquire new obligations. Its behavior is unchanged; only its message may need to reflect that a clean stop was in progress.
- **R6 — Do not change what `fabrik:paused` means.** It already means "a human must act before anything else happens", it already suppresses dispatch, and as of #1379 it also suppresses deep-fetch. Shutdown becomes a new *writer* of an existing label, not a new state. Resuming is the operator removing the label, exactly as it is today.
- **R7 — Fix the orphaned-`in_progress` startup hole as part of this, or state explicitly why not.** Even with R2, a crash or force-quit still produces the bare-`in_progress` shape that neither startup pass matches (see Problem, gap 2). R2 narrows the window; it does not close it. Recovery on next start is the backstop and should be sound independently.
- **R8 — Decide what happens to uncommitted worktree work, and record it.** Today it is silently left dirty and unpushed (Problem, gap 3). Options include: commit-and-push via the existing `commitWIP` before the cancel early-out; leave it dirty but say so in the audit comment so the operator knows where the work is; or leave it and document the behavior. Calling this "save state" while discarding the actual work product would be the wrong outcome — but committing partial work under a shutdown deadline has its own risks. **This is a design call and should be settled in Research, not assumed.**
- **R9 — Behavior when no workers are in flight is unchanged and fast.** A stop with an idle queue must not become slower or noisier than today.

### In scope

- `engine/poll.go` — the signal handler (`:177-201`), `cleanupLockedIssues` (`:676-703`), the drain (`:616-617` and its three identical copies at `:625-626`, `:649-650`, `:658-659`)
- `engine/item.go` — `handleStopRequest` (`:2556-2586`) and the cancellation early-out (`:1207-1212`)
- `engine/worker_liveness.go` — the two startup passes, for R7
- `cmd/root.go` — a drain-deadline flag/env if R4 adds one; the TUI quit path (`:1273-1280`)
- `adrs/` — a new ADR numbered after this issue (`adrs/<N>-*.md`), recording R3, R4 and R8. **There is no shutdown ADR today** — ADR-054 covers *how* we kill (SIGINT escalation, reason propagation) but never whether to drain or what to write.
- `docs/state-machine.md` — the shutdown rows are currently a one-liner (`:2915` "Graceful shutdown | Engine receives SIGINT/SIGTERM | `cleanupLockedIssues()`"); also `:2232`, `:2469`. Regenerate `docs/llms-full.txt` in the same commit.

### Out of scope

- No persistence layer for `internal/itemstate`. GitHub remains the durable store (ADR-002); in-memory bookkeeping is expected to be lost.
- No change to the SIGINT→SIGTERM→SIGKILL escalation itself (ADR-054), only to what surrounds it.
- No change to SIGHUP's drain-and-re-exec (`engine/sighup_unix.go:55`) beyond whatever falls out of sharing the drain path — if it does inherit the new behavior, say so, since a restart-in-place pausing every issue may not be wanted.
- Not a "finish the current stage" mode. Waiting for a Claude invocation to complete naturally is unbounded by nature; this issue is about stopping *promptly* and *legibly*, not about avoiding interruption.
- No change to `fabrik:paused` semantics for any other writer.

## Open Questions

None outstanding. The remaining design decisions (R3's reuse-vs-extract call, R4's concrete deadline value, R8's uncommitted-work handling) are deliberately left to Research/Plan rather than pre-decided here — they require reading the current implementation to answer soundly, and the issue is explicit about where each must be recorded (ADR, this issue's acceptance criteria).

## Prior Art / Context

- **In-repo**: `handleStopRequest` (`engine/item.go:2556-2586`) already implements this exact pattern — cancel-with-reason, `fabrik:paused` + `fabrik:awaiting-input`, audit comment — for a single TUI-initiated stop. `docs/state-machine.md:719` documents its state-transition row. This is the primary prior art and the strongest argument for R3 resolving toward reuse/extraction rather than a parallel implementation.
- **General pattern**: "stop admitting new work, then wait up to a deadline for in-flight work to finish, then force" is the standard graceful-shutdown shape (e.g. Go's `net/http.Server.Shutdown(ctx)`, Kubernetes' SIGTERM + `terminationGracePeriodSeconds` + SIGKILL). Fabrik already has the force-kill half (ADR-054's SIGINT→SIGTERM→SIGKILL escalation per worker); what's missing is the daemon-level "stop admitting work + bounded wait + durable record" wrapper around it. This issue is that wrapper, not a new escalation mechanism.

## Risks / Dependencies

- **AC1/AC3 tension**: an implementation that blocks forever on label writes trivially satisfies AC1; an implementation that writes nothing trivially satisfies AC3. Tests must be proven non-vacuous by neutralizing the change under test and confirming the expected assertion goes red (see Verification Note below).
- **Drain deadline vs. kill-grace windows**: if the chosen default drain deadline (R4) is shorter than the existing 10s/10s `kill_grace` escalation (`engine/claude.go:87,91`), the drain will preempt the escalation mid-flight, defeating its purpose. The two must be reconciled, not chosen independently.
- **SIGHUP sharing**: if the drain path is shared with SIGHUP's drain-and-re-exec (`engine/sighup_unix.go:55`), a restart-in-place would start pausing every in-flight issue, which may be an unwanted behavior change for what is currently a live restart. Must be explicitly decided, not inherited by accident.
- **Comment idempotency under retry**: AC2 requires exactly one audit comment per paused issue even if the label write is retried — the implementation must guard against double-posting on partial failure/retry paths.

## Acceptance Criteria

- **AC1** — After a clean stop with N workers in flight, all N issues carry `fabrik:paused` on GitHub and none carries `stage:<name>:in_progress`. Assert against the API/mock, not against in-memory state.
- **AC2** — The same N issues each carry exactly one audit comment naming the shutdown. Exactly one — a stop must not post duplicates if the label write is retried.
- **AC3** — A clean stop with workers in flight completes within the configured deadline. Assert on elapsed time against a deliberately slow/unresponsive worker, so the test fails if the drain is still effectively unbounded.
- **AC4** — Force-quit during a clean stop still exits promptly and is not blocked by the new work (R5).
- **AC5** — A stop with an idle queue writes no labels and no comments, and is no slower than today (R9).
- **AC6** — On the next startup, an issue paused by a clean stop is not re-dispatched, and removing `fabrik:paused` re-admits it within one poll cycle. This is the resume path and is the criterion most likely to be got wrong.
- **AC7** — R7's orphaned-`in_progress` case: an item with `stage:X:in_progress`, no lock label, and no `complete`/`failed` sibling is healed on next startup (or the decision not to is recorded).
- **AC8** — R3 and R8's decisions are recorded in the ADR, including what a resumed issue's worktree looks like.

## Verification Note

AC1 and AC3 fail in opposite directions and must hold together: AC1 passes trivially against an implementation that blocks forever writing labels, and AC3 passes trivially against one that writes nothing and exits immediately. Prove each non-vacuous by neutralizing the change and confirming the expected one goes red.

AC3 in particular must use a worker that does not exit on `SIGINT`/`SIGTERM`, or it will pass on the kill escalation alone and prove nothing about the drain deadline.

## References

- `engine/poll.go:102,162,177-201,616-617,676-703,737-748` — `Run`, signal handler, drain, lock cleanup, transient labels
- `engine/item.go:948-961,1207-1212,1441,2178-2185,2556-2586` — `release` closure, cancellation early-out, `commitWIP` call site, `handleStopRequest`
- `engine/claude.go:87,91,1099-1119` — `kill_grace` defaults, `cmd.Cancel` override
- `engine/procattr_unix.go:54-90` — the SIGINT→SIGTERM→SIGKILL group escalation (no-op on Windows, `procattr_windows.go:18`)
- `engine/worker_liveness.go:167-215,238-283` — the two conditional startup passes
- `cmd/root.go:1232,1273-1280` — TUI signal-handler suppression and SIGTERM self-raise
- ADR-002 (GitHub as state store), ADR-054 (SIGINT kill escalation — the only shutdown-adjacent ADR, scoped to *how* we kill), ADR-036 (in-memory single-owner store)
- #1379 — `fabrik:paused` now also suppresses deep-fetch admission, which makes a paused-on-shutdown board genuinely cheap to leave parked

arbeithand:
Closes #1393

## What this does

Gives the daemon a real clean-stop mode instead of a kill. On the first SIGINT/SIGTERM (bare CLI Ctrl-C, or the TUI's `q`/Ctrl-C, which both funnel into the same handler):

1. **Stops admitting new work** immediately (unchanged).
2. **Durably pauses every in-flight issue** — `fabrik:paused` + `fabrik:awaiting-input` applied on GitHub, `stage:<Name>:in_progress` cleared directly, and a one-time audit comment posted — via a new shared primitive, `pauseInterruptedIssue` (`engine/mutate.go`), used by both the daemon-wide clean stop and the pre-existing TUI single-issue stop (`handleStopRequest`). Idempotent under a retried/duplicate call.
3. **Commits and pushes uncommitted worktree changes** before releasing the lock, for any cancellation reason (daemon shutdown or TUI stop) — `finalizeStageOutcome`'s cancellation early-out no longer silently discards in-progress work.
4. **Bounds the drain** with a new `--drain-deadline`/`FABRIK_DRAIN_DEADLINE` config (default 30s, comfortably above the 20s worst-case kill-grace escalation), via a new `waitGroupTimeout` primitive. The four previously-identical inline drain blocks in `poll.go` are collapsed into one `drainAndExit` helper.
5. **Force-quit (second signal) still works** — and, while verifying this, I found and fixed a genuine pre-existing bug: the second-signal listener raced against the just-closed `ctx.Done()` instead of a real "are we still waiting for a second signal" condition, so a human physically could not press Ctrl-C twice fast enough for force-quit to ever actually fire. Confirmed via a subprocess reproduction against the pre-fix code (had to `SIGKILL` a helper process that ignored two SIGINTs) and again after the fix (clean `exit 1`).
6. **Startup recovery gets a third pass** (`runStartupBareInProgressScan`) that heals a `stage:<Name>:in_progress` label surviving with no lock label and no terminal sibling — the residual gap even the direct clear above can't fully close (crash, force-quit, or a failed pause write).

SIGHUP's drain-and-restart-in-place inherits the new *bounded* wait as a side effect of sharing `drainAndExit`, but deliberately does **not** inherit the pause — the pause-write phase is only ever launched from the SIGINT/SIGTERM goroutine.

Full design rationale, including the R3/R4/R7/R8 decisions and the force-quit fix, is recorded in `adrs/1393-clean-stop-shutdown.md`.

## How to test

- `go test -race ./engine/... ./cmd/...` — new coverage: `engine/shutdown_test.go` (AC1–AC5, the AC3/AC4 non-vacuity requirements from the issue's Verification Note, including a subprocess-based force-quit test), `engine/worker_liveness_test.go` (AC7), `engine/stop_test.go` and `engine/item_extra_test.go` (R2/R3/R8 additions), `cmd/resolve_helpers_test.go` (`TestDrainDeadline`).
- Manually: start Fabrik with an issue mid-stage, Ctrl-C once — the issue should show `fabrik:paused` + an audit comment on GitHub, worktree changes committed/pushed, and the process exit within `--drain-deadline`. Ctrl-C twice in quick succession during a stuck drain should force-exit immediately.

Note: `cmd.TestExecute_ConfigYAMLApplied` is flaky/failing in this environment due to real (unauthenticated) network calls to github.com during label-seeding — confirmed pre-existing by reproducing the identical failure against the unmodified base commit in an isolated worktree; unrelated to this change.


arbeithand:
*branch: fabrik/issue-1393 | commit: ece85c22 | main: 1b4a506dd180cc08b5d5bcc54030c54b1ef41262 | 2026-08-09 05:20 UTC*

Everything is committed and pushed. Now let's emit the PR creation marker.


Implemented daemon-wide clean-stop shutdown: SIGINT/SIGTERM now durably pauses every in-flight issue (`fabrik:paused`/`fabrik:awaiting-input`, direct `in_progress` clear, audit comment) via a primitive shared with the TUI's single-stop, commits/pushes uncommitted worktree work before releasing locks, and bounds the drain with a new configurable `--drain-deadline` (default 30s). Added a third startup-recovery pass for orphaned `in_progress` labels, and — while verifying force-quit — found and fixed a genuine pre-existing bug where the second-signal listener could never actually observe a second signal. All decisions recorded in `adrs/1393-clean-stop-shutdown.md`; docs and `llms-full.txt` updated; full engine/cmd test suites pass except one pre-existing, network-dependent flaky test confirmed unrelated to this change.

---
Used 95/100 turns, 36.4M input (190 raw + 36.0M cache-read + 433k cache-write) / 42k output tokens.

arbeithand:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-1393 | commit: 513a4c30 | main: 1b4a506dd180cc08b5d5bcc54030c54b1ef41262 | 2026-08-09 05:39 UTC*

## Review Findings

### Fixed

- **Issue**: The SIGINT/SIGTERM signal handler called `e.wg.Add(1)` (registering the new R1/R2 pause-write phase) *after* `cancel()`, with no happens-before relationship to the main poll loop's `drainAndExit` → `waitGroupTimeout` → `e.wg.Wait()`, which is only reachable by observing `ctx.Done()` — the very thing `cancel()` closes. Whenever the sole in-flight worker's own `e.wg.Done()` dropped the counter to zero in the same window as this race, the Go runtime panics with `"sync: WaitGroup misuse: Add called concurrently with Wait"` — a crash in the exact code meant to make shutdown clean, in the ordinary one-worker clean-stop case this feature exists for. I confirmed this was genuinely reachable with a standalone repro before touching the code.

  **Fix**: Extracted the shutdown-initiation sequence into `Engine.beginShutdownPause`, calling `e.wg.Add(1)` strictly before `cancel()` — since `cancel()` is the only thing that closes `ctx.Done()`, and Go sequences statements within one goroutine, no other goroutine can observe `ctx.Done()` (and thus reach `Wait()`) before the `Add` has already happened. Added `TestBeginShutdownPause_NoAddWaitRace`, which stresses the exact interleaving for 2000 iterations under `-race`; verified it reliably fails against the pre-fix ordering (both a runtime panic and a `-race` "WaitGroup reused before previous Wait has returned" report, depending on timing) and passes cleanly against the fix.

### Verified

- Rebased onto latest `main` (already current — no conflicts).
- `go build ./...`, `go vet ./...`, `gofmt -l` all clean.
- Full `engine` + `cmd` suites pass under `-race -timeout 5m`: 2451/2451 in `engine` except one pre-existing, unrelated flake (`TestInvokeClaude_FabrikEnvVarsInjected`, an env-leak issue in a file this PR doesn't touch, already noted by the Implement stage). `cmd`'s `TestExecute_ConfigYAMLApplied` also occasionally times out — I confirmed by reverting to the pristine pre-review commit that this is pre-existing and network-latency dependent (it hits real `api.github.com` during startup label-seeding, unrelated to this PR's files), not something this PR introduced.
- No new external review comments on PR #1493 since Implement.
- All AC1–AC9 tests from Implement still pass (`TestRunShutdownPause_*`, `TestDrainAndExit_BoundedByDrainDeadline_AC3`, `TestForceQuit_DuringCleanStop_AC4`, `TestHandleStopRequest_*`, worker-liveness R7 tests, R8 WIP-commit tests).
- Rest of the implementation — `pauseInterruptedIssue`/`pauseIssue` reuse (R3), direct `in_progress` clearing (R2), `runStartupBareInProgressScan` (R7), commit-before-cancel (R8), the ADR — all read correctly against the plan and issue requirements.

### Blocking

None. PR is ready for human review.

Reviewed the clean-stop shutdown implementation and found + fixed a genuine, confirmed `sync.WaitGroup` misuse race in the SIGINT/SIGTERM handler (Add-after-cancel racing the drain's Wait, reproducibly causing a panic/crash exactly in the ordinary one-worker shutdown case). Extracted the sequence into `beginShutdownPause` with the correct ordering and added a 2000-iteration `-race` regression test verified to fail pre-fix and pass post-fix. Full test suite passes except two pre-existing, unrelated environment flakes. PR is ready for human review.

---
Used 93/100 turns, 13.7M input (184 raw + 13.5M cache-read + 193k cache-write) / 49k output tokens.

arbeithand:
*branch: fabrik/issue-1393 | commit: a603208f | main: 1b4a506dd180cc08b5d5bcc54030c54b1ef41262 | 2026-08-09 05:51 UTC*

All three Pruefer review findings have been addressed:

1. **Add/Wait WaitGroup race** (already fixed in the prior turn, commit `513a4c30`) — confirmed reproducible with a standalone repro, fixed by extracting `beginShutdownPause` and calling `e.wg.Add(1)` strictly before `cancel()`, with a 2000-iteration `-race` regression test.

2. **Double-post race between TUI stop and daemon shutdown** — fixed by moving the `alreadyPaused` idempotency check inside `pauseInterruptedIssue` itself, serialized on a new per-issue mutex (`Engine.pauseIssueMu`), so a concurrent caller always observes the other's already-applied label instead of racing on stale snapshots. Verified with a new concurrency regression test (`TestPauseInterruptedIssue_ConcurrentCallers_NoDuplicateComment`) that reliably fails pre-fix and passes post-fix.

3. **Overclaiming audit comment wording** — reworded to describe the commit/push as attempted, not guaranteed, since it runs on an unsynchronized goroutine and failures are only logged.

Build, vet, gofmt, and the full `-race` test suite all pass (only the same two pre-existing, network/env-dependent flakes unrelated to this PR's files remain). Commit `a603208f` is pushed to `fabrik/issue-1393`, and all three review threads are resolved.

No issue body changes are needed — these were implementation-level review findings, not spec changes.

---
**Threads addressed:**
- `engine/poll.go:289` — resolved
- `engine/shutdown.go:109` — resolved
- `engine/shutdown.go:112` — resolved

arbeithand:
*branch: fabrik/issue-1393 | commit: 9609bf42 | main: 1b4a506dd180cc08b5d5bcc54030c54b1ef41262 | 2026-08-09 05:57 UTC*

All good. Validation complete.

## Validation Report

### Requirements: 9/9 verified against issue.md
- [x] **R1** — `runShutdownPause`/`pauseIssueForDaemonShutdown` (`engine/shutdown.go`) applies `fabrik:paused` + `fabrik:awaiting-input` + audit comment for every issue with a live `Worker()`, before `Run()` returns.
- [x] **R2** — `stage:<Name>:in_progress` is cleared directly via `removeInProgressLabel` in the shutdown code itself, not delegated to the worker's own `release()`; safe-by-construction against a concurrent removal (`ErrNotFound`-is-success).
- [x] **R3** — `handleStopRequest` and daemon shutdown both converge on the new `pauseInterruptedIssue` primitive (`engine/mutate.go`), decision recorded in ADR-1393.
- [x] **R4** — 30s default `--drain-deadline`/`FABRIK_DRAIN_DEADLINE`, exceeds the 20s kill-grace worst case by design; `warnDrainDeadlineOrdering` flags misconfiguration.
- [x] **R5** — Force-quit path unchanged and unblocked; review additionally found and fixed a genuine pre-existing bug (second-signal listener racing `ctx.Done()`) that would have prevented force-quit from ever firing during a clean stop — fixed via `drainComplete` channel.
- [x] **R6** — `fabrik:paused` semantics unchanged; shutdown is a new writer only.
- [x] **R7** — `runStartupBareInProgressScan` (`engine/worker_liveness.go`) closes the orphaned-`in_progress` gap as a third, disjoint startup pass.
- [x] **R8** — `commitWIP`/`pushBranchUnlessQueued` hoisted before the cancellation early-out in `finalizeStageOutcome`, applied uniformly to any cancellation reason.
- [x] **R9** — `runShutdownPause` returns immediately with zero writes when no worker is in flight (verified by `TestRunShutdownPause_IdleQueue_NoWrites_AC5`).

### Test Suite: PASSED (non-vacuously, per the Verification Note)
- 3961 passed, 1 skipped across 20 packages; AC1–AC9 covered by dedicated, targeted tests (`TestRunShutdownPause_*`, `TestDrainAndExit_BoundedByDrainDeadline_AC3`, `TestForceQuit_DuringCleanStop_AC4`, `TestHandleStopRequest_*`, `TestPauseInterruptedIssue_ConcurrentCallers_NoDuplicateComment`, `TestRunStartupBareInProgressScan_*`, `TestFinalizeStageOutcome_CancelledInvocation_*_R8`, `TestDrainDeadline`), all re-run individually and passing.
- 2 failures, both confirmed pre-existing and unrelated to this PR's diff (neither failing file — `engine/invoke_claude_test.go`, `cmd/root_test.go` — appears in `git diff origin/main..HEAD`): `TestInvokeClaude_FabrikEnvVarsInjected` (ambient env-var leak) and `cmd`'s `TestExecute_ConfigYAMLApplied` (hits real `api.github.com`, network-latency dependent).
- `go build`, `go vet` clean; `gofmt -l` flags only files outside this PR's diff.
- `docs/llms-full.txt` regeneration verified byte-identical to committed version.

### Fix applied during Validate
Found and corrected one stale doc detail: ADR-1393's R3 section still showed `pauseInterruptedIssue`'s pre-review-fix signature (`alreadyPaused bool` parameter), which review had removed when it moved the idempotency check inside the function under a per-issue mutex. Fixed, committed (`9609bf42`), pushed.

### Verdict: READY TO MERGE

Validation passed: all 9 requirements and AC1–AC9 verified non-vacuously against targeted tests; full suite green except two confirmed pre-existing, unrelated flakes. Fixed one stale doc detail in the ADR (outdated function signature from before review's fix) and pushed. Rebase: skipped-up-to-date. PR mergeable: MERGEABLE, mergeStateStatus: BLOCKED (required checks pending post-push, not a conflict).

---
Used 43/100 turns, 4.9M input (83 raw + 4.7M cache-read + 143k cache-write) / 10k output tokens.

arbeithand:
*branch: fabrik/issue-1393 | commit: e93f012f | main: 1b4a506dd180cc08b5d5bcc54030c54b1ef41262 | 2026-08-09 06:14 UTC*

Thread resolved. Since this was a review finding (not an explicit "issue resolved" directive from the user), I won't signal stage completion — per the skill, that's the engine's/Validate stage's call.

No issue body changes are needed — this was an implementation-level review finding, not a spec change.

---
**Threads addressed:**
- `engine/shutdown.go:93` — resolved

arbeithand:
*branch: fabrik/issue-1393 | commit: e93f012f | main: 1b4a506dd180cc08b5d5bcc54030c54b1ef41262 | 2026-08-09 06:19 UTC*

`docs/llms-full.txt` is byte-identical after regen — no drift. Everything is clean, committed, and pushed. All checks pass.

## Validation Report

### Requirements: 9/9 verified
- [x] R1–R9 all previously verified against `engine/shutdown.go`, `engine/mutate.go`, `engine/item.go`, `engine/poll.go`, `engine/worker_liveness.go`, `cmd/root.go`, and `adrs/1393-clean-stop-shutdown.md`.
- [x] The latest fix (`e93f012f`) closes the third and final Pruefer review finding: `runShutdownPause`'s in-flight snapshot is now captured synchronously in `beginShutdownPause`, strictly *before* `cancel()` fires, closing a real race where a fast-cancelling worker could vanish from the pause set before the post-cancel scan ran. Confirmed by a targeted non-vacuous regression test (`TestBeginShutdownPause_SnapshotSurvivesRaceWithWorkerExit`, which explicitly asserts the race actually occurred before checking the fix held).

### Review Threads: all resolved
No unresolved review threads remain on PR #1493 (GraphQL query confirms zero `isResolved==false` threads).

### Test Suite: PASSED (non-vacuously)
- `gofmt -l` flags only files outside this PR's diff (pre-existing).
- `go test -race -timeout 5m ./...`: 3962 passed, 3 failed (2 grouped under 1 parent), 1 skipped, 20 packages.
- Both failures independently reproduced against pristine `origin/main` (`1b4a506d`) in an isolated worktree — confirmed pre-existing and unrelated to this PR's diff: `TestInvokeClaude_FabrikEnvVarsInjected` (ambient `NVM_INC` env leak) and `TestExecute_ConfigYAMLApplied` (hits real `api.github.com` during label seeding with a fake token).

### Verdict: READY TO MERGE

Validation passed. Rebase: skipped-up-to-date. PR mergeable: MERGEABLE, mergeStateStatus: CLEAN. All 9 requirements verified; the final Pruefer finding (pre-cancel snapshot race) is fixed and regression-tested. Both remaining test failures independently reproduced on pristine main — confirmed pre-existing, unrelated to this PR.

---
Used 23/100 turns, 2.4M input (43 raw + 2.3M cache-read + 111k cache-write) / 6k output tokens.

arbeithand:
*branch: fabrik/issue-1393 | commit: bd7809c1 | main: 1b4a506dd180cc08b5d5bcc54030c54b1ef41262 | 2026-08-09 06:34 UTC*

Thread resolved. Since this was a review finding (not an explicit "issue resolved" directive from the user), I won't signal stage completion — per the skill, that decision belongs to the engine's/Validate stage's normal flow, not comment processing.

No issue body changes are needed — this was an implementation-level review finding, not a spec change.

---
**Summary of work done:**

Fixed the Pruefer review finding: `pauseIssueMu` (the sync.Map serializing concurrent `pauseInterruptedIssue` calls) never removed entries, leaking one `*sync.Mutex` per distinct issue ever paused for the life of the daemon. Replaced it with a refcounted plain map (`acquirePauseIssueMutex`/`releasePauseIssueMutex`, guarded by a new `pauseIssueMuGuard`) that deletes an entry the instant no caller holds a reference — bounding the map by concurrently-pausing issues, not all-time total. Extended the existing 200-issue concurrency regression test to assert the map is empty afterward, and confirmed the assertion is non-vacuous (fails against a deliberately reintroduced leak). Build, vet, gofmt clean; full `-race` engine suite passes except the same pre-existing `TestInvokeClaude_FabrikEnvVarsInjected` env-leak flake noted throughout this PR's history. Commit `bd7809c1` pushed; review thread resolved.

---
**Threads addressed:**
- `engine/engine.go:134` — resolved

arbeithand:
🏭 **Fabrik merge-train** — Landed via batch PR #1495.

Issue #1496, opened 2026-08-09
fix(pruefer): re-review loop at unchanged head — COMMENTED reviews never supersede in FetchPRReviews


arbeithand:
## Problem

Pruefer re-reviews the same unchanged commit on every poll cycle, indefinitely. Its
guard against this has never worked in production, and the cause is a supersession bug
in `gh.FetchPRReviews` rather than in Pruefer itself.

### Observed

On `handarbeit/fabrik#1477`, Pruefer submitted **29 reviews**. The last eleven were
submitted against one unchanged commit, `a08167a2`, at ~19-minute intervals (the poll
interval), continuing for over three hours after Fabrik had paused the issue at
`FABRIK_MAX_REVIEW_CYCLES=5` and stopped pushing:

```
07:58  6b9fa7e7
08:16  4b333a27
08:32  a08167a2   <- last push; Fabrik pauses
08:58  a08167a2
09:15  a08167a2
...
11:47  a08167a2
```

### Root cause

`Eligible` (`pruefer/select.go:76`) skips a PR when `alreadyReviewedAtHead` finds a
review by `BotLogin` whose `CommitID` equals `PR.HeadSHA` (`select.go:138`). The
identity comparison is fine — REST reports the author as `handarbeit-pruefer[bot]`,
which is exactly `slug + "[bot]"` as `auth.go:171` builds it. The SHA comparison is
what fails, and it fails because `ExistingReviews` does not contain Pruefer's latest
review.

`ExistingReviews` comes from `gh.FetchPRReviews` (`pruefer/review.go:81`), which
collapses the raw submission history to one entry per author (`github/prs.go:88`):

```go
_, seen := latestByAuthor[r.User.Login]
if !seen {
    order = append(order, r.User.Login)
} else if r.State == "COMMENTED" {
    // A comment-only follow-up does not overwrite the author's
    // existing formal verdict.
    continue
}
```

The stated intent is correct and matches GitHub's own `reviewDecision`: a `COMMENTED`
follow-up must not overwrite a prior **formal** verdict (`APPROVED`,
`CHANGES_REQUESTED`, `DISMISSED`). But the condition tests only whether *any* prior
entry exists — not whether that entry is a formal verdict. When an author's prior
entry is itself `COMMENTED`, no verdict has been established, and the new `COMMENTED`
should supersede it. Instead it is discarded.

The function's own doc comment already states the correct rule — *"Only when an
author's first submission is COMMENTED (no verdict established yet) does it become
[...]

Issue #100, opened 2026-04-03
Test issue. Specify a color change for the TUI please


verveguy:
SSIA

verveguy:
🏭 **Fabrik — stage: Specify**
*branch: fabrik/issue-100 | commit: 44879999 | 2026-04-03 13:10 UTC*

Now I have a good picture of the TUI. The color definitions are at `tui/model.go:83-97`. The current palette uses:

- `titleStyle` — `"205"` (pink/magenta), bold
- `successStyle` — `"42"` (green)
- `failStyle` — `"196"` (red)
- `dimStyle` — `"243"` (mid-gray)
- `activeStyle` — `"214"` (orange) — used for the "In Progress" section heading
- `selectedStyle` — `"238"` background (dark gray) — selected row highlight
- `borderStyle` — no color (default terminal fg)

The issue request is too vague to specify. I need to know which element(s) to change and what to change them to. I'll write a structured spec with the open questions.

FABRIK_ISSUE_UPDATE_BEGIN
## Summary

Change one or more colors in the Fabrik TUI (`tui/model.go`). The TUI is built with [lipgloss](https://github.com/charmbracelet/lipgloss) and uses a fixed set of ANSI 256-color palette entries for its six named styles. This issue asks for a targeted color update — but the specific element(s) and target color(s) are not yet defined.

## Current Color Map

| Style | Used for | Current color |
|---|---|---|
| `titleStyle` | "fabrik" header title | `205` (pink/magenta), bold |
| `successStyle` | ✓ checkmark in history | `42` (green) |
| `failStyle` | ✗ cross in history | `196` (red) |
| `dimStyle` | timestamps, stats, hints, status text | `243` (mid-gray) |
| `activeStyle` | "In Progress" section heading | `214` (orange) |
| `selectedStyle` | selected-row background highlight | bg `238` (dark gray) |
| `borderStyle` | rounded borders on all three panes | default terminal fg (no color) |

## Requirements

_(To be filled in once open questions are answered.)_

## Scope

- **In scope:** Changing lipgloss color values in `tui/model.go`
- **Out of scope:** Theming system, user-configurable colors, changes to layout or structure

## Open Questions

- [ ] Which TUI element(s) should have their color changed? (e.g., title, borders, active-jobs heading, selected row, success/fail indicators, dim text)
- [ ] What should the new color(s) be? (ANSI 256 code, hex, or descriptive name like "cyan" / "blue")
[...]

Issue #1527, opened 2026-08-10
test(e2e): the two-mode gate's on leg exhausts the hourly GraphQL budget — throttled runs are indistinguishable from failures


arbeithand:
## Problem

The `on` leg of the two-mode e2e gate (`scripts/e2e/run.sh`) cannot complete without
exhausting GitHub's hourly GraphQL allowance, so it cannot currently produce a
trustworthy verdict. This was hit three times during the 0.0.78 pre-release gate on
2026-08-09/10.

Engine log, test bed (`handarbeit/fabrik-test`, token `@arbeithand`):

```
2026-08-10T07:16:36Z [warn] GraphQL rate limit low (19% remaining) — activating rate-limit backoff
2026-08-10T09:15:16Z [warn] GraphQL rate limit low (19% remaining) — activating rate-limit backoff
2026-08-10T10:13:34Z [warn] GraphQL rate limit low (19% remaining) — activating rate-limit backoff
```

Once backoff engages, polling slows enough that board items sit in `Queued` /
`fabrik:awaiting-ci` past the scenarios' wait deadlines. The failures that result are
**all timeouts** — across three separate attempts, not one was an assertion violation:

```
mergetrain_bisect_test.go:60: timed out waiting for #4581 to land (last observed status "Queued", issue state "OPEN")
smoke_test.go:89:            timed out waiting for #4570 to close (last observed: "OPEN")
conjunctive_ci_review_gate_test.go:274: WaitForIssueClosed: transient gh error ... exit status 1 (will retry)
```

The contrast with the `off` leg is the diagnostic:

| leg | result | backoff events |
|---|---|---|
| `off` (4155s) | **69 pass, 0 fail** | **0** |
| `on`, after `off` | 36 pass, 7 fail | 1 |
| `on`, alone, full 4871-point budget | 30 pass, 13 fail | 2 |

The third row is the important one. Starting from a nearly full hourly budget, the `on`
leg still drove itself to 19% remaining **twice**, i.e. it burns more than one hour's
GraphQL allowance and then more again after the reset. This is not a case of the `off`
leg leaving too little behind — merge-train mode's own polling (trial-branch assembly,
integration-PR CI polling, per-SHA check reads) is the cost.

Issue #1219 ("make the suite affordable — board reads via GraphQL, tunable poll base"),
which `scripts/e2e/run.sh`'s header names as the prerequisite for attempting a two-mode
[...]

PR #1540, opened 2026-08-11
fix(github): page REST collection fetchers instead of truncating at page one

### LINKED ISSUE #1539: github: four REST collection fetchers silently truncate at the first page — drove a 315-review Pruefer loop and can corrupt the landing gate
## Problem

Four REST collection fetchers in `github/` issue a single request and treat the first
page as the complete result. Past the page boundary they silently return a partial set
with no error and no truncation signal — indistinguishable from a genuinely complete
answer.

This is live, not theoretical. Pruefer re-reviewed `handarbeit/fabrik#1256` **315 times**
against an unchanged head, roughly every 3.5 minutes for days, until the daemon was
killed manually.

`pruefer/review.go:109` asks "have I already reviewed this head?" via `FetchPRReviews`.
Its own diagnostic logging shows the failure directly:

```
[pr#1256 select] head-SHA check for handarbeit-pruefer[bot]:
    compared=a05573dc…   ← the PR's actual head
    found=2885fb75…      ← newest review it can see
    outcome=proceeding
```

The commit_id histogram of that PR's reviews (oldest → newest) explains `found`:

```
34 distinct SHAs, 1 review each     ← normal operation, head moving
71 × 2885fb75                        ← loop begins
...
92 × a05573dc                        ← current head
```

34 + 71 = 105, so review #100 — the last entry on a `per_page=100` response — lands
inside the `2885fb75` block. Pruefer literally cannot see its own recent reviews.

**It is a ratchet, not a plateau.** Each spurious review grows the collection, pushing the
true latest further beyond the page boundary. Once a PR passes ~100 reviews the condition
is permanent and self-reinforcing; nothing on the Pruefer side can recover.

## Blast radius beyond Pruefer

`FetchPRReviews` is not Pruefer-only. `engine/reviews.go` calls it at lines 137, 771,
1263 and 1394 — including `reviewGateBlocksLanding`, the authoritative **landing** gate.
On a PR past 100 reviews, Fabrik's merge decision is made against a truncated review
list, so a stale verdict can either block a mergeable PR indefinitely or clear a gate
that a newer `CHANGES_REQUESTED` should hold. That is a merge-correctness bug, not a
cost bug.

## The affected family

| function | current request | failure past the boundary |
|---|---|---|
| `FetchPRReviews` (`prs.go:75`) | `?per_page=100`, one page | **confirmed live**; stale review verdict |
| `FetchIssueComments` (`comments.go:116`) | `?per_page=100`, one page | newest comments on a busy issue never seen — comment processing silently stops responding |
| `ListOpenPRs` (`prs.go:798`) | `?state=open&per_page=100`, one page | merge-train member discovery misses PRs |
| `FetchCheckRuns` (`prs.go:579`) | **no `per_page` at all** (GitHub defaults to 30) | worst of the four: also discards the `total_count` the API returns, so >30 checks reads as a complete, all-green set — the CI gate can advance on an incomplete view |

Deliberately excluded: `ListPRs` (`prs.go:735`) is documented as an intentional
"50 most-recently-updated" cap for idempotency detection, and `FetchPRReviewRequests`
(`prs.go:352`) returns an object whose `users` array realistically stays under one page.
Both should be reconsidered only with evidence.

## Requirements

- R1. Each affected fetcher MUST return the complete collection, paging until exhausted.
- R2. Truncation MUST never be silent. If a hard safety bound is reached, the call MUST
  return an error rather than a short list — a partial result presented as complete is
  the actual defect here, and #1527 records the same lesson from the e2e gate.
- R3. `FetchCheckRuns` MUST additionally verify the accumulated count against the
  `total_count` the API reports, and fail loudly on a mismatch.
- R4. Pagination MUST live in one shared helper, not be re-implemented per call site —
  `FetchLabelAppliedAt` and `FetchPRFiles` already hand-roll page loops, and the next
  fetcher added should get this for free.
- R5. Existing single-page callers must be unaffected when a collection fits in one page
  (no extra request when a short page is returned).

## Acceptance

- A1. A PR with >100 reviews returns all of them; a test drives a multi-page fixture and
  asserts the last item is present.
- A2. `FetchCheckRuns` with >30 check runs returns all of them, and a `total_count`
  mismatch produces an error.
- A3. Reaching the page cap returns an error, not a truncated slice.
- A4. A single-page response issues exactly one HTTP request (no speculative second call).
- A5. Non-vacuity: each new test is verified red against the unfixed code, recorded in
  the PR body.

## Notes

Found while diagnosing Pruefer's re-review loop on #1256. Note the irony: #1497 (shipped
in v0.0.78) added three levels of truncation signalling to `FetchPRReviewThreads`
precisely so an incomplete fetch is never read as "nothing more to find" — the function
one file away had the same bug and was missed.

#1496 is **not** a fix for this. It corrected the supersession rule for `COMMENTED`
reviews, which is the right rule applied to a truncated list. Pruefer was running a
binary containing #1496 throughout the loop described above.

arbeithand:
Closes #1539

## What this fixes

Four REST collection fetchers in `github/` issued a single request and treated page one as the whole collection — returning a partial set with no error and no truncation signal, indistinguishable from a complete answer.

**This is a live production failure, not a hypothetical.** Pruefer re-reviewed `handarbeit/fabrik#1256` **315 times** against an unchanged head, roughly every 3.5 minutes for days, until the daemon was killed manually. Its own diagnostic logging shows the frozen comparison:

```
[pr#1256 select] head-SHA check for handarbeit-pruefer[bot]:
    compared=a05573dc…   ← the PR's actual head
    found=2885fb75…      ← newest review it can see
    outcome=proceeding
```

The review histogram explains `found`: 34 distinct SHAs (one review each) followed by a 71-long run of `2885fb75` puts review #100 — the last entry on a `per_page=100` response — inside that stale block. **It is a ratchet, not a plateau:** every spurious review grows the collection and pushes the true latest further beyond the boundary, so past ~100 reviews the condition is permanent and self-reinforcing.

`FetchPRReviews` is not Pruefer-only — `engine/reviews.go` calls it at lines 137, 771, 1263 and 1394, including `reviewGateBlocksLanding`. On a PR past 100 reviews, Fabrik's **authoritative landing gate** made its merge decision against a stale verdict. That is merge correctness, not just wasted spend.

## Changes

**`paginateREST`** (`github/rest.go`) — one shared generic helper rather than a fourth hand-rolled page loop (`FetchLabelAppliedAt` and `FetchPRFiles` already have their own). A page shorter than `restPageSize` terminates, so a single-page collection still costs exactly one request — no speculative probe. Reaching `restMaxPages` (100 pages / 10,000 records) returns an **error**, not the accumulated prefix: silently degrading into a short list is precisely the defect being fixed.

| function | before | after |
|---|---|---|
| `FetchPRReviews` | `per_page=100`, one page | paged; the collapse now sees each author's genuine latest review |
| `FetchIssueComments` | `per_page=100`, one page | paged; comments are oldest-first, so truncation dropped the **newest** — the ones comment processing exists to react to |
| `ListOpenPRs` | `per_page=100` + warn at exactly 100 | paged; the warning is now unnecessary |
| `FetchCheckRuns` | **no `per_page`** (GitHub default 30), `total_count` discarded | paged **and** verified against `total_count`, erroring on mismatch |

`FetchCheckRuns` was the worst of the four. Every CI gate decision (`checkCIGate`, the merge-train trial classifier) rests on that list, so a commit with >30 check runs could read as a complete, all-green set. The count check is an error rather than a warning because a verdict computed from a known-incomplete set is worse than no verdict — the caller cannot tell them apart.

**Deliberately unchanged:** `ListPRs` (documented intentional "50 most-recently-updated" cap for idempotency detection) and `FetchPRReviewRequests` (object response whose `users` array realistically stays under one page). Both should change only with evidence.

## A replaced test

`TestListOpenPRs_WarnsAtCap` pinned the old contract — stop at 100 and log a warning — which this change removes, so its assertion could not survive. Rather than delete it, it is superseded by **`TestListOpenPRs_NeverSilentlyTruncates`**, preserving the underlying intent ("a capped read is never presented as a complete one") and strengthening it to require an error. Its fixture — a server that ignores `page` and returns a full page forever — is exactly the pathology `restMaxPages` bounds, which nothing else covered.

## Non-vacuity (A5)

Every new test was verified red against neutralized code, with the fix reverted one at a time:

| neutralization | observed failure |
|---|---|
| `FetchPRReviews` back to single-page | `collapsed to commit "sha099", want "sha206"` + `served 1 requests, want 3` |
| `total_count` check wrapped in `if false &&` | `expected an error when total_count exceeds the collected set, got nil` |
| `FetchCheckRuns` `per_page` dropped (default 30) | `collected 30 of 111 reported check runs` — caught by the independent `total_count` guard |
| `FetchIssueComments` back to single-page | `got 100 comments, want 105` |
| `ListOpenPRs` back to single-page | `got 100 open PRs, want 103` |

The third is worth noting: with pagination broken, the `total_count` verification caught it independently. The two protections are not redundant.

## Testing

- `go build ./...`, `go vet ./...`, `gofmt` clean on all touched files.
- `go test -race ./...` fully green.
- Pre-existing `TestFetchCheckRuns_Success` / `TestFetchCheckRuns_Empty` pass unmodified.

One unrelated pre-existing `gofmt` drift exists in `github/types.go` on `main` (alignment in `MergeQueueEntry`). Deliberately left alone rather than mixed into this change.

## Note on #1496

#1496 is **not** a fix for this. It corrected the supersession rule for `COMMENTED` reviews — the right rule, applied to a truncated list. Pruefer was running a binary containing #1496 throughout the 315-review loop described above.

### DIFF
--- a/github/comments.go
+++ b/github/comments.go
@@ -113,8 +113,7 @@ func (c *Client) UpdateIssueBody(owner, repo string, issueNumber int, body strin
 // "/pruefer review" comment commands and apply 👀/🚀 reaction idempotency.
 // Returns nil, nil on 404.
 func (c *Client) FetchIssueComments(owner, repo string, issueNumber int) ([]Comment, error) {
-	apiURL := fmt.Sprintf("%s/repos/%s/%s/issues/%d/comments?per_page=100", c.baseURL, owner, repo, issueNumber)
-	var raw []struct {
+	type rawComment struct {
 		ID        int       `json:"id"`
 		Body      string    `json:"body"`
 		CreatedAt time.Time `json:"created_at"`
@@ -132,7 +131,15 @@ func (c *Client) FetchIssueComments(owner, repo string, issueNumber int) ([]Comm
 			Eyes     int `json:"eyes"`
 		} `json:"reactions"`
 	}
-	if err := c.restGetJSON(apiURL, &raw); err != nil {
+	// Paginated (#1539): GitHub returns issue comments oldest-first, so reading
+	// only page one drops the NEWEST comments on any thread past restPageSize —
+	// exactly the ones comment processing exists to react to. A busy issue would
+	// silently stop responding to new instructions.
+	raw, err := paginateREST[rawComment](c, fmt.Sprintf("comments for %s/%s#%d", owner, repo, issueNumber), func(page int) string {
+		return fmt.Sprintf("%s/repos/%s/%s/issues/%d/comments?per_page=%d&page=%d",
+			c.baseURL, owner, repo, issueNumber, restPageSize, page)
+	})
+	if err != nil {
 		if errors.Is(err, ErrNotFound) {
 			return nil, nil
 		}
--- a/github/labels.go
+++ b/github/labels.go
@@ -147,12 +147,18 @@ func (c *Client) RemoveLabelFromIssue(owner, repo string, issueNumber int, label
 // FetchLabelAppliedAt returns the time when labelName was last applied to the
 // given issue, using the GitHub issue events API. Returns time.Time{} (zero)
 // without error if the event is not found (fail-open: timeout will not fire).
-// Pages through all events to find the most recent application.
+// Pages through all events to find the most recent application, bounded by
+// restMaxPages (#1539 review finding 1) — this loop was previously
+// `for page := 1; ; page++` with no cap, so a server that never returns an
+// empty page would spin forever. The bound is fail-open, consistent with the
+// rest of this function: exhausting it returns the newest application seen so
+// far rather than an error, since a missed timestamp only means the timeout
+// this feeds does not fire — never a wrong action.
 func (c *Client) FetchLabelAppliedAt(owner, repo string, issueNumber int, labelName string) (time.Time, error) {
 	var applied time.Time
-	for page := 1; ; page++ {
-		apiURL := fmt.Sprintf("%s/repos/%s/%s/issues/%d/events?per_page=100&page=%d",
-			c.baseURL, owner, repo, issueNumber, page)
+	for page := 1; page <= restMaxPages; page++ {
+		apiURL := fmt.Sprintf("%s/repos/%s/%s/issues/%d/events?per_page=%d&page=%d",
+			c.baseURL, owner, repo, issueNumber, restPageSize, page)
 		var events []struct {
 			Event     string `json:"event"`
 			CreatedAt string `json:"created_at"`
--- a/github/pagination_test.go
+++ b/github/pagination_test.go
@@ -1,0 +1,246 @@
+package github
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"strconv"
+	"strings"
+	"testing"
+)
+
+// Tests for #1539: REST collection fetchers must page until exhausted rather
+// than treating page one as the whole collection.
+//
+// Each multi-page test builds a fixture strictly larger than restPageSize and
+// asserts the LAST record is present — the first page alone can never satisfy
+// that, which is what makes these non-vacuous against the pre-fix code.
+
+// pagedArrayServer serves a bare JSON array endpoint that honours page/per_page.
+// items is the full collection; each element is marshalled as-is. It records how
+// many requests it served so single-page tests can assert no speculative probe.
+func pagedArrayServer(t *testing.T, path string, items []map[string]any, requests *int) (*httptest.Server, *Client) {
+	t.Helper()
+	mux := http.NewServeMux()
+	mux.HandleFunc(path, func(w http.ResponseWriter, r *http.Request) {
+		*requests++
+		page, _ := strconv.Atoi(r.URL.Query().Get("page"))
+		if page < 1 {
+			page = 1
+		}
+		per, _ := strconv.Atoi(r.URL.Query().Get("per_page"))
+		if per < 1 {
+			per = 30
+		}
+		start := (page - 1) * per
+		if start > len(items) {
+			start = len(items)
+		}
+		end := start + per
+		if end > len(items) {
+			end = len(items)
+		}
+		if err := json.NewEncoder(w).Encode(items[start:end]); err != nil {
+			t.Errorf("encoding page %d: %v", page, err)
+		}
+	})
+	srv := httptest.NewServer(mux)
+	t.Cleanup(srv.Close)
+	return srv, NewClientWithBaseURL("test-token", srv.URL)
+}
+
+// TestFetchPRReviews_PagesBeyondFirstPage is the direct regression test for the
+// production failure: on handarbeit/fabrik#1256 the newest reviews sat past the
+// first page, so FetchPRReviews collapsed each author to a stale verdict and
+// Pruefer re-reviewed an unchanged head 315 times.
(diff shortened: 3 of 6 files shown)


arbeithand:
## Independent review round — findings addressed in 063aa692

No review bots are available for this PR, so it was reviewed by a separate agent with a clean context, explicitly instructed not to treat the commit message, PR body, or code comments as evidence of correctness. It found a hang vector the original diff missed.

### Fixed

**1. Unbounded sibling loops (most severe).** `FetchPRFiles` (`prs.go`) and `FetchLabelAppliedAt` (`labels.go`) predate `paginateREST` and loop `for page := 1; ; page++` with **no cap** — the exact pathology `restMaxPages` was added to bound. A server that never returns a short page spins forever while the accumulator grows. The original diff fixed four fetchers and left the two closest in shape unaddressed, which also undercut its own framing.

Both are now bounded. `FetchPRFiles` fails loud; `FetchLabelAppliedAt` stays fail-open, consistent with the rest of that function — a missed timestamp only means a timeout does not fire, never a wrong action.

**2. `total_count` could false-positive.** Independently flagged before the review landed, and the review added a second path to it (TOCTOU as CI adds check runs between paged requests). Guarding on `>= 0` made an *absent* `total_count` — which decodes to `0` — indistinguishable from a genuine zero, so a deployment omitting the field would hard-fail every CI gate on every commit, permanently. Real github.com always sends it (confirmed against the captured recording in `testdata`), but v0.0.78 shipped GHES support with no equivalent recording. Now fails open on absence, closed only on a real positive disagreement. Also removes the dead `totalCount := -1` sentinel, which guarded a state the loop cannot reach.

**3. Doc overstated the request count.** "A single-page collection costs exactly one request" is true only *below* `restPageSize`; exactly `restPageSize` costs a second, empty request. Wording corrected, and the boundary — which no test touched, every fixture being `+3`/`+5`/`+11`/`*2+7` — is now covered by `TestFetchPRReviews_ExactlyOnePageFullStillTerminates`.

**4. Documented limitation:** a genuine collection of exactly `restMaxPages*restPageSize` (10,000) records is indistinguishable from a non-terminating server and is reported as the latter.

### A gap the review did not catch, found while fixing it

Removing the new `FetchPRFiles` bound broke **no** test — it hung. The guard added in (1) was itself unverified. `TestFetchPRFiles_BoundedAgainstEndlessPages` now covers it.

### One behavior change worth calling out

`FetchPRFiles` now terminates on a *short* page rather than only an *empty* one — same data, one fewer round trip per call, consistent with `paginateREST`, and worth having given #1527. `TestFetchPRFiles_MultiPage`'s request-count assertion moves 3 → 2 with a comment recording why, rather than being quietly retuned.

### Deliberately not addressed

- `FetchCheckRuns` still hand-rolls its page loop; it needs the raw `total_count` envelope `paginateREST` does not expose. Real DRY risk, but parameterizing the helper is a larger change than this fix warrants.
- A 404 surfacing on page 2+ still collapses to `nil, nil`, discarding earlier pages. Low likelihood, and no worse than pre-fix.

Both are follow-ups, not silent omissions.

### Verification

`go build ./...`, `go vet ./...`, `gofmt` clean, `go test -race ./...` fully green. Non-vacuity re-confirmed after the amendments: neutralizing the `total_count` guard still reddens its test.

PR #1547, opened 2026-08-11
test(e2e): fail loud on GraphQL budget exhaustion, tighten the two-mode gate's "on" leg

### LINKED ISSUE #1527: test(e2e): the two-mode gate's on leg exhausts the hourly GraphQL budget — throttled runs are indistinguishable from failures

arbeithand:
Closes #1527

## What this does

The two-mode e2e gate's `on` leg (`scripts/e2e/run.sh`) couldn't complete without exhausting GitHub's hourly GraphQL allowance, producing a wall of indistinguishable test timeouts instead of a trustworthy verdict — hit three times during the 0.0.78 pre-release gate. This PR ships both required workstreams together, R1 (cost reduction) and R2 (fail-loud detection), plus R3's confirmation logging and A3's documentation.

**R3 — confirmed the cost driver (static analysis, this issue's own Research):** not merge-train's own CI polling (that's REST, a separate budget), but the ADR-1270/ADR-1208 settle-scan pattern — `settleAwaitingCIScan` and, merge-train-exclusively, `settleQueuedReviewFindings` — each doing an unconditional, no-cooldown GraphQL deep-fetch per matching item per poll. That per-item semantics is load-bearing for ADR-1208's correctness guarantee and is left untouched.

**R1 — cost reduction:** added `E2E_PARALLEL_ON` (default 2, half of `E2E_PARALLEL`'s default 4), applied only to the default two-mode gate's `on` leg. Fewer concurrent scenarios → smaller Queued/awaiting-CI population → less settle-scan cost, without weakening the scans' own per-item correctness guarantees. `off` and any forced single-mode run keep using `E2E_PARALLEL` unchanged.

**R2 — fail-loud detection, independent of whether R1 fully eliminates the risk:** after each leg, `run.sh` scans the bed's `fabrik.log` (scoped to just that leg's own output) for the engine's one-shot `"...activating rate-limit backoff"` line. On a match — regardless of the leg's pass/fail exit code — it prints an unambiguous `RUN INVALID` banner and exits with a new dedicated code (`3`, distinct from `go test`'s propagated `1`), short-circuiting any remaining leg.

The detection logic went through two follow-up fixes during review, both from real defects found in the byte-offset-based first cut:
1. The offset was originally captured *before* the bed restart, leaving the restart window (including the freshly-restarted engine's very first poll) unscanned.
2. Moving the capture point wasn't enough: `engine/poll.go`'s `Run()` truncates `fabrik.log` (`O_TRUNC`) on every engine startup, and the restart always launches a fresh process — so a pre-restart offset is measured against a file the restart then truncates out from under it, making `tail -c +N` silently return nothing for the rest of the leg. The fix (commit `2af83849`) replaced offset-tracking entirely with a whole-file scan: since the restart always truncates the log, the file already contains only this leg's own content the moment the restart completes, so there's nothing to scope past.

That whole-file-scan detection logic is now covered by an automated, non-vacuous regression test (`scripts/e2e/backoff_detection_test.sh`, added in response to review) — see "How to test" below.

**A3 groundwork:** each leg is now bracketed with `gh api rate_limit` calls, so real GraphQL cost is reported automatically on every future run. `queued_review_settle.go` also gets the cache-hit-vs-deep-fetch attribution log line `ci_settle.go` already had, closing a logging-parity gap.

`tests/e2e/README.md`'s GraphQL budget section is extended with the root-cause analysis, the shipped mitigations, and a placeholder cost table with the exact reproduction command (obtaining real numbers requires a multi-hour, real-GitHub-mutating run against the shared bed, which this PR does not launch unilaterally).

## Key changes

- `engine/queued_review_settle.go` — additive cache-hit/deep-fetch attribution log line (no behavior change)
- `scripts/e2e/run.sh` — `E2E_PARALLEL_ON`, per-leg `gh api rate_limit` bracketing, post-leg rate-limit-backoff log scan (via a standalone `detect_rate_limit_backoff` function) with `RUN INVALID` banner + exit code 3, a sourcing guard so the detection function is independently testable, expanded header comment
- `scripts/e2e/backoff_detection_test.sh` — new: automated, non-vacuous regression test for the backoff-detection logic (5 cases, including a neutralization case against the historical broken byte-offset approach)
- `tests/e2e/README.md` — new "two-mode gate's 'on' leg" subsection in the GraphQL budget section

## How to test

- `go test ./engine/... -race -run 'QueuedReview'` — all existing `settleQueuedReviewFindings` tests pass, new log line confirmed in test output
- `bash -n` and `shellcheck` on both `scripts/e2e/run.sh` and `scripts/e2e/backoff_detection_test.sh` — clean
- `./scripts/e2e/backoff_detection_test.sh` — automated, non-vacuous coverage for R2's detection logic, added in response to review. Exercises the *actual* `detect_rate_limit_backoff` function (sourced from `run.sh`, not a reimplementation) against 5 fixtures:
  1. no backoff activity → no match
  2. the companion per-poll "is low..." line alone (deliberately excluded — see header comment) → no match
  3. the one-shot activation line, no truncation → match
  4. **the field regression itself**: the activation line logged only *after* a mid-leg `fabrik.log` truncation (simulating the engine restart's `O_TRUNC`) → match, confirming the current whole-file-scan approach is not susceptible to the offset-based failure mode described in review
  5. **neutralization**: the same truncated fixture from case 4, run through a deliberately-reintroduced copy of the historical byte-offset approach (`tail -c "+$((log_offset+1))"` against an offset captured pre-truncation) → does **not** match, proving case 4 exercises the real regression and that a future reintroduction of offset-based scanning would be caught by this test rather than only discovered on a live multi-hour bed run

  All 5 pass; exit code 0.
- Full `go test ./... -race` passes except one pre-existing, unrelated flake (`cmd.TestExecute_ConfigYAMLApplied`, a SIGINT-timing test unrelated to this change — confirmed via `git diff --name-only origin/main...HEAD` that `cmd/` is untouched)

No live two-mode gate run was executed as part of this PR (multi-hour, real-GitHub-mutating operation against the shared bed) — A3's measured-cost table is a placeholder with reproduction instructions for the next person to fill in.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,3 +55,17 @@ jobs:
       # network call. There is no TestMain, so nothing runs. Takes <1s.
       - name: compile e2e suite (not run)
         run: go test -tags e2e -run '^$' ./tests/e2e/...
+
+      # scripts/e2e/run.sh's own header states this script exists so a future
+      # edit that regresses R2's rate-limit-backoff detection (#1527) — e.g.
+      # reintroducing the byte-offset scan that fabrik.log's O_TRUNC-on-startup
+      # silently defeated — fails a check instead of only being caught by a
+      # real, multi-hour release-gate run. That guarantee only holds if this
+      # actually runs somewhere; wired in here per #1547's review. Needs no
+      # test bed, token, or network: it sources run.sh for the function
+      # definitions only (guarded so sourcing doesn't trigger a gate run) and
+      # drives fixtures entirely via mktemp. On a bare runner with no
+      # ~/dev/fabrik-test, run.sh's TEST_BED/BED_TOKEN resolution degrades to
+      # a stderr warning rather than failing under set -euo pipefail.
+      - name: backoff-detection regression test (#1527)
+        run: bash scripts/e2e/backoff_detection_test.sh
--- a/engine/queued_review_settle.go
+++ b/engine/queued_review_settle.go
@@ -64,6 +64,21 @@ func (e *Engine) settleQueuedReviewFindings(board *gh.ProjectBoard) {
 				continue
 			}
 
+			// Mirrors settleAwaitingCIScan's equivalent pre-check log line
+			// (ci_settle.go) — purely diagnostic, doesn't gate the call.
+			// FetchItemDetails below goes through the same e.readClient
+			// (CacheImpl when caching is enabled), whose own internal
+			// freshness check (boardcache.go's cacheIsStale) decides
+			// cache-hit vs. real GraphQL fetch identically regardless of
+			// caller. Without this line, an operator attributing this scan's
+			// GraphQL cost (#1527) would see per-item log entries with no
+			// visibility into whether each fetch was a cache hit or a live
+			// API call.
+			if c := e.cache(); c != nil && !c.IsPaused() && c.IsItemCacheFresh(item.Repo, item.Number, item.UpdatedAt) {
+				e.logf(item.Number, "queued-review-settle", "reading details from cache\n")
+			} else {
+				e.logf(item.Number, "queued-review-settle", "deep-fetching details from GitHub\n")
+			}
 			if err := e.readClient.FetchItemDetails(&item); err != nil {
 				e.logf(item.Number, "queued-review-settle", "could not deep-fetch item details: %v — will retry next poll\n", err)
 				continue
--- a/scripts/e2e/backoff_detection_test.sh
+++ b/scripts/e2e/backoff_detection_test.sh
@@ -1,0 +1,131 @@
+#!/usr/bin/env bash
+# scripts/e2e/backoff_detection_test.sh — regression coverage for R2's
+# (#1527) rate-limit-backoff detection in scripts/e2e/run.sh.
+#
+# R2 is a release-gating guarantee ("a throttled run must fail loudly and
+# explicitly") with no automated coverage prior to this script — see PR
+# #1547's review thread, where a byte-offset-based version of the detector
+# shipped, was defeated in the field by fabrik.log's O_TRUNC-on-startup
+# behavior, and was fixed to a whole-file scan (see run.sh's
+# detect_rate_limit_backoff for the full account). This script exists so
+# that regression can't recur silently: it exercises the *actual*
+# detect_rate_limit_backoff function from run.sh against fixtures, plus a
+# deliberately-reintroduced copy of the historical broken approach, so a
+# future edit that reintroduces byte-offset tracking (or otherwise narrows
+# the scan) fails this script instead of only being caught by a real,
+# multi-hour bed run.
+#
+# Usage: scripts/e2e/backoff_detection_test.sh
+# Exit 0 if all assertions pass, 1 otherwise.
+
+set -euo pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+
+# Source run.sh for its function definitions only — the sourcing guard at
+# the bottom of run.sh (BASH_SOURCE[0] == $0) prevents this from triggering
+# an actual gate run. Some of run.sh's top-level setup still executes on
+# source (TEST_BED/BED_TOKEN resolution) but is read-only and degrades
+# gracefully (a missing $TEST_BED/.env just warns), so it's safe here.
+# shellcheck source=/dev/null
+source "$REPO_ROOT/scripts/e2e/run.sh"
+
+FAILED=0
+TMP_LOG="$(mktemp)"
+trap 'rm -f "$TMP_LOG"' EXIT
+
+check() {
+  local desc="$1"
+  local expect_match="$2" # "match" or "no-match"
+  if detect_rate_limit_backoff "$TMP_LOG"; then
+    local actual="match"
+  else
+    local actual="no-match"
+  fi
+  if [ "$actual" = "$expect_match" ]; then
+    echo "PASS: $desc (expected $expect_match, got $actual)"
+  else
+    echo "FAIL: $desc (expected $expect_match, got $actual)"
+    FAILED=1
+  fi
+}
+
+# --- Case 1: no backoff activity at all -> no match ---
+{
+  echo "2026-08-11T09:00:00Z [info] poll cycle complete"
+  echo "2026-08-11T09:00:30Z [info] poll cycle complete"
+} > "$TMP_LOG"
+check "no backoff activity" "no-match"
+
+# --- Case 2: only the companion per-poll "is low" line -> no match ---
+# (R2 deliberately excludes this spammy per-poll line, which fires on every
+# poll while low even without ever crossing the 20% activation threshold —
+# see run.sh's header comment.)
+{
+  echo "2026-08-11T09:00:00Z [warn] GraphQL rate limit is low (25% remaining) — consider reducing poll frequency"
+} > "$TMP_LOG"
+check "per-poll 'is low' line alone (not the activation line)" "no-match"
+
+# --- Case 3: the real one-shot activation line, no truncation -> match ---
+{
+  echo "2026-08-11T09:00:00Z [warn] GraphQL rate limit low (19% remaining) — activating rate-limit backoff"
+} > "$TMP_LOG"
+check "activation line present, no truncation" "match"
+
+# --- Case 4: the actual field regression — a restart truncates the log
+# between "before" and "after", and the activation line is logged only
+# after truncation. This is the exact scenario that defeated the
+# byte-offset version of the detector (a pre-restart offset measured
+# against a file the restart then truncates out from under it). The
+# CURRENT detector (whole-file scan) must still catch this. ---
+{
+  # Simulate a substantial pre-restart log (mirrors the ~640KB real bed log
+  # cited in PR #1547's review thread) ...
+  for _ in $(seq 1 50); do
+    echo "2026-08-11T08:00:00Z [info] poll cycle complete, nothing to do"
+  done
+} > "$TMP_LOG"
+# ... then simulate the engine restart's O_TRUNC (engine/poll.go's Run())
+# truncating the file to empty, and the freshly-restarted engine logging the
+# activation line as one of its very first polls.
+: > "$TMP_LOG"
+echo "2026-08-11T09:15:16Z [warn] GraphQL rate limit low (19% remaining) — activating rate-limit backoff" >> "$TMP_LOG"
+check "activation line logged only after a mid-leg truncation (the field regression)" "match"
+
+# --- Case 5 (neutralization): the SAME fixture as Case 4, run through the
+# historical broken byte-offset approach instead of the current
+# detect_rate_limit_backoff. This proves the fixture actually exercises the
+# regression (not just a fixture that happens to always match), and that
+# the byte-offset approach a future edit might reintroduce would indeed be
+# caught failing here. ---
+{
+  # log_offset captured BEFORE the restart, against the pre-truncation file
+  # (mirrors PR #1547's original, defeated implementation).
+  pre_restart_log="$(mktemp)"
+  for _ in $(seq 1 50); do
+    echo "2026-08-11T08:00:00Z [info] poll cycle complete, nothing to do"
+  done > "$pre_restart_log"
+  log_offset=$(wc -c < "$pre_restart_log")
+  rm -f "$pre_restart_log"
+
+  # The restart truncates the real log out from under that offset, then the
+  # freshly-restarted engine logs the activation line — same as Case 4's
+  # $TMP_LOG, which is still in that post-truncation state here.
+  if tail -c "+$((log_offset + 1))" "$TMP_LOG" 2>/dev/null | grep -q 'activating rate-limit backoff'; then
+    legacy_result="match"
+  else
+    legacy_result="no-match"
+  fi
+}
+if [ "$legacy_result" = "no-match" ]; then
+  echo "PASS: neutralization — historical byte-offset approach fails to detect the same fixture Case 4 caught (confirms Case 4 exercises the real regression, and that this approach was correctly replaced)"
+else
+  echo "FAIL: neutralization — historical byte-offset approach unexpectedly detected the fixture; Case 4 may not be exercising the regression it claims to"
+  FAILED=1
+fi
+
+if [ "$FAILED" -ne 0 ]; then
+  echo "=== backoff_detection_test.sh: FAILED ==="
+  exit 1
+fi
+echo "=== backoff_detection_test.sh: all checks passed ==="
--- a/scripts/e2e/run.sh
+++ b/scripts/e2e/run.sh
@@ -110,8 +110,12 @@
 # separate 0.0.78 pre-release gate runs, all of whose failures were timeouts,
 # not assertion violations). To make a throttled run fail loudly instead:
 # after each leg's suite invocation, this script scans the bed's own
-# fabrik.log (at $ENGINE_LOG, scoped to only the bytes written during that
-# leg) for the engine's one-shot rate-limit-backoff-activation line
+# fabrik.log (at $ENGINE_LOG, scoped to the bytes written since just before
+# that leg's own bed-restart step — not just since its suite invocation, so
+# a backoff activation logged by the freshly-restarted engine's very first
+# poll, before the suite even starts, is still caught rather than falling
+# into a blind spot between legs) for the engine's one-shot
+# rate-limit-backoff-activation line
 # ("...activating rate-limit backoff", engine/poll.go — NOT the companion
 # per-poll "...is low (...) consider reducing poll frequency" line, which
 # fires on every poll while low and would over-trigger on a run that dipped
@@ -303,6 +307,22 @@ switch_and_run() {
   local mode="$1"
   local parallel="$2"
   shift 2
+
+  # Captured BEFORE the restart step below (not just before the suite
+  # invocation) so the post-leg backoff scan's window covers the restart
+  # itself and the freshly-restarted engine's very first poll — a fresh
+  # process has no memory of already having activated backoff, so if the
+  # account's real (unreset-by-restart) GraphQL remaining is already at or
+  # near the 20% hysteresis threshold, that first poll's own bootstrap probe
+  # can immediately log the activation line before the suite has even
+  # started. Capturing the offset only after the restart (as an earlier
+  # version of this script did) left exactly that window — plus the idle
+  # gap since the previous leg's own scan already ran — unscanned by either
+  # leg. See engine/poll.go's rate-limit-low log line and
+  # engine/terminal.go's runProbeAndDeepFetch bootstrap probe.
+  local log_offset
+  log_offset=$(wc -c < "$ENGINE_LOG" 2>/dev/null || echo 0)
+
   echo "== switching test bed to FABRIK_MERGE_TRAIN=${mode} =="
   # KNOWN GAP: this restart step has none of the classification/auto-teardown
   # machinery below — it's a plain `-v` (non-`-json`) run with its own fixed
@@ -322,20 +342,17 @@ switch_and_run() {
   local jsonlog="${TMPDIR:-/tmp}/fabrik-e2e-${mode}-$$.json"
   local rc=0
 
-  # Snapshot the bed's GraphQL budget and this leg's position in fabrik.log
-  # before the suite runs — the former lets the post-leg report show this
-  # leg's actual GraphQL cost (A3, #1527), the latter scopes the post-leg
-  # backoff scan below to only what this leg itself wrote, so a prior leg's
-  # (or a prior invocation's) backoff event can never cause a false positive
-  # here. Both guarded with `|| echo ...`/`2>/dev/null || echo 0` so a
-  # transient `gh` failure or a not-yet-existing log file degrades to a
-  # skipped report rather than aborting the script under `set -e`.
-  local budget_before budget_after log_offset
+  # Snapshot the bed's GraphQL budget right before the suite runs (not
+  # log_offset above, which is deliberately captured earlier — see its own
+  # comment) — this is purely for the A3 cost report, which should reflect
+  # the suite's own consumption, not the restart step's. Guarded with
+  # `|| echo ...` so a transient `gh` failure degrades to a skipped report
+  # rather than aborting the script under `set -e`.
+  local budget_before budget_after
   budget_before=""
   if [ -n "$BED_TOKEN" ]; then
     budget_before=$(GH_TOKEN="$BED_TOKEN" gh api rate_limit --jq '.resources.graphql.remaining' 2>/dev/null || echo "")
   fi
-  log_offset=$(wc -c < "$ENGINE_LOG" 2>/dev/null || echo 0)
 
   E2E_TRAIN_MODE="$mode" go test -tags=e2e -json -count=1 -timeout "$TIMEOUT" -parallel "$parallel" \
       ./tests/e2e/... "$@" 2>&1 \
--- a/scripts/e2e/run.sh
+++ b/scripts/e2e/run.sh
@@ -130,9 +130,13 @@
 # zero new API calls against the very budget being protected, and the log
 # line already fires at the moment of interest. Each leg is also bracketed
 # with a `gh api rate_limit` call before/after (a REST metadata call, free
-# against the GraphQL budget) to report the leg's actual consumed points —
-# this is a complementary, separate purpose (durable cost visibility, A3 of
-# #1527) from the log-scrape's fail-loud detection.
+# against the GraphQL budget), scoped to the bed's own FABRIK_TOKEN (read
+# from $TEST_BED/.env into $BED_TOKEN, mirroring reset.sh's gh_() token-
+# scoping precedent — the ambient `gh` CLI's own auth is very likely a
+# different identity than the bed's @arbeithand token and would silently
+# measure the wrong account's budget) to report the leg's actual consumed
+# points — this is a complementary, separate purpose (durable cost
+# visibility, A3 of #1527) from the log-scrape's fail-loud detection.
 #
 # Prerequisites (one-time setup):
 #   - ~/dev/fabrik-test/ exists with .env (FABRIK_TOKEN for @arbeithand)
@@ -151,6 +155,19 @@ cd "$REPO_ROOT"
 # (R2, #1527) to scope its scan to this bed's own fabrik.log.
 TEST_BED="${FABRIK_TEST_DIR:-$HOME/dev/fabrik-test}"
 ENGINE_LOG="$TEST_BED/.fabrik/fabrik.log"
+
+# The bed's own PAT (same FABRIK_TOKEN reset.sh reads and scopes its gh_()
+# helper to) — the A3 budget report below must measure the bed's own
+# GraphQL consumption, never whatever identity the ambient `gh` CLI happens
+# to be authenticated as in the invoking shell, which is very likely a
+# different account than the bed's @arbeithand token and would silently
+# report an unrelated budget. A missing/unreadable token only degrades the
+# budget report (skipped, warned) — unlike reset.sh, nothing else in this
+# script depends on it, so it's not fatal here.
+BED_TOKEN=$( { grep '^FABRIK_TOKEN=' "$TEST_BED/.env" 2>/dev/null | head -1 | cut -d= -f2-; } || echo "")
+if [ -z "$BED_TOKEN" ]; then
+  echo "warning: could not read FABRIK_TOKEN from $TEST_BED/.env — GraphQL budget reporting will be skipped" >&2
+fi
 
 # Distinct exit code for a run invalidated by GraphQL budget exhaustion — see
 # "GraphQL budget exhaustion detection" in the header comment above.
@@ -314,7 +331,10 @@ switch_and_run() {
   # transient `gh` failure or a not-yet-existing log file degrades to a
   # skipped report rather than aborting the script under `set -e`.
   local budget_before budget_after log_offset
-  budget_before=$(gh api rate_limit --jq '.resources.graphql.remaining' 2>/dev/null || echo "")
+  budget_before=""
+  if [ -n "$BED_TOKEN" ]; then
+    budget_before=$(GH_TOKEN="$BED_TOKEN" gh api rate_limit --jq '.resources.graphql.remaining' 2>/dev/null || echo "")
+  fi
   log_offset=$(wc -c < "$ENGINE_LOG" 2>/dev/null || echo 0)
 
   E2E_TRAIN_MODE="$mode" go test -tags=e2e -json -count=1 -timeout "$TIMEOUT" -parallel "$parallel" \
@@ -323,7 +343,10 @@ switch_and_run() {
     | { jq -R -r 'fromjson? // empty | select(.Action=="output") | .Output' 2>/dev/null || true; } \
     || rc=$?
 
-  budget_after=$(gh api rate_limit --jq '.resources.graphql.remaining' 2>/dev/null || echo "")
+  budget_after=""
+  if [ -n "$BED_TOKEN" ]; then
+    budget_after=$(GH_TOKEN="$BED_TOKEN" gh api rate_limit --jq '.resources.graphql.remaining' 2>/dev/null || echo "")
+  fi
   if [ -n "$budget_before" ] && [ -n "$budget_after" ]; then
     if [ "$budget_after" -le "$budget_before" ]; then
       echo "== GraphQL budget (leg: ${mode}): ${budget_before} -> ${budget_after} remaining (consumed $((budget_before - budget_after)) pts) =="
--- a/scripts/e2e/run.sh
+++ b/scripts/e2e/run.sh
@@ -110,11 +110,19 @@
 # separate 0.0.78 pre-release gate runs, all of whose failures were timeouts,
 # not assertion violations). To make a throttled run fail loudly instead:
 # after each leg's suite invocation, this script scans the bed's own
-# fabrik.log (at $ENGINE_LOG, scoped to the bytes written since just before
-# that leg's own bed-restart step — not just since its suite invocation, so
-# a backoff activation logged by the freshly-restarted engine's very first
-# poll, before the suite even starts, is still caught rather than falling
-# into a blind spot between legs) for the engine's one-shot
+# fabrik.log (at $ENGINE_LOG) in full — not from a captured byte offset. Each
+# leg's own bed-restart step (TestSwitchTrainMode, via StopFabrikTestBed +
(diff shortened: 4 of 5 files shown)


arbeithand:
*branch: fabrik/issue-1527 | commit: 1c892daf | main: ccf7971072bae615d2dde5c7bd4df853c5362ce8 | 2026-08-11 12:56 UTC*

Everything is clean, committed, and pushed. Now emitting the PR creation marker.


Shipped both required workstreams: R1 (cost reduction via `E2E_PARALLEL_ON`, capping the "on" leg's concurrency to shrink the ADR-1270/ADR-1208 settle-scan population Research identified as the actual GraphQL cost driver) and R2 (fail-loud detection — `run.sh` now scans `fabrik.log` per leg for the rate-limit-backoff-activation line and exits with a dedicated code 3 on a match, distinct from a normal test failure). Also added R3 confirmation logging, extended `tests/e2e/README.md` per A3 with a placeholder cost table (a live multi-hour bed run wasn't launched unilaterally), and verified detection logic against synthetic log fixtures. Build/vet/shellcheck clean; full test suite passes except one pre-existing, unrelated flake in `cmd/`.

---
Used 72/100 turns, 9.1M input (136 raw + 9.0M cache-read + 171k cache-write) / 35k output tokens.

arbeithand:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-1527 | commit: ab3e3bdf | main: ccf7971072bae615d2dde5c7bd4df853c5362ce8 | 2026-08-11 13:12 UTC*

Everything looks solid. The PR is ready for human review.

## Review Findings

### Fixed
- **Issue**: `run.sh`'s new `gh api rate_limit` calls (A3 budget report) used whatever identity the ambient `gh` CLI happened to be authenticated as in the invoking shell, not the test bed's own `@arbeithand` `FABRIK_TOKEN` — the identity the engine actually polls GitHub as. This would silently measure an unrelated account's budget on any machine where the two differ, defeating A3's whole purpose. **Fix**: read `FABRIK_TOKEN` from `$TEST_BED/.env` into `$BED_TOKEN` (mirroring `reset.sh`'s existing `gh_()` precedent) and scope both `gh api rate_limit` calls to it via `GH_TOKEN`.
- **Issue**: The `BED_TOKEN` read itself (`var=$(pipeline)`) would abort the entire script under `set -e`/`pipefail` if `.env` was missing or lacked a `FABRIK_TOKEN=` line — verified this empirically — silently defeating the intended "degrade gracefully, just skip the budget report" behavior. **Fix**: wrapped the pipeline in `{ ... } || echo ""`; verified all three cases (missing file, file without the key, file with the key) now degrade correctly instead of aborting.
- **Issue** (flagged by handarbeit-pruefer's review): the fail-loud backoff scan's `log_offset` was captured *after* the bed restart completed, leaving the restart window — including the freshly-restarted engine's very first poll, whose bootstrap probe can immediately log a backoff activation if the account's real (restart-doesn't-reset-it) GraphQL remaining is already near the 20% threshold — unscanned by either leg's detection. **Fix**: moved `log_offset` capture to the top of `switch_and_run()`, before the restart step, while leaving `budget_before` (a separate A3 cost-reporting concern) captured immediately before the suite invocation as before.

### Verified
- `gofmt -l engine/queued_review_settle.go` clean
- `go test -race ./engine/...` and full `go test ./...`: only failure is `TestInvokeClaude_FabrikEnvVarsInjected`, a pre-existing environmental flake caused by this very Review session's own ambient `FABRIK_PR=1527` env var — confirmed identical failure occurs with this PR's changes stashed out against the prior commit, so not a regression
- `bash -n scripts/e2e/run.sh` syntax clean after both fixes
- `TestSettleQueuedReviewFindings_*` (9 tests) pass with the new attribution log line firing correctly
- Rebased/up to date with `origin/main`; no external bot feedback remains unaddressed

Reviewed the two-mode e2e gate's GraphQL cost/fail-loud-detection work. Fixed two bugs: the A3 `gh api rate_limit` calls were measuring the wrong GitHub identity's budget (ambient `gh` auth instead of the bed's own token) and would abort the whole script under `set -e` if the token was unreadable; and closed a restart-window blind spot in R2's backoff-activation scan (flagged by Pruefer's review) where a freshly-restarted engine's first poll could log activation before either leg's detection window began. Build/vet/tests clean; PR is ready for human review.

---
Used 86/100 turns, 11.5M input (169 raw + 11.4M cache-read + 177k cache-write) / 35k output tokens.

arbeithand:
*branch: fabrik/issue-1527 | commit: ab3e3bdf | main: ccf7971072bae615d2dde5c7bd4df853c5362ce8 | 2026-08-11 13:14 UTC*

Thread resolved. No code changes are needed here since both comments describe a finding already fixed in the immediately-preceding Review turn (commit `ab3e3bdf`), and no other findings were raised. The issue body needs no update either — this is implementation-detail feedback already reflected in the merged fix, not new spec information.

Working tree is clean, nothing to commit or push. Returning control to the engine.

---
**Threads addressed:**
- `scripts/e2e/run.sh:295` — resolved

arbeithand:
*branch: fabrik/issue-1527 | commit: ab3e3bdf | main: ccf7971072bae615d2dde5c7bd4df853c5362ce8 | 2026-08-11 13:26 UTC*

Everything is verified. Working tree is clean, branch is pushed, and the Pre-Completion Gate confirms mergeability. Final validation report:

## Validation Report

### Requirements: 4/4 implemented (R1/R4 empirically unverified — see below)
- [x] R2 (fail-loud detection): `run.sh` scans `fabrik.log` per leg for the one-shot `"activating rate-limit backoff"` line (never the spammy per-poll `"...is low..."` line), exits with dedicated code `3`, short-circuits remaining legs. Independently verified with synthetic log fixtures — correctly distinguishes the two lines. The Review stage's fix for the restart-window blind spot (capturing `log_offset` before the bed restart, not after) is present and correct.
- [x] R3 (cost driver investigation): Research's static analysis (confirmed by my own read of `queued_review_settle.go`/`ci_settle.go`) correctly identifies the ADR-1270/ADR-1208 settle-scan `FetchItemDetails` deep-fetch pattern — not merge-train's REST-based CI polling — as the GraphQL cost driver. Attribution logging added to `queued_review_settle.go` mirrors `ci_settle.go`'s existing pattern exactly; verified firing correctly in all 9 `TestSettleQueuedReviewFindings_*` tests.
- [x] R4 (cheaper mode): `E2E_PARALLEL_ON` (default 2) implemented, scoped correctly to only the default two-mode gate's `on` leg — `off` and forced single-mode runs still use `E2E_PARALLEL` unchanged (verified in diff).
- [~] R1 (zero-backoff full run): mitigation shipped, but **not empirically validated** — doing so requires a multi-hour live run against `~/dev/fabrik-test` that mutates real GitHub state, which Implement/Review correctly declined to run unilaterally and which is out of scope for a single Validate turn (see Pre-Completion Gate guidance on backgrounding/long-running verification). This is explicitly anticipated in the issue's own Risks/Dependencies section.

### Acceptance
- A1 (zero rate-limit-low events on a full run): **not verified** — same live-run dependency as R1 above.
- A2 (unambiguous invalid-run verdict): **verified** — synthetic fixture test confirms correct detection; byte-offset scoping (`tail -c "+$((offset+1))"`) confirmed correct under macOS BSD `tail`.
- A3 (measured cost recorded in README): **partially met** — the README section is honest about this: real per-leg numbers are marked "pending measurement" with the exact reproduction command, rather than fabricated. This matches the issue's own acknowledged risk ("if R1's cost work doesn't fully succeed, the measurement may need to come from per-leg or instrumented runs").

### Test Suite: PASSED
- `go build ./...`, `go vet ./...`: clean
- `gofmt -l engine/queued_review_settle.go`: clean
- `bash -n scripts/e2e/run.sh` + `shellcheck` (v0.11.0): clean
- `go test -race -timeout 5m ./...`: all packages pass. One failure (`TestInvokeClaude_FabrikEnvVarsInjected`) occurred only because this Validate session's own ambient `FABRIK_*` env vars leaked into the test process — confirmed by re-running with those vars stripped (`env -u FABRIK_TOKEN -u ...`), which passes cleanly, and by confirming it's unrelated to this PR's diff (same flake Review already identified and diagnosed).
- `docs/llms-full.txt` regen: no diff (no canonical doc touched by this PR — correctly skipped per Plan's assessment)

### Branch State
- Clean, committed, pushed, up to date with `origin/main` (0 commits behind)

### Verdict: READY TO MERGE

R1/A1/A3's empirical gaps are pre-disclosed, expected consequences of this being test-infrastructure work whose full validation requires a multi-hour live run against real GitHub state — not something Validate should trigger unilaterally. R2/A2 (the durable safety net, and the requirement explicitly *not* gated on R1) is fully implemented and independently verified correct. Landing this now gets the mitigation and fail-loud detection live for the next real gate run, which is exactly what will produce the missing numbers.

Validation passed. Rebase: skipped-up-to-date. PR mergeable: MERGEABLE, mergeStateStatus: CLEAN. R2/R3/R4 fully verified (synthetic fixture tests, code review, unit tests all green); R1/A1/A3 remain empirically unverified pending a live multi-hour two-mode bed run, which the issue's own Risks section anticipated as a real possibility — README documents this honestly with a reproduction command rather than fabricating numbers. Full test suite passes (one env-leak flake from this session's own ambient FABRIK_* vars, confirmed unrelated to the diff).

---
Used 40/100 turns, 3.0M input (57 raw + 2.9M cache-read + 107k cache-write) / 9k output tokens.

arbeithand:
*branch: fabrik/issue-1527 | commit: 2af83849 | main: ccf7971072bae615d2dde5c7bd4df853c5362ce8 | 2026-08-11 13:32 UTC*

Committed and pushed. No issue body update is needed — this is implementation-detail feedback about a bug in the fail-loud detection mechanism, not new spec information.

Both Pruefer's inline comment and the bot review finding identified a genuine, critical regression: the previous fix's `log_offset` was captured *before* the bed-restart step, but `engine/poll.go`'s `Run()` truncates `fabrik.log` (`O_TRUNC`) on every engine startup, and the restart step always launches a fresh engine process. Since the pre-restart offset almost always exceeded the post-truncation file's size, `tail -c +N` silently returned nothing — defeating R2's fail-loud detection for the *entire* leg, not just the restart window it was meant to fix.

I verified this empirically (confirmed `tail -c +N` on a file shorter than N returns nothing) and fixed it by removing the offset-tracking entirely: since the restart always truncates the log, scanning the whole current file after the suite run naturally covers the restart window through the suite's end, with no offset bookkeeping and no way for truncation to defeat it. Verified against synthetic fixtures for both the no-backoff and backoff-during-restart cases, and confirmed `go build`/`go vet`/`shellcheck`/`bash -n` all pass clean. Committed (`2af83849`) and pushed.

---
**Threads addressed:**
- `scripts/e2e/run.sh:324` — resolved

arbeithand:
*branch: fabrik/issue-1527 | commit: 2af83849 | main: ccf7971072bae615d2dde5c7bd4df853c5362ce8 | 2026-08-11 13:42 UTC*

## Validation Report

### Requirements: 4/4 implemented (R1/A1/A3's live-run measurement remains empirically pending, as pre-disclosed)

- [x] **R2 (fail-loud detection)** — `run.sh` scans the bed's `fabrik.log` in full after each leg for the one-shot `"...activating rate-limit backoff"` line (never the spammy per-poll `"...is low..."` line), prints an unambiguous `RUN INVALID` banner, and exits `3` (distinct from `go test`'s `1`), short-circuiting remaining legs. The prior Review-stage fix (commit `2af83849`) that removed byte-offset tracking — since the bed-restart's `O_TRUNC` on `fabrik.log` defeated the earlier offset approach — is present and correct; confirmed by reading the diff and the header comment's incident writeup.
- [x] **R3 (cost driver confirmed)** — `queued_review_settle.go` now carries the same cache-hit/deep-fetch attribution log line `ci_settle.go` already had; verified it compiles clean and is `gofmt`-clean.
- [x] **R4 (cheaper mode)** — `E2E_PARALLEL_ON` (default 2) is correctly scoped to only the default two-mode gate's `on` leg; `off` and forced single-mode runs still use `E2E_PARALLEL` unchanged (confirmed in diff).
- [~] **R1 (zero-backoff full run)** — mitigation shipped but not empirically validated; this requires a multi-hour live run against `~/dev/fabrik-test` mutating real GitHub state, correctly declined to run unilaterally in any stage so far — pre-disclosed in the issue's own Risks section.

### Acceptance
- A1: not verified (same live-run dependency as R1).
- A2: verified — detection logic and exit-code mechanism read correctly in the diff; synthetic-fixture verification was done in prior stages.
- A3: partially met — README documents the mechanism and knobs with an honest "pending measurement" table and exact reproduction command, matching the issue's own acknowledged risk.

### Test Suite: PASSED
- `go build ./...`, `go vet ./...`: clean
- `gofmt -l` on this PR's changed Go file (`engine/queued_review_settle.go`): clean (other files flagged by a repo-wide `gofmt -l` sweep are pre-existing and untouched by this diff)
- `bash -n scripts/e2e/run.sh`: clean
- `go test -race -timeout 5m ./...`: **all packages pass** once run with the ambient `FABRIK_PR`/`FABRIK_TOKEN`/`FABRIK_ISSUE` env vars stripped (`env -u ... go test ...`). Without stripping, `TestInvokeClaude_FabrikEnvVarsInjected` fails — confirmed this is caused purely by my own Validate session's ambient `FABRIK_PR=1547`/etc. leaking into the test process, not by the diff, matching the exact flake both prior Review and Validate turns already diagnosed and fixed for.
- `docs/llms-full.txt` regen: correctly skipped — no canonical doc (`USER_GUIDE.md`/`state-machine.md`/`stage-lifecycle.md`/`positioning.md`) touched by this PR.

### Branch State
Clean, committed, pushed, 0 commits behind `origin/main`.

Re-validated PR #1547 after the restart-window-blindspot fix. Rebase: skipped-up-to-date (0 commits behind main). PR mergeable: MERGEABLE, mergeStateStatus: CLEAN, all 5 required CI checks pass. Full test suite green (one pre-existing env-leak flake confirmed unrelated to the diff, same as prior stages found). R2/R3/R4 fully verified; R1/A1/A3 remain empirically pending a live multi-hour bed run, as pre-disclosed in the issue. Verdict: READY TO MERGE.

---
Used 31/100 turns, 3.2M input (59 raw + 3.1M cache-read + 113k cache-write) / 7k output tokens.

arbeithand:
*branch: fabrik/issue-1527 | commit: 2af83849 | main: ccf7971072bae615d2dde5c7bd4df853c5362ce8 | 2026-08-11 13:43 UTC*

Pruefer's bot review finding is a confirmation pass — "No defects found," verifying the `2af83849` fix (full-log-scan approach relying on `fabrik.log`'s `O_TRUNC`-on-startup) is correct, along with the `queued_review_settle.go` attribution log line and the rest of the shipped implementation. There is nothing actionable here per the No-Op Contract: no code change is warranted, and the working tree is already clean, committed, and pushed. No issue body update is needed either — this is confirmation of already-implemented, already-validated work, not new spec information.

Returning control to the engine without signaling completion (per the skill's default — this comment doesn't explicitly direct completion).

---
**Threads addressed:**

verveguy:
## Review: R2's backoff detection never fires — it fails open

**Blocking.** R2 is the requirement this PR exists to guarantee, and as implemented it
silently cannot fire. A throttled run still reports as an ordinary pass or fail, which is
precisely the condition #1527 was filed to eliminate.

### The defect

`switch_and_run` captures the log offset **before** the bed restart:

```bash
local log_offset
log_offset=$(wc -c < "$ENGINE_LOG" 2>/dev/null || echo 0)

echo "== switching test bed to FABRIK_MERGE_TRAIN=${mode} =="
E2E_TRAIN_SWITCH=1 ... -run '^TestSwitchTrainMode$' ./tests/e2e/...
```

and later scans from it:

```bash
tail -c "+$((log_offset + 1))" "$ENGINE_LOG" | grep -q 'activating rate-limit backoff'
```

But the engine **truncates `fabrik.log` on every startup** — `engine/poll.go:231-234`:

```go
// Open the persistent poll log file. Truncated on each startup so the file
// always reflects the current run only. Non-fatal if the open fails.
logPath := filepath.Join(e.fabrikDir, ".fabrik", "fabrik.log")
if lf, err := os.OpenFile(logPath, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0600); err != nil {
```

The restart that `TestSwitchTrainMode` performs resets the file to zero bytes. The offset
captured beforehand therefore points far beyond the end of the new log, and the scan
reads nothing until the fresh run happens to exceed the *previous* run's size.

The test bed's `fabrik.log` is currently **643,858 bytes**. A leg would have to emit more
than 600KB of fresh log before the detector examines a single byte.

### Reproduction

```
offset captured before restart: 800 bytes
new log size after restart + backoff line: 101 bytes
scan from captured offset:  *** NOT DETECTED — fails open ***
control, scan from offset 0:    detected
```

i.e. write a file, capture `wc -c`, truncate it (`: > file`), append the literal
activation line, then run the PR's own `tail -c "+$((off + 1))" | grep -q` expression.
It reports no match. The same scan from offset 0 matches.

### Why the existing verification missed it

The PR body records: *"Detection logic manually verified against three synthetic
`fabrik.log` fixtures ... an activation line before the leg's own log offset is correctly
excluded by the byte-offset scoping."*

Those fixtures exercise the offset **arithmetic**, which is correct. None of them models
the **environment** — a log that gets truncated between capture and scan. The mechanism
was validated; the setting it runs in was not.

### The fix

Capture `log_offset` **after** the restart step, not before it — or simply scan the whole
file, since truncation-on-startup already scopes it to the current run.

Note this reverses a deliberate change. The current code carries a twelve-line comment
justifying the early capture and explicitly rejecting the alternative:

> *Capturing the offset only after the restart (as an earlier version of this script did)
> left exactly that window — plus the idle gap since the previous leg's own scan already
> ran — unscanned by either leg.*

That reasoning does not hold once truncation is accounted for. The window it worries
about — a backoff activation logged by the freshly-restarted engine's very first poll,
before the suite starts — is **already covered** by capturing after the restart, because
the post-restart log contains *only* the current run from byte 0. Capturing after the
restart is both correct and strictly more inclusive than the current code. Please replace
that comment rather than leaving it to justify the old position.

### Verification required

The standing bar on this repo is non-vacuity: neutralize the fix and confirm the check
goes red, then record the neutralization and the observed failure in the PR body.
"Verified manually" is not sufficient.

Concretely, for this change:

1. Build a fixture that **truncates** the log between offset capture and scan (the
   reproduction above is sufficient).
2. Show the detector **matches** with the fix applied.
3. Revert the fix in isolation (restore the pre-restart capture) and show the same
   fixture **fails to match** — this is the neutralization step, and it is the whole
   point, since the current code passes every fixture that omits truncation.
4. Record both observations in the PR body.

Consider whether this belongs in a small shell-level test rather than a manual fixture,
given R2 is a release-gating guarantee and currently has no automated coverage at all.

---

### Non-blocking note

`E2E_PARALLEL_ON`'s default of 2 is explicitly untuned, and the cost table in
`tests/e2e/README.md` ships as `_pending measurement_`. That is disclosed honestly in the
PR body and is acceptable **only because R2 makes an under-tuned run fail loudly rather
than lie** — which makes it all the more important that R2 actually works. No change
requested here; flagging so the placeholder table isn't later mistaken for a measured
result.

arbeithand:
*branch: fabrik/issue-1527 | commit: dd1f9740 | main: ccf7971072bae615d2dde5c7bd4df853c5362ce8 | 2026-08-11 14:20 UTC*

## Validation Report

This is a re-validation of PR #1547 following @verveguy's blocking review comment. That comment was written against a stale rendering of the code — the "log_offset captured before restart" defect it describes was already fixed in commit `2af83849` (a prior comment-review turn responding to Pruefer's independent finding of the same bug), before the review comment was posted. The comment-review turn (`dd1f9740`, current HEAD) correctly identified this and, rather than dismissing the review, treated its "verification required" ask as still valid and satisfied it: extracted `detect_rate_limit_backoff()` into a standalone, sourceable function and added `scripts/e2e/backoff_detection_test.sh` — 4 direct fixture cases plus a genuine neutralization case (Case 5 reintroduces the historical byte-offset approach against the same fixture Case 4 catches, and confirms it fails, proving the fixture exercises the real regression rather than trivially always matching).

### Requirements: 4/4 implemented (R1/A1/A3's live-run measurement remains empirically pending, as pre-disclosed)

- [x] **R2 (fail-loud detection)** — Read `detect_rate_limit_backoff()` directly: scans the whole current `fabrik.log` (not a byte offset), matching only the one-shot `"activating rate-limit backoff"` line, never the spammy per-poll `"...is low..."` line. Correct because `engine/poll.go`'s `Run()` truncates the log on every startup, so a full-file scan after the leg's restart naturally covers the restart window through the suite's end with no offset bookkeeping. Exits `$BUDGET_EXHAUSTED_EXIT` (3), short-circuits remaining legs. **Now has non-vacuous automated coverage**: ran `bash scripts/e2e/backoff_detection_test.sh` myself — all 5 cases pass, including the neutralization case, confirming the fixture would catch a regression back to byte-offset scanning.
- [x] **R3 (cost driver confirmed)** — attribution logging in `queued_review_settle.go` mirrors `ci_settle.go`'s existing pattern; verified via `go test -race ./engine/...` (all `TestSettleQueuedReviewFindings_*` pass).
- [x] **R4 (cheaper mode)** — `E2E_PARALLEL_ON` (default 2) confirmed scoped to only the default two-mode gate's `on` leg (`switch_and_run on "$PARALLEL_ON"`); `off` leg and forced single-mode (`E2E_TRAIN_MODE` set explicitly) still use `E2E_PARALLEL` unchanged — verified by reading `run.sh`'s dispatch block directly.
- [~] **R1 (zero-backoff full run)** — mitigation shipped, not empirically validated; requires a multi-hour live run against `~/dev/fabrik-test` mutating real GitHub state, correctly declined to run unilaterally in every stage so far, pre-disclosed in the issue's own Risks section.

### Acceptance
- A1: not verified (same live-run dependency as R1).
- A2: **verified** — `detect_rate_limit_backoff` reads correctly and now has automated, neutralization-tested coverage (not just manual fixture inspection).
- A3: partially met — `tests/e2e/README.md`'s new "#1527" subsection documents the mechanism, root cause, and mitigation honestly, with real per-leg numbers marked pending measurement and an exact reproduction command.

### Test Suite: PASSED
- `go build ./...`, `go vet ./...`: clean
- `gofmt -l engine/queued_review_settle.go`: clean
- `bash -n scripts/e2e/run.sh`, `bash -n scripts/e2e/backoff_detection_test.sh`: clean
- `shellcheck` on both shell files: clean (no findings)
- `bash scripts/e2e/backoff_detection_test.sh`: all 5 assertions pass, including the neutralization case
- `go test -race -timeout 5m ./...` (ambient `FABRIK_*`/`GH_TOKEN` env stripped): **all 24 packages pass**, no failures, no race reports
- `docs/llms-full.txt` regen: correctly skipped — no canonical doc (`USER_GUIDE.md`/`state-machine.md`/`stage-lifecycle.md`/`positioning.md`) touched by this PR; `tests/e2e/README.md` is not in that list

### Branch State
Clean, committed, pushed, 0 commits behind `origin/main`.

Rebase: **skipped-up-to-date** (`git rev-list --count HEAD..origin/main` = 0). PR mergeable: `MERGEABLE`, mergeStateStatus: `BLOCKED` (expected — required checks still settling post-push: 4/5 pass, `Verify llms-full.txt is up to date` was still pending at check time, not failed). No conflicts.

### Verdict: READY TO MERGE

R2/A2 — the requirement this PR exists to guarantee — is now independently, non-vacuously verified: I ran the new test script myself and confirmed the neutralization case actually fails the historical broken approach. @verveguy's review was correct to demand real verification; it was answered by the immediately-following comment-review turn, not dismissed. R1/A1/A3's numeric gaps remain exactly as pre-disclosed in the issue's own Risks section — a live multi-hour bed run, appropriately not triggered unilaterally by any pipeline stage.

Re-validated PR #1547 after @verveguy's review, which was based on a stale code snapshot — the defect it flagged (log_offset captured before restart) was already fixed in commit 2af83849, and the follow-up commit dd1f9740 added the non-vacuous test coverage the review demanded (verified by running it myself: all 5 cases pass, including the byte-offset neutralization case). Rebase: skipped-up-to-date. PR mergeable: MERGEABLE, mergeStateStatus: BLOCKED (checks still settling, none failed). Full test suite green (24 packages, race-clean). R2/R3/R4 fully verified; R1/A1/A3 remain empirically pending a live multi-hour bed run, as pre-disclosed in the issue.

---
Used 21/100 turns, 1.6M input (30 raw + 1.5M cache-read + 109k cache-write) / 6k output tokens.

verveguy:
## Re-review: the fix is correct and the test is genuinely non-vacuous — one gap left

**The R2 defect is fixed properly.** `detect_rate_limit_backoff` now scans the whole file:

```bash
detect_rate_limit_backoff() {
  local logfile="$1"
  [ -f "$logfile" ] && grep -q 'activating rate-limit backoff' "$logfile" 2>/dev/null
}
```

That's the right call — `O_TRUNC`-on-startup already scopes the file to the current run,
so no offset bookkeeping is needed and the restart-window case is covered by construction.
The misleading comment defending the old position was replaced rather than left standing,
which was the other thing I asked for.

**I verified the test independently rather than taking it on trust.** Ran
`scripts/e2e/backoff_detection_test.sh` on the branch — all 5 cases pass. Then neutralized
the fix myself by restoring a pre-restart byte-offset scan into
`detect_rate_limit_backoff` and re-ran:

```
FAIL: activation line present, no truncation (expected match, got no-match)
FAIL: activation line logged only after a mid-leg truncation (the field regression)
=== backoff_detection_test.sh: FAILED ===
```

It goes red against the broken implementation. The coverage is real, and Case 5's inline
neutralization is a good pattern — better than a one-off manual revert, because it makes
the broken approach permanently self-identifying.

---

## Remaining gap (blocking): nothing runs this script

`scripts/e2e/backoff_detection_test.sh` is not invoked anywhere — not in
`.github/workflows/ci.yml`, not from `go test`, not from any script or Makefile target.
CI runs only:

```yaml
go vet -tags e2e ./...
go test -race -timeout 5m -skip TestExecute_ConfigYAMLApplied ./...
go test -tags e2e -run '^$' ./tests/e2e/...   # compile-only
```

The script's own header states its purpose:

> *"This script exists so that regression can't recur silently: ... so a future edit that
> reintroduces byte-offset tracking (or otherwise narrows the scan) fails this script
> instead of only being caught by a real, multi-hour bed run."*

As wired, a future edit fails nothing. The script only protects the guarantee if someone
remembers to run it by hand — which is the same class of problem as the original defect:
a safeguard that looks present and isn't active. Given R2 is release-gating, an unrun test
is precisely the "quietly reduced to an optimization" outcome #1527 exists to prevent.

**Requested:** add a step to `.github/workflows/ci.yml` that runs it. It needs no test bed,
no GitHub token, and no network — it sources `run.sh` for the function and drives fixtures
in `mktemp`, so it costs roughly nothing in CI:

```yaml
- name: Backoff-detection regression test
  run: bash scripts/e2e/backoff_detection_test.sh
```

Please confirm it passes in CI on this PR, so the guard is demonstrated active rather than
merely present.

Two small things worth checking while you're in there, neither blocking:

1. The script sources `run.sh`, which does top-level `TEST_BED`/`BED_TOKEN` resolution.
   That's read-only and degrades to a warning on a missing `$TEST_BED/.env`, so it should
   be fine on a bare CI runner — but confirm it doesn't emit a failing exit under
   `set -euo pipefail` there, since CI has no `~/dev/fabrik-test`.
2. `shellcheck` the new script if the toolchain is available, matching what the PR body
   reports was done for `run.sh`.

arbeithand:
*branch: fabrik/issue-1527 | commit: ce2ec951 | main: ccf7971072bae615d2dde5c7bd4df853c5362ce8 | 2026-08-11 14:27 UTC*

Pruefer's bot review finding is a confirmation pass — "No new defects found," verifying the whole-file-scan fix in `run.sh`, the `backoff_detection_test.sh` regression coverage, the `queued_review_settle.go` logging parity change, and the README updates are all correct. Per the No-Op Contract, there's nothing actionable here: no valid finding to fix, no code change warranted. The working tree is already clean, committed, and pushed (`ce2ec951`).

No issue body update is needed either — this is confirmation of already-implemented, already-validated work, not new spec information.

Returning control to the engine without signaling completion — this is a bot review confirmation, not an explicit user statement that the issue is resolved.

---
**Threads addressed:**

arbeithand:
*branch: fabrik/issue-1527 | commit: ce2ec951 | main: ccf7971072bae615d2dde5c7bd4df853c5362ce8 | 2026-08-11 14:29 UTC*

Pruefer's bot review finding is another confirmation pass — this time explicitly covering the CI wiring added in `ce2ec951` (`.github/workflows/ci.yml`), along with re-confirming the whole-file-scan fix, the fixture set, and the README/`queued_review_settle.go` changes. It independently ran `bash -n`/`shellcheck`/`backoff_detection_test.sh` itself and got the same clean result I did. No new defects found. Per the No-Op Contract, there's nothing actionable here — no code change is warranted, and the branch is already clean, committed, and pushed.

No issue body update is needed — this is confirmation of already-implemented, already-validated work, not new spec information.

Returning control to the engine without signaling completion — this is a bot review confirmation, not an explicit user statement that the issue is resolved.

---
**Threads addressed:**

Issue #1563, opened 2026-08-12
pruefer: Hookdeck protocol drift is undetectable — every delivery can fail signature verification silently


arbeithand:
## Problem

Pruefer's Hookdeck adapter rests on an assumption that a third party can
invalidate without notice, and if it is invalidated the failure is silent — it
looks identical to "no PRs needed review."

`processAttempt` verifies GitHub's HMAC over the body Hookdeck forwarded:

```go
if !events.VerifySignature([]byte(attempt.Request.DataString), sig, s.cfg.WebhookSecret) {
```

This requires `attempt.request.data_string` to be **byte-identical** to the body
GitHub signed. Nothing enforces that. It is a property of Hookdeck's
CLI-session wire format, derived by reading `hookdeck-cli`'s source — not a
documented contract, not versioned, and not covered by CI, which mocks the
transport entirely (ADR-1254 already names this last point).

If Hookdeck ever re-serializes the body in transit — re-encodes, normalizes
whitespace, reorders JSON keys — **every** signature check fails.

### Why the failure is invisible

The drop path is deliberately quiet, and each individual choice is defensible:

1. A failed verification drops the event.
2. The attempt is still acked `Status: 200`, so nothing accumulates Hookdeck-side.
3. The signature-failure warning is rate-limited to one line per 30s
   (`sigFailureLogInterval`) and goes to the log, which nothing watches.
4. The reconciliation fallback ticker keeps polling every 2 minutes, so reviews
   still happen — just at poll latency.

Composed, they mean a total protocol break degrades Pruefer to poll-only and
reports nothing. The observable difference between "Hookdeck changed its wire
format and every event is being discarded" and "nothing has happened lately" is
a rate-limited log line.

Degrading to poll rather than dying is the right behavior (ADR-1254 D3, "poll
remains the source-of-truth fallback"). Degrading *undetectably* is not.

### What ADR-1254 does and does not cover

Already recorded, and correctly:

- D2: the protocol is a reimplementation of an undocumented internal protocol,
  an accepted risk, contained behind the `EventSource` boundary.
- Consequences: it is "an internal implementation detail of `hookdeck-cli`, not
  a versioned public API. A future upstream change could break it."
- Consequences: "No live Hookdeck account is exercised in CI."

Not recorded:

- That **HMAC verification specifically** depends on byte-exact body
  preservation — the single most load-bearing detail of the reimplementation,
  and the one whose breakage is silent rather than loud.
[...]

PR #1564, opened 2026-08-12
docs(adr-1254): record the byte-exactness assumption and its silent failure mode

arbeithand:
Documentation-only. Amends ADR-1254 with the two specifics that were load-bearing but implicit, and corrects one overstated comment.

**What ADR-1254 already said, correctly:** the CLI-session protocol is a reimplementation of an undocumented internal protocol; it is "an internal implementation detail of `hookdeck-cli`, not a versioned public API"; no live Hookdeck account is exercised in CI.

**What it did not say:**

1. **Signature verification depends on byte-exact body preservation.** `processAttempt` computes GitHub's HMAC over `attempt.Request.DataString`, so it is correct only if Hookdeck forwards the body byte-for-byte. Any re-serialization in transit makes *every* check fail — and unlike a struct-shape mismatch, which fails loudly at unmarshal, this fails as a uniform silent rejection.

2. **The failure is unobservable.** Drop + ack-200 + a 30s rate-limited warning + the poll fallback compose so that a total protocol break is distinguishable from a quiet hour only by reading a log line. Degrading to poll is Decision 3 working as designed; degrading *undetectably* is not.

Also corrects `client.go`'s "the verified upstream protocol" — traced from upstream source is accurate; verified is not.

No functional change; `go build ./...` clean. Making the degradation detectable is tracked separately in #1563, along with the ADR for whatever detection mechanism is chosen.

### DIFF
--- a/adrs/1254-event-driven-hookdeck-ingestion.md
+++ b/adrs/1254-event-driven-hookdeck-ingestion.md
@@ -69,6 +69,54 @@ Before this issue, `poll()` created a fresh per-cycle semaphore local to itself.
 - `daemon.go` gained real complexity: a shared semaphore, an extracted `reviewOne`, a `runPollOnly`/`runEventDriven` split, and a health-transition callback (`HealthHandler`) — all net-new surface area even though `runPollOnly`'s own behavior is unchanged.
 - No live Hookdeck account is exercised in CI; all coverage is `httptest`/`gorilla/websocket`-mocked at the transport boundary (signature validation, delivery-ID dedupe, normalized-event mapping, reconnect-after-drop, session-creation-failure retry). A manual smoke test against a real Hookdeck source remains a pre-production step outside this repo's test suite.
 
+## Amendment (2026-08-12, from #1563): what the reimplementation actually rests on
+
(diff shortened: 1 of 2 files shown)


arbeithand:
Closing as superseded by #1565.

#1565 (from issue #1563, whose R1/R5 asked for exactly this) amends `adrs/1254-event-driven-hookdeck-ingestion.md` with the same two points — the byte-exactness dependency of signature verification, and the four-decision composition that made a total protocol break undetectable — and does so more thoroughly, enumerating why each of the four decisions is individually correct and noting that Research found the raw signal already present at nine distinct drop points.

It also makes the same R5 correction to `client.go`:

```diff
-// per the verified upstream protocol (see protocol.go).
+// per the protocol traced from hookdeck-cli's source (see protocol.go) —
```

And unlike this PR it ships the actual detection (`OnDrop`/`OnSignatureDrift`, `SignatureDriftThreshold`, TUI surfacing) rather than only documenting the gap.

This PR was written by hand while #1563 was already in flight to do the same work — a duplicate of my own making. Merging both would conflict in `adrs/1254` for no added value. Nothing here is lost; branch `docs/adr-1254-byte-exactness` is preserved.

PR #1638, opened 2026-08-22
notes: findings from running the spec-authoring plugin on its first real project

verveguy:
## Summary

86ED is the project `CONVENTIONS.md` already cites as the motivating case for stage 3 being a living document — *"eleven specifications written over days drifted, because each new feature changed the meaning of earlier ones and nothing owned the shared entities."* This records what happened when that stage was finally run against it.

Four days, nine baseline versions, four new specifications, two constitution amendments. Fields with no feature that could own them went **25 → 1**, and every drop came from a specification being written rather than from further work on the architecture document.

Six substantial findings and three smaller ones. Each has what happened, what the method says today, and a proposed change. **Nothing here is a defect in the method as written** — they are things it does not currently say.

## The findings

1. **Stage 3 finds missing specifications and there is no route back to stage 2.** 24 of the 25 unowned fields were missing specs. The method says saying so is "a finding addressed to the product owner" and stops. Proposes describing the loop as a first-class beat, alongside the revision beat it resembles.
2. **"Spec amendments owed" earned a template slot.** Resolved-but-unapplied is a distinct state from unresolved, and a more dangerous one — the baseline reads as settled while the specs still say the old thing. This project carried twenty at peak.
3. **Topic 4 should ask *how* it will be built.** Answering "one human plus an agent pipeline" changed the completeness bar for everything: an engineer asks, an agent invents. Agent-built and human-built projects have different definitions of a finished spec.
4. **A clean checklist does not mean a consistent corpus.** A spec came back 16/16 with no markers while contradicting another spec. Not a `write-spec` defect — it validates a document against itself — but the product owner reasonably read 16/16 as "done".
5. **The gate decays silently.** Four version bumps without re-evaluation. The conventions say when to *preserve* an evaluation, never when to *run* one.
6. **A fourth field disposition.** Deliberately out of scope — restricted, deferred or excluded — is neither owned nor a gap, and labelling it a gap invites someone to fill it.

Plus: the method produces AI-assisted documents and says nothing about labelling them, while projects routinely write exactly that rule for themselves; exclusions should record *why* so they can be audited for staleness; and a question spanning two domains should be split before it is asked.

## What this does not do

No changes to `CONVENTIONS.md` or any skill. Several of these bind three other projects, so they should be read as proposals before they become rules.

The note closes with what worked and should not be changed, which is most of it — the three dispositions, counts over claims, preserved evaluations, stage 3 staying Draft, and the register's Class column.

### DIFF
--- a/notes/entwurf-findings-86ed-2026-08-22.md
+++ b/notes/entwurf-findings-86ed-2026-08-22.md
@@ -28,7 +28,7 @@ v2.0.0    1        after three more specifications
 Every drop came from a specification being written. None came from further work on the
 architecture document. That is the headline, and several findings below follow from it.
 
-**Ten findings and three smaller ones.** Nothing here is a defect in the method as written —
+**Eleven findings and three smaller ones.** Nothing here is a defect in the method as written —
 these are things it does not currently say. Findings 3 through 5 all follow from one fact the
 method never states: the reader of these documents is an agent.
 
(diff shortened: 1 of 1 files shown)


verveguy:
Closing without merging. the spec-authoring plugin has been broken out of this repository into its own — it is valuable on its own and is not being open-sourced for now — so these findings moved with it, along with the full `the plugin directory/` history.

This note also names a real client project and its architecture in some detail, which is a second reason not to land it in a public repository.

Nothing here is lost: the file is committed in the spec-authoring plugin repository, with a dated note recording that two sentences about the two plugins shipping together were corrected. Every finding stands unchanged.

PR #1675, opened 2026-08-29
fix(pruefer/tui): fill the terminal, hide skips, align columns, move Watched Repos last

### LINKED ISSUE #1674: pruefer TUI: fill the terminal height, hide skips from Completed Reviews, align its columns, move Watched Repos last
## Problem

Four issues with the Pruefer TUI, all visible in one screenshot on a ~62-row
terminal. They are filed together because they are the same layout pass.

### 1. The TUI does not fill the terminal

It renders a fixed-height layout and leaves the bottom half of the window empty.

The cause is structural, not a sizing bug: **terminal height is captured and
never used.** `pruefer/tui/model.go:116` stores it —

```go
case tea.WindowSizeMsg:
    m.height = ev.Height
```

— and `grep -rn "m\.height" pruefer/tui/` returns that single line. Nothing reads
it. Every component's signature is `View(width int)`: width is threaded through,
height is not. Panel sizes are therefore fixed constants, most visibly
`maxDisplayedHistoryRows = 10` (`history.go:170`).

So on a tall terminal the panels stay small and the space goes unused; on a short
one the layout presumably overflows.

### 2. Skipped PRs dominate Completed Reviews

Every outcome lands in the same list, and skips are by far the most common —
`history.go:103-105` renders them as `skipped: <reason>`. In the observed
screenshot **9 of 10 visible rows** were `skipped: already reviewed at this head
SHA`, pushing actual reviews out of view. The steady state of a healthy daemon is
mostly skips, so the panel's most valuable content is the content it evicts.

### 3. Completed Reviews columns are ragged

`history.go:91-93` truncates the repo/PR label to a fixed `maxWidth = 20` and
then concatenates timestamp, duration, and reason. Nothing aligns to a column
grid, so entries of differing name length stagger:

```
verveguy/liminis-context-graph#286 22:41:41 00:02  sk…
verveguy/86ed#141      22:41:39 00:01  skipped: alr…
```

The truncation also lands mid-token, and the trailing reason is cut with an
ellipsis at an arbitrary point.

### 4. Watched Repos is the largest panel and sits at the top

Render order (`model.go:229-240`) is header → repos → active → detail → history →
footer. Watched Repos is static, grows linearly with the repo count (15 rows
today), and pushes the panels that actually change — In-Flight and Completed —
down the screen.

## Requirements

**R1 — Make the layout height-aware.** Thread the terminal height through to
components so the layout fills the window. Panels that list rows (Watched Repos,
Completed Reviews) should size to the space available rather than to a constant.
Define what happens when content exceeds available height — scroll, truncate with
a count, or a stated priority order for which panel yields first — and when it
falls short.

**R2 — Filter skips out of Completed Reviews by default.** Show reviews that
actually produced a review. Skips should remain *discoverable* — a count, a
toggle key, or a dim summary line — but must not occupy the list by default.
`already reviewed at this head SHA` in particular is the daemon working
correctly and carries no information at steady state. Errors must **not** be
filtered out with skips; a failed review is the most important row on the panel.

**R3 — Align Completed Reviews into columns.** Compute column widths from the
rendered set and pad to a grid, so repo/PR, time, duration, and outcome line up.
Truncate at a column boundary rather than mid-token, and prefer eliding the
middle of a long repo name over cutting its PR number, which is the identifying
part.

**R4 — Reorder so Watched Repos is last.** Put the changing panels (In-Flight,
Completed) above the static one. Watched Repos moves to the bottom, above the
footer.

## Scope

**In scope**
- `pruefer/tui/model.go` — height propagation, layout budget, panel order
- `pruefer/tui/history.go` — skip filtering, column layout, height-aware row count
- `pruefer/tui/repos.go` — height-aware sizing at its new position
- `pruefer/tui/active.go`, `footer.go`, `detail.go` — signature changes from R1
- Tests

**Out of scope**
- Which events the daemon emits (`ptui.*Event`) — this is presentation only
- The `--notui` headless path and the log file format
- Colour scheme, borders, and typography
- New interactions beyond whatever R2's skip-discoverability needs

## Acceptance criteria

1. On terminals of at least three distinct heights, the rendered output occupies
   the full height without overflowing, asserted by measuring rendered line count
   against the simulated `WindowSizeMsg` — not by eyeballing a screenshot.
2. A `WindowSizeMsg` resize re-lays-out to the new height.
3. With a history containing both skips and completed reviews, the default view
   shows only the latter; a history of skips alone renders an empty-with-count
   state rather than a misleading blank.
4. A review that errored is shown, never filtered as a skip (R2).
5. Column positions are identical across rows for a set with widely varying repo
   name lengths (R3), asserted on the rendered strings.
6. Panel order is header → in-flight → completed → watched repos → footer (R4).
7. Content exceeding the available height degrades per R1's stated rule, with a
   test at a deliberately short height.
8. `go test -race ./pruefer/...` and `go vet ./...` pass; every new test
   demonstrated non-vacuous.

arbeithand:
Closes #1674. All four requirements.

## Before / after

The pane that made this obvious — same data, 100x62 terminal:

```
  Completed Reviews (7) · 22 skipped
✗ verveguy/liminis-context-graph#286 23:14:59 00:03  fetching diff: context deadline exceeded
✓ verveguy/claude-topics#205         23:14:59 01:35  9 turns $0.0123
✓ verveguy/86ed#204                  23:14:59 01:34  8 turns $0.0123
✓ shadoworg/fantasy#203              23:14:59 01:33  7 turns $0.0123
```

Previously nine of ten rows here were `skipped: already reviewed at this head SHA`, and the columns staggered with repo-name length.

## R1 — height

Terminal height was captured at `model.go:116` and never read again; every component is `View(width int)`, so panel heights were constants (`maxDisplayedHistoryRows = 10`). `Model.View` now budgets the remaining height between the two flexible panes and each stretches to its grant, so the layout reaches the bottom of the window.

The allocation rule is **stated, not incidental**: the changing pane is served first. Completed Reviews takes what its content needs, Watched Repos takes the remainder and hands surplus back, both keep a `minPaneRows` floor, and overflow windows with `… N more`.

## R2 — skips

Filtered from the list, disclosed as a count. **An errored entry is never filtered**, even if `Skipped` is also set — that is the trap in this requirement and it has its own test.

## R3 — columns

Column width computed from the rendered set, padded on **rune** count (an elided key contains a multi-byte `…` that `%-*s` under-pads). Over-long keys elide from the middle, preserving the PR number.

## R4 — order

header → in-flight → completed → watched repos → footer.

## Verification

`TestHistoryPane_AllSkipReasonsRenderInHistory` asserted the behavior R2 changes, so it is **replaced** rather than deleted, by a test asserting the inverse plus the errored-skip guard.

Non-vacuity, by neutralising each fix:

```
height plumbing off  → height=30: rendered 62 rows, overflows the terminal
                       height=90: rendered only 62 rows, leaving 28 unused
skip predicate off   → View() still lists skip reason "draft"
```

`go test -race ./pruefer/...`, `go vet ./...`, `gofmt` all clean.

Verified visually by rendering the model with synthetic events at several window sizes rather than against the live daemon, which stayed up throughout.


arbeithand:
Both findings addressed — both were real defects I introduced, one per change.

**`history.go:213` — selection index.** Correct and well-diagnosed, including the repro. `h.idx` is a position in the *rendered* list, but navigation bounds and `Selected()` still resolved against `h.entries`, which includes the filtered skips. With `[real1, skip1, real2]`, one `down` highlighted `real1` while `Selected()` returned `skip1` — highlighted row and detail panel disagreeing as soon as one skip existed. Both now resolve against `visibleEntries()`, so a filtered skip is stepped over rather than selectable, and navigation stops at the last visible row.

**`history.go:278` — `Height()` vs `View()`.** Also correct. `View` pads to the granted budget, so a pane with a budget always renders `budget+3`; `Height()` returned `min(visible, budget)+3` and under-reported whenever content was shorter (budget 20, 3 entries → 6 vs 23 actually rendered). Both flexible panes now report the padded height. You correctly noted `Height()` has no non-test caller today — I fixed it anyway because it is part of the `Component` contract and would silently mis-budget any future caller, and applied the same fix to `RepoPaneComponent`, which pads identically and had the same latent bug.

Regression tests for both, each demonstrated non-vacuous by reverting its fix:

```
selection off → after one 'down', selected PR 2 (the skip), want PR 1
height off    → history Height() = 4, but View() renders 23 rows
```

`go test -race ./pruefer/...`, `go vet ./...`, `gofmt` clean.

arbeithand:
Superseded — this branch's commits (fill-the-terminal layout, skip filtering/discoverability, column alignment, panel reorder, plus the selection-index and Height()/View() review fixes) have been merged into the Fabrik-managed branch for issue #1674 (`fabrik/issue-1674`), along with additional test coverage, an ADR, and a README update. Fabrik's engine will open a fresh PR from that branch tracking issue #1674 going forward; closing this one to avoid two open PRs for the same issue. No code is lost — this branch's tip is an ancestor of the new one.

PR #1689, opened 2026-09-03
fix(tui): stop labelling every comment pass as a pending retry

arbeithand:
The history pane renders every comment-processing invocation as `↻ (retry)`, and the detail panel as `incomplete`. Neither is true, and it affects **all** of them rather than some.

Spotted in the TUI, where the giveaway is a row like:

```
#221  Validate 💬 ↻  00:30  ...  6/200 turns  $0.72  (retry)  Replace in-proce
```

Six turns of two hundred. Nothing was exhausted; nothing is pending.

## Cause

Both sites key purely on `HistoryEntry.Completed`, which is set from whether the invocation emitted `FABRIK_STAGE_COMPLETE`:

```go
} else if !he.Completed {
    status = dimStyle.Render("↻")
    if he.TurnLimited { result = "  (turn limit)" } else { result = "  (retry)" }
}
```

But comment processing is **instructed never to emit that marker**. Every `fabrik-*-comment` skill says so outright:

> **Do NOT output `FABRIK_STAGE_COMPLETE`.** Comment processing in Review returns control to the engine without advancing the pipeline.

> - **Do not signal stage completion** — never output `FABRIK_STAGE_COMPLETE`

So `Completed` is false for every comment invocation by design, and the status branch reads that as "incomplete, will be retried". A comment pass that did exactly what it was asked gets the same glyph as a stage run that ran out of turns and is genuinely queued for re-dispatch.

The information needed to tell them apart was already there: `history.go` reads `he.IsComment` thirty-five lines below the status branch, purely to append the 💬 glyph. This PR uses it in the branch where it changes the meaning.

## Change

A comment pass that succeeded, was not blocked on input, and did not hit the turn cap is complete — `✓` in the history pane, `success` in the detail panel.

Deliberately narrow:

- **turn-limited comment passes still report `(turn limit)`.** The comment skills have their own "If You Hit the Turn Limit" section, so that state is real for them and must not be swallowed by the new branch.
- **the stage-run path is untouched.** A stage ending without the marker really will be re-dispatched, and still shows `↻ (retry)`.
- `DetailItem` gains `IsComment` (populated in `prepareDetailItem`) because it did not carry the field; nothing else consumes it.

## Tests

Three cases added to `TestViewHistory_TurnLimitClassification`: the comment-complete case, the turn-capped comment case, and the unchanged stage-run case — so the fix cannot be satisfied by an implementation that simply stops reporting turn limits, or one that changes stage behaviour.

**Demonstrated non-vacuous:** neutralising the new branch (`else if false`) fails `comment_pass_is_complete,_not_a_retry` specifically, and only that case.

`gofmt`, `go vet ./...`, and `go test -race ./tui/` all clean.

## Related

Same class as #1178, which corrected this exact branch once before for turn-budget exhaustion — `tui/detail_test.go` still carries a comment referencing that fix. This is the second reading of `Completed` that turned out not to mean what the renderer assumed.

Filed from a local worktree as a small self-contained fix, with no issue behind it.

https://claude.ai/code/session_01WGkxr9LQo5Lme39rfQMJ3W

### DIFF
--- a/tui/detail.go
+++ b/tui/detail.go
@@ -16,6 +16,7 @@ type DetailItem struct {
 	StageName      string
 	StageModel     string
 	IsActive       bool // true for in-flight jobs, false for history entries
+	IsComment      bool // comment-processing invocation rather than a stage run
 	Elapsed        time.Duration
 	Duration       time.Duration
 	Success        bool
@@ -71,6 +72,12 @@ func (d DetailPanelComponent) View(width int) string {
 		} else if !item.Completed {
 			if item.TurnLimited {
 				statusStr = "incomplete (turn limit)"
+			} else if item.IsComment {
+				// Comment processing never emits FABRIK_STAGE_COMPLETE by design —
+				// see the same guard in tui/history.go's rebuildViewportContent for
+				// the full reasoning. "incomplete" would be wrong for every comment
+				// pass, not just some of them.
+				statusStr = "success"
 			} else {
 				statusStr = "incomplete"
 			}
--- a/tui/detail_test.go
+++ b/tui/detail_test.go
@@ -65,6 +65,51 @@ func TestDetailPanel_TurnLimitStatus(t *testing.T) {
 	}
 }
 
+// TestDetailPanel_CommentPassStatus is the detail-panel half of the comment-pass
+// fix (the history-pane half is TestViewHistory_TurnLimitClassification's
+// comment cases). The two are separate implementations rather than shared code,
+// so each needs its own coverage — a regression in one would not be caught by
+// the other.
+//
+// Comment processing never emits FABRIK_STAGE_COMPLETE: every fabrik-*-comment
+// skill forbids it ("Do NOT output FABRIK_STAGE_COMPLETE ... returns control to
+// the engine without advancing the pipeline"). Completed is therefore false for
+// every comment invocation by design, and reporting that as "incomplete"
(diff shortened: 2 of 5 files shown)


arbeithand:
Fixed in 7d975533 — the finding was correct and the gap was mine.

I wrote three cases for \`history.go\` and none for \`detail.go\`, despite them being separate implementations rather than shared code. The review is right that a future edit could have regressed the detail panel with nothing catching it.

\`TestDetailPanel_CommentPassStatus\` now covers the same three cases against the detail panel's own rendering:

- a completed comment pass renders `Status:   success`, and explicitly asserts `incomplete` is absent
- a turn-capped comment pass still renders `incomplete (turn limit)` — the comment skills carry their own "If You Hit the Turn Limit" section, so that state is real for them and must not be swallowed by the new branch
- a stage run without the marker still renders `incomplete`, unchanged

Demonstrated non-vacuous: neutralising `detail.go`'s `IsComment` branch (`else if false`) fails `TestDetailPanel_CommentPassStatus` specifically, and only that test.

`gofmt`, `go vet ./tui/`, and `go test -race ./tui/` clean.

Also noting the upstream trace in the review — following `IsComment` through `engine/comments.go` → `tui/events.go` → `tui/model.go`/`tui/active.go` to confirm it is populated consistently — was the right thing to check before trusting the flag, and more than I had verified myself.

Issue #1644, opened 2026-08-23
feat(merge-train): skip the trial for a single up-to-date member — land its PR directly


arbeithand:
## Problem

A merge-train batch containing a **single member** pays the full trial cycle — fork a trial
worktree off the pinned base, merge the member, push the branch, open a draft CI PR, wait out
a complete CI run, land, then clean up — even when that trial is guaranteed to produce a tree
**byte-identical** to one CI has already validated.

`landSingleton` exists but is only a *fallback* after a batch fails
(`engine/merge_train.go:1148`, "one-at-a-time fallback"). A lone Queued member takes the
normal `assembleAndValidate` path.

### The correctness argument

The trial branch is `pinned_base + member`. If the **pinned base SHA is already an ancestor of
the member's head**, that merge is a fast-forward — so the trial's tree is identical to the
member's own PR head tree, which its own CI already ran against. The trial spends a full CI
cycle re-deriving a result already in hand.

**This is the common case, by construction.** Fabrik's own convention has Validate rebase onto
the base branch before signalling completion, so an item arriving fresh from Validate is
typically already up to date with base.

**When the trial does earn its keep:** base moved after that rebase. Then the trial tests a
combination nothing has tested, and skipping it would land code never validated against the
tip it is merging into — the entire point of a merge train. So the fast path must be guarded
by positive evidence, never assumed.

### Measured evidence (verveguy/liminis-context-graph, 2026-08-23)

Issue #477, a single-member batch, observed end to end:

```
21:26:06  pinned base main (37aab7b4)
21:26:06  assembled 1 train member(s)
21:26:07  merged #477 cleanly into trial branch
21:26:09  opened draft CI PR #479 (1 survivor)
21:26:35  fast checks resolve (host, announce, build-global-artifacts -> skipped)
21:37:09  the real suite finally STARTS  <-- 11 minutes of Actions runner queueing
21:41:16  test + mcp_real_corpus_mutation_e2e still IN_PROGRESS
```

15+ minutes elapsed, of which **11 were runner queue wait**, to re-validate a tree that had
already passed CI on the member's own PR. On this repo the suite is a real-corpus e2e battery,
so the cost is dominated by queueing and is paid **twice per issue** under merge-train: once
for the member's PR, once for the trial.

Note the engine behaved *correctly* throughout — ADR-1153 requires check completeness before
declaring green, and `pollTrainCI` is bounded by `ciBackstopTimeout`. This is not a bug
[...]

Issue #1622, opened 2026-08-16
test(merge-train): cover the R8 fail-closed guard so it cannot rot unnoticed


arbeithand:
## Problem

`isTrainPR`'s R8 companion guard — the fail-closed re-check immediately before the
reconstruct sweep's destructive `CloseIssue` call (`engine/merge_train.go`, the
`head ref %q is not a train branch, not closing` branch) — **has no test coverage**.
It can be deleted and the entire suite stays green.

Verified by neutralization on `3984dc3a` (the #1615 fix, now merged as `7a1c9623`):

| mutation | result |
|---|---|
| revert R7 only (restore `isTrainPR`'s marker `\|\|` arm) | 🔴 `TestIsTrainPR` |
| **remove the R8 guard only** | 🟢 **entire suite green** |
| revert both (reproduces the #1615 incident) | 🔴 `TestIsTrainPR` + `TestReconstructTrainState_MarkerOnlyNonTrainBranchPR_NeverClosed` |

R8's stated purpose, in its own code comment, is to be redundant *today* so that a
future change to `isTrainPR` or to the reconstruct loop "cannot silently re-open the
#1615 incident by letting an unrecognized PR reach the close call." That value is
entirely conditional on R8 still being there when R7 regresses. As it stands, R8 can
rot out of the codebase — deleted in a refactor, or dropped in a merge resolution —
with nothing failing. The belt is tested; the braces are not. A backstop nobody can
notice the loss of is not a backstop.

This is not hypothetical: #1615 happened because a second call site of the same
marker-matching pattern was never checked. The failure mode here is the same shape —
an unverified assumption about a defence that is not exercised.

## Summary

Extract the R8 guard's predicate — "is this PR's head ref a genuine
`fabrik/merge-train/…` branch?" — out of the reconstruct sweep's inline `if` into a
small, directly testable pure function, called from the same call site with identical
behavior. Add direct unit tests for that predicate (including the exact #1615 shape:
a non-train head ref whose body quotes the batch marker) so that deleting the guard's
identity check turns a test red, and add a test confirming R8 is independently
effective even when R7 is reverted. This is a testability refactor only — no change to
`isTrainPR`, no change to when PRs are closed, no production seam or injection point.

## Requirements

- **R1** — Extract the guard's predicate into a small, directly testable pure function
  (e.g. `shouldCloseStaleTrainPR(pr gh.PRDetails) bool`), called from the reconstruct
  loop at the same point, with identical behaviour.
[...]

PR #1690, opened 2026-09-03
test: close #1684's orchestrator wiring gap, record guard-testing convention (R1–R4)

### LINKED ISSUE #1684: fix(e2e): detect competing API consumers and a missing review bot before the gate spends budget; document the up/down contract
## Problem

Two of the release gate's operational requirements exist only in operators' heads, and getting
either wrong costs an hour or more per mistake:

1. **Which processes must be UP.** The e2e suite depends on the review bot (Pruefer) to answer
   its review-gated scenarios. If it is down, PRs sit with zero reviews and scenarios fail ~10
   minutes in with a timeout that looks like a Fabrik defect.
2. **Which processes must be DOWN.** The bed and the harness share one `arbeithand` PAT with a
   5,000/hour GraphQL bucket, and a single live leg consumes ~4,000. Any other poller on that
   token pushes the run into backoff, which invalidates it — *after* it has already spent an
   hour and passed.

Both were hit during the v0.0.81 cut, and one directly caused the other: Pruefer was stopped to
free budget, which broke every review-gated scenario. The oscillation between "stop it for
budget" and "start it for reviews" is the symptom.

## Summary

Add two preflight checks to the release gate's startup path (ahead of `scripts/e2e/run.sh`'s
existing bed preflight/pregate steps, before any live API spend): one that detects a competing
consumer of the shared `arbeithand` GraphQL token, and one that verifies the review bot is
reachable before a run that needs it. Document the operational up/down contract and measured
budget numbers in `tests/e2e/README.md` so the tradeoff is explicit instead of tribal knowledge.

## Requirements

- **R1** — Detect and report competing consumers at startup. Before the gate spends live budget,
  check whether another process is polling GitHub on the same token and warn (or refuse) with a
  specific message naming what it found. Discovering this 78 minutes in via backoff is the
  failure mode to eliminate.
- **R2** — Check the review bot is reachable before starting a run that needs it, and fail fast
  with a clear message if not. A run that cannot possibly pass its review-gated scenarios should
  not spend an hour discovering that.
- **R3** — Document the operational contract in `tests/e2e/README.md`: which processes must be
  up, which must be down, why, and the measured numbers (~4,000 GraphQL points per leg against
  a 5,000/hour shared bucket). Include the trap explicitly — stopping the review bot to save
  budget breaks the suite, and probably saves nothing, since Pruefer authenticates as a GitHub
  App with per-installation buckets rather than drawing on the user PAT.

## Scope

**In scope:**
- R1/R2 checks integrated into the existing gate startup path (`scripts/e2e/run.sh`), ahead of
  any live GitHub/Claude call — the same "before it spends budget" position the existing
  sim+wire-contract pre-gate (#1454) and bed preflight already occupy.
- Following the repo's established preflight convention, each check should get a named,
  documented escape hatch env var for iteration (mirroring `E2E_SKIP_PREGATE`,
  `E2E_SKIP_PREP`) rather than being unconditionally mandatory — Research/Plan to determine
  the concrete flag names and default (warn vs. refuse).
- R3 documentation in `tests/e2e/README.md`.

**Out of scope (deferred to #1678):**
- Giving the bed its own credential so it no longer shares a GraphQL bucket with the harness
  token at all (blocked on #770 — GitHub App-based auth). This issue only covers what can be
  done without that structural fix.
- Any change to Pruefer's own authentication model (it already uses per-installation GitHub App
  buckets, not the user PAT — this issue only checks its reachability, not its billing).

**Assumption (stated, not a blocking question):** R2's review-bot-reachable check applies to a
full/default gate run (which includes review-gated scenarios). Whether a narrowed `-run` subset
that provably excludes all review-gated scenarios should skip the check is left to Research to
determine against AC4 (no false positives on a correctly-prepared run) — the check must not
block a legitimately review-bot-free scenario subset.

## Acceptance Criteria

- **AC1** — Starting a gate run with a competing poller on the same token produces an explicit
  warning or refusal naming it, before any live API spend. Proved non-vacuous by removing the
  check and observing the silent start return.
- **AC2** — Starting a run with the review bot unreachable fails fast with a diagnostic, rather
  than failing scenarios ~10 minutes later on review timeouts.
- **AC3** — `tests/e2e/README.md` states the up/down contract and the measured budget figures.
- **AC4** — Neither check produces false positives on a correctly-prepared run.

## Risks / Dependencies

- The structural fix (bed's own credential, #770/#1678) is blocked on GitHub App-based auth and
  is explicitly not this issue's job — these checks are a mitigation for the shared-token window
  that remains until that lands.
- Detection of "a competing consumer on the same token" has no direct GitHub API for "who else is
  polling" — the mechanism (e.g., rate-limit consumption sampling, process-level detection) is a
  Research-stage decision, not prescribed here.

## Notes

Split from #1678 so it is not blocked behind #770. R3 alone would have saved two of the eleven
attempts the v0.0.81 cut required.

### LINKED ISSUE #1687: test: guards are deletable without failing a test — three instances of the same wiring-vs-logic blind spot
## Problem

Three times in one working session, a guard was **deletable with no test failing** — and in
every case the pattern was identical: the guard's *logic* was thoroughly tested, while the
*wiring that invokes it* was not.

| issue | the guard | mutation that stayed green |
|---|---|---|
| **#1622** | merge train's R8 fail-closed identity check | deleting the guard entirely |
| **#1644** | singleton fast path's base-ancestry check (AC2) | deleting `if behind > 0 { return false }` |
| **#1684** | `check_competing_token_consumers` orchestrator | replacing the function body with `return 0` |

Each was found only by neutralization — running the mutation the test claims to catch and
observing it stay green. None would have been caught by reading the tests, and all three test
suites looked thorough.

### Why it keeps happening

The failures share a shape:

1. **The unit is tested; the call site isn't.** #1684's detection helper has six well-chosen
   cases; the function that *acts on* the detection has none that would fail if it stopped
   acting.
2. **Assertions check the outcome, not the reason.** #1644's `..._BaseNotAncestor_FallsThrough`
   asserted `eligible == false` — true for *any* rejection, so the zero-check-runs arm satisfied
   it and masked the ancestry arm entirely. It passed for the wrong reason.
3. **Defence-in-depth hides itself.** #1622's R8 was deliberately redundant with R7, so removing
   either alone changed nothing observable. The redundancy was the point — and it made the loss
   of one layer undetectable.

Item 2 is the most insidious: those tests were *written* to cover the case and genuinely
believed to.

Not a defect in any shipped behaviour — every guard above is *correct*. This is about whether
we would notice if one stopped being correct. Given #1614 happened because a second call site
of a known-bad pattern went unchecked, "we would not have noticed" is the failure mode worth
spending on. The neutralization discipline itself is the recommendation's core: none of these
were visible by inspection, and all three were obvious within minutes of running the mutation.

## Summary

Record a durable testing convention — assert the rejection *reason*, not just the outcome;
test guard *wiring*, not only guard *logic*; make deliberate redundancy either directly
testable or explicitly documented as unreachable-by-construction; and neutralize any new guard
in review — then apply it in two concrete places: close the still-open #1684 orchestrator gap,
and spot-check existing `_FallsThrough`/`_Rejects`-style tests for the same "outcome, not
reason" blind spot.

## Requirements

- **R1 — Assert the reason, not just the outcome.** For any test whose purpose is to prove a
  specific guard rejects, the assertion must distinguish that guard's rejection from every
  other possible rejection reason — e.g. via a shared fixture that makes every earlier
  condition pass so the guard under test is the only thing that can reject, plus an assertion
  on the returned reason/error string. (#1644's `..._BaseNotAncestor_FallsThrough` fix is the
  model.)
- **R2 — Make deliberate redundancy accountable.** When a guard is deliberately redundant with
  another (defence-in-depth), the code or its test must state how the guard's loss would be
  detected. If the honest answer is "it wouldn't," either extract it into a directly-testable
  predicate (#1622's `shouldCloseStaleTrainPR` is the model) or record explicitly that it is
  unreachable-by-construction and therefore untested.
- **R3 — Test the wiring, not only the unit.** Every guard needs at least one test case where
  the orchestrator/call site is exercised end to end and the guard's effect is observable —
  not only a unit test of the guard's own logic in isolation.
- **R4 — Neutralization in review.** Any change that adds a guard must, in the PR description
  or review notes, state which mutation the new test catches and show that mutation turning
  the test red.

## Scope

**In scope:**
- Recording R1–R4 as guidance in a location authors will actually see before/while writing
  guard tests (e.g. `tests/sim/README.md` and/or the `fabrik-review` skill), including the
  three worked examples (#1622, #1644, #1684) above.
- Closing the #1684 gap specifically: adding a wiring-level test for
  `check_competing_token_consumers` (`scripts/e2e/run.sh`) such that replacing its body with a
  no-op turns at least one existing test red.
- A spot-check (not an exhaustive audit) of existing tests following the
  `_FallsThrough`/`_Rejects` naming convention, checking each for the "asserts outcome, not
  reason" shape described in R1, with findings written down (in the issue, a comment, or a
  tracked follow-up — implementer's choice).

**Out of scope:**
- Fixing every `_FallsThrough`/`_Rejects` test found during the AC3 spot-check — the issue
  explicitly allows recording findings and treating fixes as follow-up work.
- Applying R1–R4 retroactively to every existing guard in the codebase beyond the AC3
  spot-check and the AC2 fix.
- Any change to the behavior of `check_competing_token_consumers`, the merge-train R8 check,
  or the singleton fast path's ancestry check — all three are already correct; only their test
  coverage is in scope.

## Acceptance Criteria

- **AC1** — Guidance recorded where authors will see it (`tests/sim/README.md` or the review
  skill, or both), including the three worked examples above (#1622, #1644, #1684).
- **AC2** — The #1684 orchestrator gap is closed: neutering `check_competing_token_consumers`
  (replacing its body with a no-op / `return 0`) turns at least one test red.
- **AC3** — A spot-check of existing `_FallsThrough`/`_Rejects`-style tests for the "asserts
  outcome, not reason" shape is performed, with findings recorded (fixing them can be
  follow-up work).

## Prior Art / Context

- `check_competing_token_consumers` lives in `scripts/e2e/run.sh` (currently exercised by
  `scripts/e2e/token_consumer_check_test.sh`, a shell-based test suite, not Go — the fix will
  need to work within that existing test harness).
- A repo-wide search finds 36 occurrences of `_FallsThrough`/`_Rejects`-style test names across
  9 files (`engine/merge_train_test.go`, `engine/catch_up_handlers_test.go`,
  `engine/ci_test.go`, `engine/reviews_test.go`, `engine/reviews.go`, `engine/worktree_test.go`,
  `github/contents_test.go`, `cmd/init_test.go`, `adrs/1518-review-gate-non-convergence-terminal-check.md`)
  — a reasonable starting set for the AC3 spot-check, though the check is explicitly a sample,
  not a requirement to cover all of them.
- Candidate homes for AC1's guidance: `tests/sim/README.md` (already documents the sim-bed
  testing philosophy) and `plugin/fabrik-workflows/skills/fabrik-review/SKILL.md` (the review
  skill, read by every Review-stage invocation).

## Risks / Dependencies

- None of the three underlying guards (#1622, #1644, #1684) need behavioral changes — this
  issue is test-coverage and documentation only, which keeps risk low.
- The spot-check (AC3) could surface additional gaps beyond the three named here; per Scope,
  recording those findings satisfies AC3 even if fixing them is deferred.

arbeithand:
Closes #1687

## What this does

Addresses #1687 — a recurring pattern found across three prior issues (#1622, #1644, #1684) where a guard was deletable with no test failing, because the guard's *logic* was tested but the *wiring that invokes it* wasn't, or the test asserted the outcome rather than the specific rejection reason. This PR closes the one remaining open gap (#1684) and records the convention so it's caught earlier next time.

**AC2 — closes the #1684 orchestrator gap** (`scripts/e2e/token_consumer_check_test.sh`): adds Case 9, a wiring-level test of `check_competing_token_consumers` (`scripts/e2e/run.sh`). Cases 1–6 already covered `find_competing_token_consumers` (the pure detection unit) in isolation; cases 7–8 only exercised orchestrator paths that never reach the match/refuse logic. Nothing proved the orchestrator itself calls `discover_fabrik_process_dirs`, feeds the result through `find_competing_token_consumers`, and exits `PRECONDITION_FAILED_EXIT` on a match. Case 9 overrides the untestable OS-dependent half (`discover_fabrik_process_dirs`) with a canned candidate, then asserts the orchestrator's own exit code and refusal output (naming the failure, the competing PID, and the competing directory).

**Self-applied R4 (neutralization in review):** verified manually — replacing `check_competing_token_consumers`'s body with `return 0` turns Case 9 red (all four of its assertions fail) while cases 1–8 stay green; reverting restores a clean pass. No behavioral change to `check_competing_token_consumers` itself.

**AC1 — guidance recorded** in two places, split by audience:
- `tests/sim/README.md`: new "Guard-testing convention" section stating R1 (assert the reason, not the outcome), R2 (make deliberate redundancy accountable), and R3 (test the wiring, not only the unit), each with its worked example (#1644's `TestSingletonFastPathEligible_BaseNotAncestor_FallsThrough`, #1622's `shouldCloseStaleTrainPR`, and this PR's Case 9), plus a pointer to `simgh/nonvacuity.sh` as the same neutralization discipline already applied at package scale.
- `plugin/fabrik-workflows/skills/fabrik-review/SKILL.md`: four new bullets in the Testing checklist (R1–R4), with R4 as the actionable review-time check — cross-referencing the README section rather than duplicating the worked examples.

**AC3 — spot-check already recorded**: the [Research stage comment on #1687](https://github.com/handarbeit/fabrik/issues/1687#issuecomment-thread) reviewed all 36 `_FallsThrough`/`_Rejects`-named tests across the 9 files the issue names, and flagged three candidates showing the "asserts outcome, not reason" shape (`github/contents_test.go`'s two `Rejects*` tests, `cmd/init_test.go`'s `TestRunInit_RejectsInvalidURL`, `engine/ci_test.go`'s `TestCheckCIGate_MergeableStateBlocked_FallsThroughToCheckRuns`). Per the issue's own Scope, fixing those is available follow-up work, not required here.

## No behavior changes

None of the three underlying guards (#1622's R8 identity check, #1644's ancestry check, #1684's `check_competing_token_consumers`) were modified — this is test-coverage and documentation only, confirmed by Research/Plan and reflected in the diff (no Go source touched at all).

## How to test

```bash
bash scripts/e2e/token_consumer_check_test.sh   # all 9 cases, standalone
go vet ./...
go test -race ./...
```

Note: `go test -race ./...` has one pre-existing, unrelated failure — `TestExecute_ConfigYAMLApplied` in `cmd/` (a live-network-dependent SIGINT-shutdown test hitting real GitHub API 404s) — reproducible identically on `main` at the branch point (`c0961347`), since this PR touches no Go source.

### DIFF
--- a/plugin/fabrik-workflows/skills/fabrik-review/SKILL.md
+++ b/plugin/fabrik-workflows/skills/fabrik-review/SKILL.md
@@ -135,6 +135,22 @@ git diff "origin/$base_branch"..HEAD
 - Do tests cover error paths, not just happy paths?
 - Are tests actually testing behavior, not just exercising code?
 - Run the test suite: do all tests pass?
+- (#1687, R1) If a test's purpose is to prove a specific guard rejects, does its
+  assertion distinguish that guard's rejection from every other possible rejection
+  reason — not just a boolean/exit-code outcome that any rejection would satisfy? See
+  `tests/sim/README.md`'s "Guard-testing convention" for the worked example and fix
+  pattern.
+- (#1687, R2) If a guard is deliberately redundant with another (defence-in-depth), does
+  the code or its test state how the guard's loss would be detected — either via a
+  directly-testable extracted predicate, or an explicit unreachable-by-construction note?
+  See `tests/sim/README.md`'s same section.
+- (#1687, R3) Does every new guard have at least one test exercising its call site end to
+  end (not only a unit test of the guard's own logic in isolation) — proving the
+  orchestrator actually invokes the guard and observably acts on its result?
+- (#1687, R4 — neutralization in review) For any change adding a new guard: does the PR
+  description or review notes state which mutation the new test catches? Confirm that
+  mutation was actually run against the guard and observed to turn the test red — not
+  merely reasoned about. A guard test that hasn't been neutralized hasn't been trusted.
 
 **Security**:
 - No command injection, SQL injection, XSS, or path traversal
--- a/scripts/e2e/token_consumer_check_test.sh
+++ b/scripts/e2e/token_consumer_check_test.sh
@@ -112,6 +112,47 @@ assert_eq "E2E_SKIP_TOKEN_CHECK=1 exits 0" "0" "$rc"
 rc=$?
 assert_eq "empty BED_TOKEN degrades to a pass (warning only)" "0" "$rc"
 
+# --- Case 9 (#1687, R3): wiring-level coverage of
+# check_competing_token_consumers' match->refuse path itself. Cases 1-6
+# above only prove find_competing_token_consumers (the pure detection unit)
+# works in isolation; cases 7-8 only exercise check_competing_token_consumers
+# paths that never reach the match/refuse logic. Nothing before this proved
+# the orchestrator actually calls discover_fabrik_process_dirs, feeds its
+# output through find_competing_token_consumers, and exit()s
+# PRECONDITION_FAILED_EXIT on a positive match — replacing
+# check_competing_token_consumers' body with `return 0` left every prior
+# case green. Here discover_fabrik_process_dirs (the untestable,
+# OS-dependent half) is overridden with a canned candidate so the
+# orchestrator's own match-handling can be exercised deterministically,
+# without a real competing process.
+#
+# Left intentionally un-restored: this is the last case in the file, so the
+# override can't leak into a later one — a future editor appending Case 10+
+# after this should redefine discover_fabrik_process_dirs back to its
+# original body first (or move this case to the end again).
+discover_fabrik_process_dirs() { printf '9999\t%s\n' "$OTHER_DIR"; }
+out="$( ( TEST_BED="$BED_DIR" BED_TOKEN="shared-token-abc" check_competing_token_consumers ) 2>&1 )"
+rc=$?
+assert_eq "orchestrator match: exits PRECONDITION_FAILED_EXIT" "$PRECONDITION_FAILED_EXIT" "$rc"
+if ! printf '%s' "$out" | grep -q "PRECONDITION FAILED"; then
+  echo "FAIL: orchestrator match: output names PRECONDITION FAILED"
+  FAILED=1
+else
+  echo "PASS: orchestrator match: output names PRECONDITION FAILED"
+fi
+if ! printf '%s' "$out" | grep -q "9999"; then
+  echo "FAIL: orchestrator match: output names competing PID"
+  FAILED=1
+else
+  echo "PASS: orchestrator match: output names competing PID"
+fi
+if ! printf '%s' "$out" | grep -qF "$OTHER_DIR"; then
+  echo "FAIL: orchestrator match: output names competing dir"
+  FAILED=1
+else
+  echo "PASS: orchestrator match: output names competing dir"
+fi
+
 if [ "$FAILED" -ne 0 ]; then
   echo "=== token_consumer_check_test.sh: FAILED ==="
   exit 1
--- a/tests/sim/README.md
+++ b/tests/sim/README.md
@@ -837,6 +837,77 @@ first, not a defect in the refund mechanism itself.
   `CommentReviewCompleted` (both of which commit), mirroring
   `simclaude/scripts.go`'s existing naming convention.
 
+## Guard-testing convention (R1–R4, #1687)
+
+Three times in one working session, a guard (a check that rejects, refuses, or falls
+through under some condition) was deletable with no test failing. In every case the
+guard's *logic* was thoroughly tested while the *wiring that invokes it* was not — the
+neutralization discipline this package's own `nonvacuity.sh` (below) already applies at
+package scale, just never asked of an individual guard's own test. None of the three
+underlying guards were ever wrong; this is about whether a regression in one would be
+noticed, and none would have been caught by reading the tests — every suite looked
+thorough until someone actually ran the mutation.
+
+The convention below (R1–R3; R4 is a review-time obligation and lives in the
+`fabrik-review` skill instead) applies identically whether the guard is a Go function
+returning `(bool, string)` or a bash function with an exit code and stdout — it is not
+specific to this package's own Go-and-`simgh` regime.
+
+**R1 — Assert the reason, not just the outcome.** A test proving a specific guard
+rejects must distinguish that guard's rejection from every other possible rejection
+reason — e.g. via a shared fixture that makes every earlier condition pass, so the guard
+under test is the only thing that can reject, plus an assertion on the returned
+reason/error string, not only the boolean/exit-code outcome.
+
+*Worked example:* `engine/merge_train_test.go`'s
+`TestSingletonFastPathEligible_BaseNotAncestor_FallsThrough` (part of the
+`singletonFastPathEligible` battery, `engine/merge_train.go`) originally asserted only
+`eligible == false` — true for *any* rejection. Its own fixture defaulted to zero check
+runs, which rejects via an unrelated arm, so deleting the ancestry guard's own
+`if behind > 0 { return false }` left the test green: it passed for the wrong reason. The
+fix (now in place) rebuilds the fixture so every other guard passes and the ancestry
+check is the only thing that can reject, then asserts `strings.Contains(reason, ...)` on
+the specific rejection string. Every sibling test in that same battery follows the
+identical shape.
+
+**R2 — Make deliberate redundancy accountable.** When a guard is deliberately redundant
+with another (defence-in-depth), the code or its test must state how the guard's loss
+would be detected. If the honest answer is "it wouldn't," either extract it into a
+directly-testable predicate, or record explicitly that it is unreachable-by-construction
+and therefore untested.
+
+*Worked example:* the merge train's R8 fail-closed identity check
+(`shouldCloseStaleTrainPR`, `engine/merge_train.go`) is deliberately redundant with R7 —
+removing either alone changes no observable behavior, which is the point, and which made
+the loss of one layer undetectable by outcome alone. It was extracted into its own
+directly-testable predicate specifically so it can be asserted in isolation:
+`engine/merge_train_test.go`'s test carries the comment "deleting
+`shouldCloseStaleTrainPR`'s identity check turns this test red (AC1)" — stating exactly
+how the guard's loss would be caught, per this rule.
+
+**R3 — Test the wiring, not only the unit.** Every guard needs at least one test case
+where the orchestrator/call site is exercised end to end and the guard's effect is
+observable — not only a unit test of the guard's own logic in isolation.
+
+*Worked example:* `scripts/e2e/run.sh`'s `check_competing_token_consumers` orchestrator
+wraps a well-tested pure detection function (`find_competing_token_consumers`, six cases
+in `scripts/e2e/token_consumer_check_test.sh`) but, until #1687, had no test exercising
+its own match→refuse path — replacing its body with `return 0` left every existing case
+green. The fix added a case that overrides the untestable OS-dependent half
+(`discover_fabrik_process_dirs`) with a canned candidate, then asserts the orchestrator
+itself detects the match and exits `PRECONDITION_FAILED_EXIT` — proving the wiring, not
+just the detection logic it calls.
+
+**Neutralize before trusting a guard test.** The only way any of the three gaps above
+were found was by actually running the mutation the test claims to catch and observing
+whether it stayed green — not by reading the test. `tests/sim/simgh/nonvacuity.sh`
+applies the identical discipline at package scale (neutralize each modeled `simgh`
+behavior in turn, fail loudly if any leaves the suite green); treat a new guard test the
+same way before trusting it. The review-time version of this obligation (R4 — state which
+mutation a new guard's test catches, and show it turning the test red) is a
+`fabrik-review` skill checklist item, not a sim-bed one, since it's an author obligation
+applied during review, not a testing-layer property.
+
 ## Running it
 
 ```bash


arbeithand:
*branch: fabrik/issue-1687 | commit: decf8703 | main: c09613470457ea686f1c895a8b6ed739b5728f88 | 2026-09-03 03:21 UTC*

Everything is committed, pushed, and clean. Now let's emit the PR creation marker.


Closed the #1684 orchestrator wiring gap by adding a Case 9 test that exercises `check_competing_token_consumers`'s match→refuse path end to end (verified: neutering it to `return 0` turns Case 9 red, confirmed manually then reverted). Recorded the R1–R4 guard-testing convention with worked examples in `tests/sim/README.md` and four review-checklist bullets in the `fabrik-review` skill. AC3's spot-check was already recorded in Research's stage comment; all Plan tasks are checked off and pushed.

---
Used 41/100 turns, 4.3M input (78 raw + 4.2M cache-read + 120k cache-write) / 14k output tokens.

arbeithand:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-1687 | commit: decf8703 | main: c09613470457ea686f1c895a8b6ed739b5728f88 | 2026-09-03 03:31 UTC*

Everything is clean, committed, and pushed. Nothing further to do.

## Review Findings

### Fixed
- None needed. This PR is documentation- and shell-test-only (no Go source touched, matching Plan's stated scope), and every claim it makes checked out on independent verification.

### Verified
- Base branch (`main`) confirmed via `gh pr view`; branch already up to date with `origin/main` (0 commits behind) — no rebase needed.
- No external review-bot/human PR comments to address yet.
- `git diff origin/main..HEAD` reviewed in full: `scripts/e2e/token_consumer_check_test.sh` (+41), `tests/sim/README.md` (+71), `plugin/fabrik-workflows/skills/fabrik-review/SKILL.md` (+16).
- Ran `bash scripts/e2e/token_consumer_check_test.sh` standalone: all 9 cases pass, including the new Case 9 wiring test for `check_competing_token_consumers`.
- **Independently re-verified the PR's own R4 neutralization claim**: temporarily replaced `check_competing_token_consumers`'s body with `return 0`, reran the suite — Case 9's four assertions all failed (rc `0` vs expected `7`, and all three output-content checks), cases 1–8 stayed green as expected. Reverted cleanly; full suite passes again. This is exactly the mutation the PR claims to catch, confirmed turning red.
- `go build ./...` and `go vet ./...`: clean.
- `go test -race -timeout 5m ./...`: one failure, `TestMergeTrainRunaway_DoesNotTripWithGenerousWindow` in `tests/sim`. Re-ran it in isolation (`go test -race -timeout 150s ./tests/sim/ -run TestMergeTrainRunaway_DoesNotTripWithGenerousWindow -v`) — passes standalone in 40s. Given this PR touches zero Go code, this is pre-existing timing-sensitive flakiness under full-suite `-race` contention, not a regression introduced here (consistent with the PR description's own note about a separate pre-existing `cmd/` failure on `main`).
- Cross-checked documentation accuracy against the actual code: `shouldCloseStaleTrainPR` at `engine/merge_train.go:3043` (comment block from :3030) and its "AC1" test comment at `engine/merge_train_test.go:5663` match the README's citation exactly. The `## Guard-testing convention (R1–R4, #1687)` section is correctly placed immediately before `## Running it` in `tests/sim/README.md`, and the SKILL.md bullets cross-reference it by the correct heading text.
- Confirmed `docs/llms-full.txt` regeneration is correctly not required — `scripts/generate-llms-full.sh` only bundles `USER_GUIDE.md`, `state-machine.md`, `stage-lifecycle.md`, `positioning.md`; neither file this PR touches is among them.
- AC3's spot-check (recorded in the Research stage comment on #1687, referenced from the PR description) satisfies the issue's "findings recorded, not necessarily fixed" scope — no further action needed here.

### Blocking
None.

Reviewed the #1687 PR (#1690): guard-testing convention docs (R1–R4) plus the AC2 wiring test for `check_competing_token_consumers`. Independently re-verified the neutralization claim (neutering the orchestrator to `return 0` does turn the new Case 9 red) and cross-checked all doc citations against the actual code — all accurate. Build/vet clean, shell suite passes, one pre-existing timing-flaky Go test (`TestMergeTrainRunaway_DoesNotTripWithGenerousWindow`, unrelated — this PR touches no Go code, and it passes standalone). No fixes needed; PR is ready for human review.

---
Used 34/100 turns, 2.8M input (52 raw + 2.7M cache-read + 113k cache-write) / 7k output tokens.

arbeithand:
*branch: fabrik/issue-1687 | commit: decf8703 | main: c09613470457ea686f1c895a8b6ed739b5728f88 | 2026-09-03 03:31 UTC*

Pruefer's review is a `[Bot Review Finding]` that traced through Case 9, the CI wiring, the README/SKILL.md consistency, and the worked-example citations, and explicitly found **no defects**. There's nothing actionable here — per the No-Op Contract, the correct response is to change nothing. This matches my own independent verification during the Review stage (I re-ran the neutralization mutation myself and confirmed it turns Case 9 red, and cross-checked every doc citation against the actual code).

No issue body changes are needed — Pruefer's comment doesn't surface any new requirement or scope change.

---
**Threads addressed:**

arbeithand:
*branch: fabrik/issue-1687 | commit: decf8703 | main: c09613470457ea686f1c895a8b6ed739b5728f88 | 2026-09-03 03:43 UTC*

All checks pass. Working tree is clean, everything committed and pushed, no rebase needed, PR is mergeable/clean, and no leftover TODOs. Final validation report:

## Validation Report

### Requirements: 3/3 passed
- [x] **AC1** — Guidance recorded: `tests/sim/README.md` gained a "Guard-testing convention (R1–R4, #1687)" section with all three worked examples (#1622, #1644, #1684) verbatim-consistent with the issue; `fabrik-review/SKILL.md`'s Testing checklist gained 4 cross-referencing bullets (R1–R4).
- [x] **AC2** — The #1684 orchestrator gap is closed. Added Case 9 to `scripts/e2e/token_consumer_check_test.sh`, exercising `check_competing_token_consumers`'s match→refuse path via a `discover_fabrik_process_dirs` override. **Independently re-verified the neutralization claim**: temporarily replaced the orchestrator's body with `return 0` — Case 9's 4 assertions all failed (exit code 0 vs expected 7, all 3 output-content checks failed), cases 1–8 stayed green. Reverted cleanly (`git status` confirmed no diff after revert); full suite passes again.
- [x] **AC3** — Spot-check recorded in Research's stage comment on this issue (all 36 candidates reviewed, 3 flagged with specific reasoning); referenced in the PR description per the issue's own acceptance wording.

### Test Suite: PASSED
- `go build ./...`: clean
- `go vet ./...`: clean
- `go test -race -timeout 5m ./...`: all packages pass. (One spurious failure on first run — `TestInvokeClaude_FabrikEnvVarsInjected` — traced to this Validate session's own ambient `FABRIK_PR=1690` env var, set by the outer Fabrik engine orchestrating this very invocation, leaking into the test subprocess; confirmed the test passes cleanly with those ambient vars unset, and this PR touches zero Go code, so it's a self-inflicted artifact of running Validate inside a Fabrik-managed shell, not a regression.)
- `bash scripts/e2e/token_consumer_check_test.sh`: all 9 cases pass standalone.

Diff is additive-only across 3 files (128 insertions, 0 deletions): docs + one new shell test case. No source behavior touched.

### Issues Found
None.

### Verdict: READY TO MERGE

Validation passed. Rebase: skipped-up-to-date. PR mergeable: MERGEABLE, mergeStateStatus: CLEAN. All 3 ACs verified — independently re-confirmed AC2's neutralization claim (Case 9 turns red when the orchestrator is neutered), and cross-checked AC1's doc citations against code. Go suite and shell suite both pass clean; one env-artifact test failure traced to this session's own ambient FABRIK_PR var, not a regression.

---
Used 32/100 turns, 3.3M input (62 raw + 3.2M cache-read + 109k cache-write) / 8k output tokens.

Issue #1275, opened 2026-07-31
merge-train: retry the best-effort 'landed via' comment post on member PR


arbeithand:
## Problem

Both merge-train landing paths post a "landed via ..." comment on the member's own PR to record which PR actually landed the change, then move on regardless of whether the post succeeded:

- `landSingleton` (`engine/merge_train.go:1032`, currently): `AddComment(...)` on failure only `logf`s a `warn:` and continues.
- `landMergeTrainBatch` (`engine/merge_train.go:2104`, currently): same pattern.

(Line numbers have shifted from the original report due to unrelated merges since filing; both call sites and their surrounding logic were re-verified against the current `main` and match the description below exactly.)

Neither call is retried. If it fails transiently (rate limit, brief API hiccup) while the landing itself completes normally, the member's own PR permanently loses the only record of which integration/singleton PR actually landed it — there is no other reliable, cross-landing-path, member-scoped signal to reconstruct this after the fact:

- `closedByPullRequestsReferences` on the member issue doesn't work as a substitute: `landSingleton`'s landing PR body says `"Lands #%d"`, not `"Closes #%d"` (`engine/merge_train.go:998`, currently), so GitHub never links it via the issue's closing-PR references for that path. (The batch path's integration PR body does carry real `Closes #N` lines, but that alone isn't a general solution across both paths.)
- Engine logs aren't a general substitute either: `landSingleton`'s "merged singleton landing PR #%d for #%d" is member-scoped, but the batch path's "merged integration PR #%d for %s" is only repo-scoped — ambiguous under concurrent merge-train activity in the same repo.

So today, a single transient comment-post failure permanently and silently degrades the audit trail for that member (production impact: a user watching their PR loses the "which PR actually landed my change" pointer), and — see #1271 — makes `tests/e2e/auto_merge_test.go`'s `TestYoloAutoMergeLabel` (train-mode-on) unable to distinguish "engine regressed to merge the member's own PR directly" from "landing succeeded but the comment post flaked," since both look like the same test timeout.

## Summary

Wrap the "landed via ..." `AddComment` call in both `landSingleton` and `landMergeTrainBatch` with a small, bounded in-line retry (short backoff, no durable cross-poll state) so transient API failures don't permanently drop the landing-provenance comment on the member PR. If all retries are exhausted, fall back to the existing warn-and-continue behavior unchanged — the comment post must never block or delay the landing itself.

## Requirements

[...]

PR #1692, opened 2026-09-06
fix(merge-train): singleton fast path lands silently — no landed comment, no "landing complete"

arbeithand:
## Problem

#1644's singleton fast path shipped in v0.0.81 as a fourth merge-train landing path, wired into **neither** of the observability channels the other three use:

| channel | emitted by | fast path |
|---|---|---|
| member-scoped `🏭 Fabrik merge-train — Landed …` PR comment (#1275) | `landMergeTrainBatch`, `landSingleton` | **silent** |
| repo-scoped `landing complete for …` log line | `landMergeTrainBatch` only | **silent** |

So the fast path merged PRs correctly and invisibly: no member-scoped record of which PR landed the change or what validated it, and — from the log alone — a train that started and never finished, with nothing on its TUI job row.

Found by the release gate. `TestYoloAutoMergeLabel` failed on the v0.0.81 cut and was written off as "#1275's un-retried cosmetic comment was absent". It was not: the comment was never attempted.

## Changes

**`finishSingletonFastPathLanding` posts the landed comment.** Wording names the fast path and states the evidence that authorized the direct landing. The cited PR is the **member's own**, because under this path that PR *is* the merge vehicle — there is no separate integration PR to name, by construction. Consumers must key off the path named in the text, not off the landing PR being distinct from the member's.

Guarded on `fabrik:awaiting-landing-verification`'s absence — unlike the other two paths, which close the member PR immediately after commenting and so can never re-enter, this one leaves the PR to GitHub's own merge-close and can legitimately be re-entered while the member is still Queued (the `pr.Merged` restart branch, and a retry after a failed advance).

**`finishSingletonFastPathLanding` emits `landing complete`.** Repo-scoped via `logfRepo`, not per-issue: a per-issue line would not reach the train's job row, so routing is part of the contract and is asserted as such. Emitted unconditionally rather than guarded like the comment — a repeat on re-entry is accurate, and is itself the signal that a re-entry happened.

**e2e repair.** `auto_merge_test.go` required the landing PR to be *distinct* from the member's — an invariant #1644 deliberately breaks. It now asserts on the **path**, both directions: self-citation on a non-fast-path landing still fails as a direct-merge regression, and a fast-path comment citing some other PR fails as comment/mechanism disagreement. An unvalidated direct merge posts no comment at all and still fails on timeout, so the guard is not weakened. `landedPRPattern` gains the third alternative; `waitForLandingPRDetail` returns which path posted.

## Verification

Six neutralization mutations, each turning **exactly one** test red:

| mutation | red test |
|---|---|
| comment call deleted | `PostsLandedComment` |
| wording changed to the batch path's | `PostsLandedComment` |
| idempotency guard removed | `AlreadyVerified_SkipsDuplicateComment` |
| landing-complete line removed | `PostsLandedComment` |
| landing-complete routed per-issue | `PostsLandedComment` |
| "singleton fast path" dropped from its text | `PostsLandedComment` |

`go test -race ./...` green. Sim suite (227s) + wire-contract (22s) green.

**Live e2e, isolated, train-on leg, against a bed built from this branch:**

```
auto_merge_test.go:125: member PR #5658 confirmed MERGED via the
                        singleton fast path — train landing contract verified
auto_merge_test.go:131: member PR #5658 reached a terminal state
auto_merge_test.go:134: bed log shows a completed train landing
auto_merge_test.go:137: fabrik:auto-merge-enabled was never applied
--- PASS: TestYoloAutoMergeLabel (626.65s)
```

552 GraphQL points of 5000, no backoff, no `RUN INVALID`.

## Not fixed here

- **`landSingleton` has the same `landing complete` gap** and emits no such line either. Pre-dates #1644; no test depends on it today. Left alone rather than widening scope.
- **`E2E_BED_REF` is broken for any ref but `main`** — `run.sh` documents it as accepting any ref, but the bed checkout is a single-branch clone (`+refs/heads/main:refs/remotes/origin/main`), so `git rev-parse` dies with `fatal: ambiguous argument`. Hand-fetched the branch into the bed for each run above. The fix belongs in `preflight_bed` (fetch the specific ref rather than relying on the refspec).

Docs: `docs/state-machine.md` §6.20 updated, bundle regenerated.


arbeithand:
Addressed in `b370c247` — the finding was correct, and specifically it was a case the previous commit's own code comment *claimed* to cover, so the claim was wrong rather than merely incomplete.

`awaitingLandingVerificationLabel` is applied only after `recordAdvanceOutcome` succeeds, so on a failed advance the member stays Queued with its PR already merged, re-enters via `trySingletonFastPath`'s `pr.Merged` branch each poll, and finds the guard still open — one duplicate per poll until `MaxRetries` escalates to `fabrik:paused`.

The comment is now posted in lockstep with the label: only when this call actually recorded a completed landing (`landed`, set exactly where the label is applied), **and** only when the label is absent from the item's own snapshot. Both halves are load-bearing:

- without `landed`, the failed-advance retry reposts — the reported bug;
- without the label check, the already-Done restart branch (which re-applies the label idempotently on every re-entry) re-comments each poll.

Suppressing the duplicate does not suppress the comment permanently: the retry that eventually advances posts exactly one, and the new test asserts that second half explicitly rather than only asserting the absence of duplicates.

New test `TestFinishSingletonFastPathLanding_FailedAdvanceRetry_PostsNoDuplicate` covers the scenario you named — two polls with a failing `UpdateProjectItemStatus`, labels deliberately left empty across both (which is what the real board snapshot looks like on the retry poll), then a third poll with the advance succeeding.

Neutralization, including the reported bug itself:

| mutation | red test |
|---|---|
| restore the label-absence-only guard | `FailedAdvanceRetry_PostsNoDuplicate` |
| drop the label half | `AlreadyVerified_SkipsDuplicateComment` |
| never set `landed` on the success path | `PostsLandedComment` + `FailedAdvanceRetry_PostsNoDuplicate` |

`go test -race ./...` green. `docs/state-machine.md` §6.20 updated to state the two-part guard and why the label's absence alone is insufficient; bundle regenerated.

PR #1697, opened 2026-09-07
fix(e2e): the release gate budget probe measured nothing; move suite reads to REST

arbeithand:
## Problem

The v0.0.82 gate's **off leg passed all 47 of its tests and was then invalidated** by GraphQL rate-limit backoff. Three separate defects combined to make that both inevitable and invisible.

### 1. The budget report was reading a dead gauge

`run.sh` bracketed each leg with `gh api rate_limit --jq .resources.graphql.remaining`. For the bed's token that REST endpoint reports a permanently full bucket. Measured, same token, seconds apart:

```
inline rateLimit : 4248 remaining, reset 15:44   (5 board queries -> 4243)
REST rate_limit  : 5000 remaining, reset 16:33   (5 board queries -> 5000)
```

The REST figure never moves. **Every `GraphQL budget (leg: X): 5000 -> 5000 remaining (consumed 0 pts)` line this script has ever printed was that dead gauge** — including for the off leg, which tripped the engine's own 20% backoff while the probe still reported a full bucket.

A budget report that cannot go down is worse than no report: it was read as evidence of headroom before launching runs that then exhausted the budget. It is also why the true cost of a leg has never been known.

### 2. Nothing configured the bed engine's poll cadence

The bed ran at the engine's 30s default. From its own inline telemetry: **~1,320 GraphQL points/hour while a leg is active** (~310 idle, where idle-upgrade backoff already stretches the cadence). Over a ~70 minute leg that is ~1,500 points before the suite's own calls are counted.

The v0.0.81 cut set `E2E_POLL_INTERVAL=60s` and still hit backoff — because that only throttles the *suite*, and the engine kept polling at 30s throughout.

### 3. The suite spent GraphQL budget on reads REST serves for free

`gh issue view`, `gh pr view`, `gh issue list`, `gh pr list` are served by GraphQL. `gh api <path>` is REST — **a separate 5,000/hour bucket these runs leave almost entirely idle**. Measured, 10 calls each:

| call | GraphQL cost |
|---|---|
| `gh issue view --json labels` | ~10 pts |
| `gh api repos/{o}/{r}/issues/{n}` | **0 pts** |

~20 scenarios each polling labels every 30–60s across a ~70 minute leg is thousands of points spent on reads REST would serve free.

## Changes

**`c61f25ee`** — `graphql_budget_remaining` replaces both REST probe sites with an inline `rateLimit` query (1 point per call, 2 per leg — the price of the report being true). It takes the token as an argument rather than being wrapped in `env GH_TOKEN=...`, because `env` execs a binary and cannot invoke a shell function; `with_timeout` runs `"$@" &` in-shell and invokes it fine.

Both bed launch sites now pass `-poll` explicitly (`defaultBedPollSeconds=60`, override `E2E_BED_POLL_SECONDS`), and `defaultPollBase` goes 30s → 60s. **Both launch sites must agree**: `switch_and_run` restarts the bed through the Go path on every mode switch, so a cadence set only in `run.sh` would be silently reverted for exactly the legs that matter. Passed as a flag rather than left to `FABRIK_POLL` in the bed's untracked `.env`, so it travels with the repo.

**`efcff197`** — the read paths that run inside wait loops move to REST behind four helpers, so the gh-vs-REST field-shape differences are handled once: `IssueLabels`, `tryIssueState`, `tryIssueLabels`, `IssueState`, `PRBaseRef`, `PRAuthor`, and the merge-train `waitForPRClosed`/`assertPRMerged` state reads. `tryPRComments` already used REST and is the precedent.

## The trap in the REST conversion

Two shape differences are correctness issues, not cosmetics:

- **REST `/pulls/{n}.state` is only `open`/`closed` and NEVER reports merged** — merged-ness is a separate `.merged` boolean. A naive `.state` mapping would have made *every merge assertion in the suite silently wrong*, including the one guarding #1692's singleton fast path. `restPRState` uses `if .merged then "MERGED" else (.state|ascii_upcase) end`.
- `gh` reports `OPEN`/`CLOSED`/`MERGED` uppercase; REST reports lowercase.

Every mapping was verified against live data before conversion — 5 PRs across open/closed/merged × state/baseRef/author, 5 issues × state/labels — all equivalent. The mapping table is in the helper block's doc comment.

## Verification

- `graphql_budget_remaining` tracks consumption: 4190 → 4185 across 5 known queries, where the REST probe it replaces stayed pinned at 5000.
- REST/gh equivalence check passed on live data (table above).
- `go test -race ./...` green; `bash -n scripts/e2e/run.sh` and `go vet -tags e2e ./tests/e2e/` clean.

**Not verified: a real leg.** These are budget-and-instrumentation changes; proving the numbers needs a live run with the dev daemon down (R1 competing-consumer check). I'd rather have this reviewed before spending an hour of live budget on it, particularly the REST mappings.

## Left on GraphQL deliberately

`tryPRReviewState` (needs a second REST call for reviews) and the issue-list search paths (REST search semantics differ). Both one-shot or low-frequency; ~1 point per call, and the conversion risk outweighs the saving.

## Related

Same family as #1676/#1677/#1693/#1694 — the release gate's own failure modes being quieter than the failures it exists to catch. This one is the quietest yet: the instrument that was supposed to detect budget exhaustion reported full headroom while the budget was being exhausted.

### DIFF
--- a/scripts/e2e/run.sh
+++ b/scripts/e2e/run.sh
@@ -262,11 +262,12 @@
 # present as a pass just as easily as a fail — and short-circuits any
 # remaining legs immediately (exit, not return).
 #
-# This is a log-scrape, not a live `gh api rate_limit` polling loop: it needs
-# zero new API calls against the very budget being protected, and the log
-# line already fires at the moment of interest. Each leg is also bracketed
-# with a `gh api rate_limit` call before/after (a REST metadata call, free
-# against the GraphQL budget), scoped to the bed's own FABRIK_TOKEN (read
+# This is a log-scrape, not a live budget-polling loop: it needs zero new API
+# calls against the very budget being protected, and the log line already
+# fires at the moment of interest. Each leg is also bracketed with a
+# graphql_budget_remaining call before/after (an inline `rateLimit` query
+# costing 1 point each — see that function for why the REST endpoint cannot
+# be used despite being free), scoped to the bed's own FABRIK_TOKEN (read
 # from $TEST_BED/.env into $BED_TOKEN, mirroring reset.sh's gh_() token-
 # scoping precedent — the ambient `gh` CLI's own auth is very likely a
 # different identity than the bed's @arbeithand token and would silently
@@ -1079,8 +1080,14 @@ preflight_bed_start() {
 
   # No --auto-upgrade: it would replace this freshly built binary with a
   # release mid-suite (README item 17).
-  echo "== preflight: starting bed instance (-notui, no --auto-upgrade) =="
-  ( cd "$TEST_BED" && nohup ./fabrik -notui > "$TEST_BED/bed-run.log" 2>&1 & )
+  #
+  # -poll is passed explicitly, matching StartFabrikTestBed's own launch (see
+  # defaultBedPollSeconds in tests/e2e/lifecycle.go for the measured budget
+  # rationale). Both launch sites must agree: switch_and_run restarts the bed
+  # through the Go path on every mode switch, so a cadence set only here would
+  # be silently reverted for the legs that actually matter.
+  echo "== preflight: starting bed instance (-notui -poll ${BED_POLL_SECONDS}s, no --auto-upgrade) =="
+  ( cd "$TEST_BED" && nohup ./fabrik -notui -poll "$BED_POLL_SECONDS" > "$TEST_BED/bed-run.log" 2>&1 & )
 
   # The startup banner goes to the engine's STDOUT (captured in bed-run.log),
   # while ENGINE_LOG holds the structured per-item log — which never contains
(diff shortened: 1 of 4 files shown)


arbeithand:
Addressed in `137a50c1`. The finding was correct and the framing was right: it was **missed, not deliberately left**, and the PR's "left on GraphQL deliberately" list wrongly implied otherwise.

`WaitForIssueComment` now uses `tryPRComments`. Before converting I checked your claim rather than taking it — `repos/{o}/{r}/issues/{n}/comments` against four live issues that are genuinely issues (`.pull_request == null`) and carry comments, comparing body sets by hash to `gh issue view --json comments`:

```
  ok   issue#5794   bodies identical (b85e0113)
  ok   issue#5792   bodies identical (f72f4386)
  ok   issue#5788   bodies identical (8acabc8e)
  ok   issue#5785   bodies identical (95452846)
```

Identical. Your cost estimate holds too: 10s interval × up to 25 minutes in `mergetrain_bisect_test.go` is ~150 GraphQL points per use.

Re-auditing after your comment turned up more than the one site, so correcting the record properly: **six** GraphQL reads remain, not the two the description claimed.

| site | why still GraphQL |
|---|---|
| `tryPRReviewState` (harness.go:457) | needs a second REST call for reviews |
| `WaitForChildIssueInRepo` (harness.go:780), harness.go:869 | REST search semantics differ |
| `WaitForIntegrationPR` (mergetrain_helpers.go:220), `WaitForNoStaleTrainArtifacts` (:480), harness.go:1495 | `pr list` → REST needs `head=owner:branch` query semantics |

The three `pr list` ones are in wait loops and do carry real cost, so they are worth doing — but their failure mode is *silently returning nothing*, which in a wait loop reads as a timeout rather than an error. Given this branch already contained one shape trap that would have made every merge assertion silently pass (`/pulls/{n}.state` never reporting merged), I'd rather they land as a follow-up with their own live equivalence check than be folded in here unverified.

Issue #1693, opened 2026-09-06
e2e: E2E_BED_REF is unusable for any ref but main — bed is a single-branch clone


arbeithand:
## Problem

`scripts/e2e/run.sh` documents `E2E_BED_REF` as taking any ref:

```
#   E2E_BED_REF=<ref>      ref to test (default: origin/main)
```

It cannot. The bed checkout (`~/dev/fabrik-test`) is a single-branch clone:

```
$ git -C ~/dev/fabrik-test config --get-all remote.origin.fetch
+refs/heads/main:refs/remotes/origin/main
```

`preflight_bed` runs a bare `git fetch origin`, which honours that refspec and so only ever updates `origin/main`. The subsequent `git rev-parse "$ref"` then dies for anything else:

```
== preflight: bed at /Users/bpja/dev/fabrik-test, target ref origin/fabrik/my-branch ==
fatal: ambiguous argument 'origin/fabrik/my-branch': unknown revision or path not in the working tree.
```

The failure is also unguarded — it escapes `set -e` as git's own message with no preflight context, the same class of bare-fetch problem the adjacent `git fetch` wrapper (`preflight_bed`'s existing SSH-key-failure handling) was added to fix.

## Why It Matters

Validating an engine fix against the live bed before merging it is exactly the workflow the release gate wants, and it is currently impossible without hand-fetching:

```bash
git -C ~/dev/fabrik-test fetch origin 'my-branch:refs/remotes/origin/my-branch'
```

Hit while verifying #1692 (merge-train singleton fast path landed silently). Each of the three bed rebuilds needed that manual step; forgetting it on the first attempt meant a run that built the bed from `origin/main` — i.e. tested the *unfixed* engine while `go test` ran the updated assertions. That is worse than an error: it produces a confidently wrong result. Caught before it spent budget, but only by reading the preflight's own `bed binary: dev(864a1e9)` line closely.

## Summary

Make `preflight_bed` (`scripts/e2e/run.sh`) resolve any `E2E_BED_REF` the `origin` remote actually has, regardless of the bed clone's single-branch refspec, by fetching the specific ref rather than relying on a bare `git fetch` + `rev-parse`. An unresolvable ref must fail with a wrapped, preflight-context error, not git's raw message. The default path (`origin/main`) and the existing "binary actually carries the ref under test" guard are both unchanged.

[...]

PR #1698, opened 2026-09-07
fix(e2e): detached bed leaked run.sh output pipe, wedging the gate and silently skipping legs

### LINKED ISSUE #1694: e2e: post-suite tee/jq consumer stalls after go test exits, aborting the run before its remaining legs
## Problem

`switch_and_run`'s post-suite bookkeeping can stall indefinitely draining `go test`'s output pipe. #1676's watchdog catches it (good — it exits 6 instead of hanging), but the underlying stall is a real defect, and its blast radius is larger than "one leg's report is missing": **the run aborts before its remaining legs execute.**

Observed on the v0.0.82 gate (`main` @ `819c48e2`, 2026-09-06T04:44Z):

```
############################################################
## POST-SUITE WATCHDOG (leg: on): go test exited 300s ago,
## but the script has not progressed past its post-suite steps
## within 300s (E2E_POST_SUITE_WATCHDOG).
## Stuck in: waiting for tee/jq consumer to drain
## Last engine log line: 2026-09-06T06:06:44Z [poll] nothing to do
############################################################
```

The suite itself was entirely healthy. `go test`'s own JSON stream is complete and authoritative:

- 46 top-level tests started, 46 finished — 44 pass, 2 skip (`TestCIFixReinvoke`, `TestSwitchTrainMode`), **0 fail**
- package-level result: `pass`, `Elapsed: 4619.879`
- every merge-train scenario green: `HappyPathLanding`, `TwoBasesConcurrent`, `BisectionEjectsPoisoner`, `RestartSafety`, `RedSingletonReroutesOffQueued`, `YoloAutoMergeLabel`

But the console log is truncated mid-line, the script never reached its "JSON log: …" line or its outcome classification, and — the actual cost — it never reached the **second** `switch_and_run` in the `E2E_TRAIN_MODE=on` branch:

```sh
switch_and_run on "$PARALLEL" -skip "$TRAIN_ISOLATED_RE" "$@"
switch_and_run on "$PARALLEL" -run "^(${TRAIN_ISOLATED_RE})\$"   # <-- never ran
```

So `TestMergeTrainRunawayGuardPausesBatch` silently did not execute. Nothing in the output says so; it simply isn't there. It took a set-difference against the off leg's JSON to notice — the off leg *skips* it (mode gate) and so at least names it, while the on leg omits it entirely.

## Why this is worse than a missing report

An operator reading the tail sees a watchdog banner about post-suite bookkeeping and a suite that plainly passed, and concludes the leg is fine and the runner is merely untidy on exit. It is not fine: a scenario the release gate is supposed to certify never ran, and the failure mode is *absence*, which no amount of reading the log surfaces. This is the same shape as #1622/#1644/#1687 — a guard whose omission produces silence rather than a signal.

## Requirements

- **R1** — the post-suite drain cannot block indefinitely. Bound it, and treat a stall as a reportable condition rather than a wedge.
- **R2** — a stalled or aborted leg must not silently swallow the legs that follow it. Either continue to the remaining legs after recording the failure, or fail with an explicit enumeration of which legs did not run.
- **R3** — the watchdog's message must name the legs skipped as a consequence, not only the leg that stalled. "go test exited but bookkeeping stalled" reads as cosmetic; "…and the isolated runaway-guard leg was therefore not run" does not.
- **R4** — when `go test`'s JSON stream is complete but the console capture is truncated, prefer the JSON for the outcome classification rather than reporting nothing. The data to classify this run correctly was on disk the whole time.

## Acceptance

- A leg whose post-suite drain stalls still runs the subsequent legs, or exits naming them explicitly.
- The outcome classification is produced from the JSON stream even when console capture is truncated.
- Reproduce by stalling the tee/jq consumer artificially (see `scripts/e2e/hang_hardening_test.sh` for the existing harness) and assert the isolated leg still runs or is explicitly reported as not-run.

## Context

Related: #1676 (the watchdog that caught this), #1693 (`E2E_BED_REF` unusable for non-`main` refs). Both are in the same family as this: the release gate's own failure modes being quieter than the failures it exists to catch.

arbeithand:
Fixes #1694.

## Root cause

The gate aborted at the post-suite watchdog on the `on` leg of **both** the v0.0.82 run and the post-#1697 run, after every test had already passed:

```
## POST-SUITE WATCHDOG (leg: on): go test exited 300s ago,
## Stuck in: waiting for tee/jq consumer to drain
```

The real cost was never the missing report. The abort skipped the script's **remaining legs**, so on the default gate path the isolated runaway-guard leg never ran — and nothing said so. It took a set-difference against the off leg's JSON to notice it was absent.

Confirmed by `lsof` against a bed still running hours after its run had finished:

```
fabrik 13897 bpja 3w FIFO ... /private/var/folders/.../tmp.xHMuUtjYKA/fifo
```

`run.sh` feeds `go test`'s output through a fifo — `exec 3> "$fifo"`, `go test ... >&3`, then `exec 3>&-` to drop the shell's own copy — and afterwards waits for the consumer to drain. That wait needs EOF, which needs **every** write end closed.

`go test` inherits fd 3. The test binary inherits it. And `StartFabrikTestBed`'s detached bed inherited it too: `os/exec` sets up `ProcAttr.Files` and `ExtraFiles` but does not close arbitrary inherited descriptors, and descriptors created by a parent shell are not close-on-exec. Setting `cmd.Stdin/Stdout/Stderr` to `/dev/null` covers fds 0–2 only. The bed is *designed* to outlive the run, so it held the write end open indefinitely and the consumer never saw EOF.

It only ever bit the `on` leg because that is the only leg where the bed is (re)started from **inside** a `go test` invocation rather than by `run.sh` before the pipe exists.

## Fix

**Root cause** — `markInheritedFDsCloseOnExec()` before launching the bed, so nothing this suite spawns crosses an exec boundary holding the runner's descriptors. Marking a descriptor close-on-exec doesn't affect this process's own use of it; the flag is consulted only at `exec(2)`.

**Defence in depth** — `drain_output_consumer` bounds the wait (`E2E_POST_SUITE_DRAIN_TIMEOUT`, default 30s, well under the 300s watchdog), says what to look for, and lets the run **continue to its remaining legs** rather than wedging. `go test` has already exited by then, so its JSON is complete on disk and the leg's results are unaffected — asserted in the test, not assumed.

## On the guards — worth reading

The first version of this guard was two greps: one for `POST_SUITE_DRAIN_TIMEOUT` in `run.sh`, one for `markInheritedFDsCloseOnExec` in `lifecycle.go`. **Both passed with the fix removed** — the strings still matched a surrounding comment and the function's own `func` definition respectively.

That is precisely the vacuous-guard shape #1687 was filed about, in the fix meant to stop silent skipping. It was caught only because the mutations were actually run rather than reasoned about.

So the drain is now **extracted as a function specifically to be testable**, and `hang_hardening_test.sh` exercises the real implementation against a real wedged consumer — a fifo whose write end is held by a descendant that outlives its parent, i.e. the actual mechanism.

Neutralization, each turning exactly one case red, and legibly rather than by hanging (the guard calls the drain under `with_timeout` and cleans up unconditionally, so a wedged implementation cannot stall the suite):

| mutation | result |
|---|---|
| bound removed (bare `wait`) | `FAIL: drain_output_consumer did not return within 15s …` |
| never abandons, always returns 0 | `FAIL: … reported a clean drain against a deliberately wedged consumer` |
| restored | all checks passed, rc=0 |

The `markInheritedFDsCloseOnExec` half was verified separately, both directions, against a verbatim copy of the helper: without it a detached child inherits the descriptor; with it it does not.

## Verification

- `go test -race ./...` green
- `scripts/e2e/hang_hardening_test.sh` — all checks passed (2 new cases)
- `go vet -tags e2e ./tests/e2e/`, `bash -n scripts/e2e/run.sh` clean

**Not verified: a live gate run.** That is the actual proof — the `on` leg reaching its second invocation and the isolated runaway-guard leg running. Worth doing after review rather than before, since a full gate is ~4 hours.

## Context

Last of the three release-gate defects found during the v0.0.82 cut (#1693 `E2E_BED_REF`, #1694 this, plus #1697's dead budget gauge). All the same family: the gate's own failure modes being quieter than the failures it exists to catch.

### DIFF
--- a/scripts/e2e/hang_hardening_test.sh
+++ b/scripts/e2e/hang_hardening_test.sh
@@ -413,9 +413,15 @@ mkfifo "$drain_fifo"
 { tee "$drain_json" >/dev/null; } < "$drain_fifo" &
 drain_consumer=$!
 exec 4> "$drain_fifo"
+drain_holder_pidfile="$drain_dir/holder.pid"
 (
   echo '{"Action":"output","Output":"hello"}' >&4
   ( sleep 60 ) >&4 2>/dev/null &     # the "bed": outlives go test holding fd 4
+  # Record it: this is a GRANDchild, reparented the moment its parent subshell
+  # exits, so neither this case's own cleanup nor the trailing `pkill -P $$`
+  # can reach it by process tree. Without the pidfile it lingers as an orphan
+  # for the full 60s after the script finishes (found in review).
+  echo $! > "$drain_holder_pidfile"
 ) &
 drain_suite=$!
 exec 4>&-
@@ -454,6 +460,9 @@ fi
 # #1694 hung the gate — turning a legible FAIL into an unexplained stall.
 kill -TERM "$drain_consumer" 2>/dev/null || true
 wait "$drain_consumer" 2>/dev/null || true
+# Read the pidfile BEFORE rm -rf takes the directory with it.
+drain_holder="$(cat "$drain_holder_pidfile" 2>/dev/null || echo "")"
+[ -n "$drain_holder" ] && kill -TERM "$drain_holder" 2>/dev/null
 pkill -P $$ >/dev/null 2>&1 || true
 rm -rf "$drain_dir"
 
--- a/scripts/e2e/hang_hardening_test.sh
+++ b/scripts/e2e/hang_hardening_test.sh
@@ -394,6 +394,92 @@ else
 fi
 rm -f "$probe_err"
 
+# Case 16 (#1694): the post-suite output drain must be BOUNDED, and this guard
+# must exercise run.sh's REAL drain_output_consumer — not a local copy of the
+# pattern, and not a grep for a string that also appears in a comment. The
+# first attempt at this guard did exactly that and passed with the bound
+# removed; see drain_output_consumer's own comment.
+#
+# Mechanism being modelled: the consumer only exits on EOF, which needs every
+# write end of the fifo closed. A descendant of `go test` that outlives it
+# holding fd 3 keeps it open forever — which is what the detached bed did. A
+# bare `wait` then blocks until the post-suite watchdog kills the whole script,
+# silently skipping the remaining legs (on the default gate path, the isolated
+# runaway-guard leg never ran and nothing said so).
+drain_dir="$(mktemp -d)"
+drain_fifo="$drain_dir/fifo"
+drain_json="$drain_dir/out.json"
+mkfifo "$drain_fifo"
+{ tee "$drain_json" >/dev/null; } < "$drain_fifo" &
+drain_consumer=$!
+exec 4> "$drain_fifo"
+(
+  echo '{"Action":"output","Output":"hello"}' >&4
+  ( sleep 60 ) >&4 2>/dev/null &     # the "bed": outlives go test holding fd 4
+) &
+drain_suite=$!
+exec 4>&-
+wait "$drain_suite" 2>/dev/null || true
+
+# Sanity first: the wedge must actually exist, or this case proves nothing.
+sleep 1
+if ! kill -0 "$drain_consumer" 2>/dev/null; then
+  fail "consumer exited on its own despite a descendant holding the fifo open — this case no longer reproduces #1694's mechanism and would not catch its return"
+else
+  drain_started=$SECONDS
+  # Called under with_timeout so that an UNBOUNDED implementation (the #1694
+  # bug: a bare `wait`) fails this case legibly instead of hanging the whole
+  # test script — verified by neutralization: without the bound this branch
+  # reports rc=124 rather than the suite wedging.
+  with_timeout 15 drain_output_consumer "$drain_consumer" 3 "test" 2>/dev/null
+  drain_rc=$?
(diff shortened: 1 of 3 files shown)


arbeithand:
Addressed in `99a03e5c`. You were right that neither the case's own cleanup nor the trailing `pkill -P $$` could reach it — it's a grandchild, reparented the moment its parent subshell exits, so there's no process-tree path to it. It now records its pid to a file and is terminated during cleanup (reading the pidfile before `rm -rf` takes the directory with it).

Verified: no `sleep 60` remains after the suite exits, and all checks still pass (rc=0).

Also, thank you for the checks I'd asserted but not actually verified myself — no `ExtraFiles` usage the sweep would break, the `RLIMIT_NOFILE` unlimited case being safe via the `cur > 0` guard, Go's own descriptors already being close-on-exec so the loop only touches genuinely inherited fds, and `set -m` being active even under sourcing so the process-group kill reaches the real `tee`/`jq` pipeline. That last one in particular was load-bearing for the fix and I had only reasoned about it.

Issue #1082, opened 2026-07-26
Missing board Status option silently strands merged issues — dependents stay blocked forever


bdueck:
## Summary

When a PR merges, the terminal path advances the issue through a **`Queued`** Status option on the
way to `Done`. If the project board has no such option, the transition fails, and the failure is
reported **only as a log warning**:

```
[#NNN pr-terminal] PR #NNN merged — filling gate-checked completion labels and advancing to Done
[#NNN warn] pr-terminal: could not advance to Done: no status option "Queued" found on project
            board (available: [Todo In Progress Implement Backlog Specify Research Plan Review
            Validate Done])
```

The issue stays **open**. Because auto-sequencing reads GitHub's native dependency edges, and those
clear on **close** rather than on merge, every downstream issue stays `fabrik:blocked` — waiting on
work that has already shipped.

## Why this is worse than a normal misconfiguration

The whole wave stalls with **no failure signal anywhere**:

- The PR is merged and green.
- The issue shows every stage complete.
- The blocked dependents show a perfectly reasonable `blocked` label and a comment naming a blocker
  that a human can see is done.
- Nothing is red. Nothing is paused. Nothing asks for attention.

Observed: three merged, deployed, verified issues sat open while five downstream slices waited on
them. The stall was found only by reading `fabrik.log` directly — and it had been silently in effect
for some time before that. A stall that presents as "quietly working" is the most expensive kind,
because the natural operator response is to wait longer.

## Why the existing startup check doesn't catch it

Startup already validates the board in the *other* direction:

```
[startup] warning: board has columns with no matching stage: In Progress, Todo
```

That reports **extra** columns. It does not report **missing required** ones — so the board looks
validated while lacking the single option the terminal transition depends on.

## Proposed mechanism

1. **Validate required Status options at startup**, alongside the existing extra-columns warning.
   If a Status option the pipeline depends on is absent, say so loudly at boot — that is the moment
   an operator can fix it, before any work is queued.
2. **Escalate the runtime failure past the log.** A terminal transition that cannot complete should
   surface where the operator is already looking: an issue comment, or the same pause/label
   mechanism used for other human-attention conditions. A merged-but-unadvanced issue is precisely a
   "needs a human" state.
3. **Consider a fallback**: if `Queued` is absent but `Done` exists, advancing straight to `Done`
   preserves the invariant that matters downstream (the issue closes, edges clear). If the
[...]

Issue #770, opened 2026-05-23
Convert Fabrik to use a GitHub App (bot) identity


arbeithand:
## Problem

Fabrik currently authenticates all GitHub API calls and git operations using a Personal Access Token (PAT) tied to a specific human user account (configured via `FABRIK_TOKEN`/`GITHUB_TOKEN`). When multiple Fabrik instances run under different user identities, each instance can only edit GitHub comments it created — because the GitHub API enforces that only the comment author (or an admin) can update a comment.

This is a direct blocker for issue #84 (allowing any Fabrik instance to update Stage Comments from prior stages). In a multi-user scenario, Fabrik userA posts a "Research" stage comment. Fabrik userB's token cannot update that comment, because the GitHub API returns a 403 — the comment was authored by userA's account.

The fix is to have all Fabrik instances share a single GitHub App bot identity for API and git operations. Comments written by `fabrik[bot]` can be updated by any instance that holds a valid installation access token for the same app, regardless of which human user configured that instance.

## Summary

Replace the human-PAT-based GitHub identity with a GitHub App installation for all API calls and git operations. All Fabrik instances in the same GitHub organization will authenticate as the shared app bot (`fabrik[bot]` or equivalent). The human user identity (`FABRIK_USER`) is preserved for locking semantics only — it determines which `fabrik:locked:<user>` label governs concurrency, but is no longer the credential identity used for GitHub operations.

PAT mode must remain supported as a fallback for backward compatibility with single-user deployments that do not configure a GitHub App.

## Requirements

- **GitHub App auth mode**: When `GITHUB_APP_CLIENT_ID` and `GITHUB_APP_PRIVATE_KEY` are present, Fabrik authenticates as a GitHub App installation instead of using a PAT.
- **PAT fallback**: When App credentials are absent, Fabrik falls back to the existing `FABRIK_TOKEN`/`GITHUB_TOKEN` PAT behavior unchanged. No regression for existing single-user deployments.
- **JWT generation**: Fabrik generates a short-lived JWT (signed with the app's RSA private key) to bootstrap installation access token requests.
- **Installation access token exchange**: Fabrik calls `POST /app/installations/{installation_id}/access_tokens` to obtain a `ghs_...` token with a 1-hour expiry.
- **Token refresh**: Fabrik proactively refreshes the installation access token before expiry (e.g., ~5 minutes before the `expires_at` timestamp). All in-flight API calls continue to use the current valid token.
[...]

PR #1717, opened 2026-09-13
fix(githubauth): omit hook_attributes entirely from App manifest

arbeithand:
Closes #1711

## What this does

`buildManifest` (`internal/githubauth/manifest.go`) was emitting a `hook_attributes` object with only `active: false` and no `url` sub-key. GitHub requires `hook_attributes.url` whenever `hook_attributes` is present at all, and rejects the entire manifest when it's missing — with a rejection message (`"url" wasn't supplied`) that misleadingly points at the manifest's top-level (present, correct) `url` field instead.

Pruefer V1 is polling-only (ADR-1113 §1, ADR-032), so there's no webhook URL to supply. The fix is to omit `hook_attributes` entirely, which is how GitHub's manifest schema expresses "no webhook."

## Key changes

- **`internal/githubauth/manifest.go`** — deleted the `hook_attributes` entry from `buildManifest`'s returned map; rewrote the doc comment to explain the omission and why (referencing this issue's misleading-error finding).
- **`internal/githubauth/manifest_test.go`** — inverted `TestBuildManifest_NoActiveWebhook` (which previously asserted the broken shape — `hook_attributes` present with `active: false`) into `TestBuildManifest_NoHookAttributes`, which asserts the key is **absent**. Added `manifestContractViolations`, a small helper encoding GitHub's documented manifest contract (top-level `url` always required; `hook_attributes`, if present, requires a non-empty `url`), and two new guard tests:
  - `TestManifestContract_CatchesHookAttributesWithoutURL` — proves the guard fires if this defect is ever reintroduced.
  - `TestManifestContract_AllowsHookAttributesWithURL` — proves the guard doesn't over-reject a legitimate future webhook addition.
- **`adrs/1253-github-app-manifest-auth-reconciler.md`** — corrected the "Positive" consequence claiming the manifest flow makes Pruefer "genuinely self-hostable." It was never exercised against real GitHub prior to this fix (every test in the package is `httptest`-mocked, which is why the defect went undetected), so `handarbeit-pruefer` was almost certainly created by hand — corroborated by its live `issues: read` permission differing from `buildManifest`'s requested `issues: write`.

## How to test

```bash
go test -race ./internal/githubauth/...
```

All manifest-related tests pass, including the inverted assertion and the two new contract-guard tests. `go build ./...` and `go vet ./...` are clean.

[...]

PR #1700, opened 2026-09-10
fix(e2e): preflight_bed resolves any origin ref, not just main

arbeithand:
Closes #1693

## What

`preflight_bed` (`scripts/e2e/run.sh`) used a bare `git fetch origin` to prepare `E2E_BED_REF` for resolution. Because the bed checkout (`~/dev/fabrik-test`) is a single-branch clone (`+refs/heads/main:refs/remotes/origin/main`), that bare fetch only ever updated `origin/main` — any other `E2E_BED_REF` died with git's raw, unwrapped `fatal: ambiguous argument` on the subsequent `rev-parse`, outside any preflight error handling.

## Key changes

- **`scripts/e2e/run.sh`**: the fetch step now supplies an explicit destination refspec (`"$ref_name:refs/remotes/origin/$ref_name"`), which writes to that ref regardless of the remote's configured single-branch refspec — the same mechanism the issue's own manual workaround demonstrated. This is a strict, harmless generalization of the old bare fetch for the default `origin/main` path, so the default path is unaffected (R4). The wrapped fetch-failure message is generalized to name both possible causes (SSH key not loaded, or the ref not existing on `origin`) along with `$ref` and `$TEST_BED`, since there's now only one fetch call to wrap (R1/R2). The downstream `rev-parse`, checkout, build, and the `./fabrik --version` / `$want_short` guard (R3) are untouched.
- **`scripts/e2e/preflight_bed_ref_test.sh`** (new): regression coverage modeled on `pregate_test.sh`'s sourced-function pattern. Builds two scratch git repos (a bare `origin` and a single-branch `bed` clone matching the real bed's refspec shape — asserted directly) and exercises three scenarios: a never-fetched branch resolves; a bogus ref fails with the wrapped preflight-context message (not git's raw error) and exits `PREFLIGHT_FAILED_EXIT`; the unset-default path is unaffected. Verified this test fails against the pre-fix code and passes against the fix.
- **`.github/workflows/ci.yml`**: wires the new test in alongside `pregate_test.sh` and friends.

## How to test

```bash
bash scripts/e2e/preflight_bed_ref_test.sh
```

Also manually verified against the real bed (`~/dev/fabrik-test`, verify-only mode): `E2E_BED_REF=origin/fabrik/issue-1693` (a branch never previously fetched into the bed) resolved correctly with no manual fetch, and a bogus ref failed with the wrapped message — bed left untouched in both cases.

PR #1718, opened 2026-09-13
feat(init): create, validate and repair the project board from stage configs

### LINKED ISSUE #1714: feat(init): create, validate and repair the project board from stage configs
## Problem

Fabrik's most common onboarding failure is a project board whose column names do not
exactly match the stage names in `.fabrik/stages/*.yaml`. It is the first entry under
*Common Issues* in `CLAUDE.md`:

> **Startup board validation failure**: Stage names in `.fabrik/stages/*.yaml` must match
> the column names on your GitHub Project board exactly. Check both for typos.

The failure class exists because the board is hand-built: a human reads the stage configs
and retypes nine names into GitHub's UI. Fabrik already knows the correct names — it wrote
them.

The Projects v2 API can create and fully configure a board. Verified 2026-09-13:
`createProjectV2`, `updateProjectV2Field` (setting the exact stage column set in order),
`linkProjectV2ToRepository` and `updateProjectV2` all succeed. **This works with an
ordinary PAT as well as an App token**, so it is not blocked on #770 and benefits every
user immediately.

## Summary

Teach `fabrik init` to create a fully-configured project board directly from the stage
configs it just wrote — Status options in stage order, linked to the repo, described —
so a fresh onboarding never requires hand-retyping column names. Separately, give an
existing board an explicit, opt-in **repair** path that adds missing stage columns
without disturbing any item's current Status, and sharpen board validation to name
exactly which columns are missing or extra instead of only reporting failure.

## Requirements

- **R1** — `fabrik init` can create a project board whose Status options are derived from
  the stage configs it just wrote, in stage order, so board and config cannot disagree.
  Link it to the repo and set a description. The exact CLI invocation (a new flag on
  `fabrik init` vs. a separate mode) is Research/Plan's call, guided by `cmd/init.go`'s
  existing flag conventions; the board owner defaults to the same `owner` value used
  elsewhere in `.fabrik/config.yaml` unless Research finds a reason it must differ.
- **R2** — Board validation reports **which** columns are missing and which are extra, by
  name, rather than only that validation failed. Extra columns remain a warning, not an
  error, as today. `engine/startup.go`'s `checkStageColumnAlignment` already computes and
  reports both a missing-stage list and an extra-column list by name — Research must
  confirm whether the existing output already satisfies AC2 as-is, or identify the
  specific gap, rather than assuming this is net-new work.
- **R3** — Board **repair** can add missing stage columns to an existing board, opt-in,
  never automatic. Invocation shape is Research/Plan's call (see R1).
- **R4** — **Repair must preserve existing option ids.** `updateProjectV2Field` replaces
  the *entire* option set, and `ProjectV2SingleSelectFieldOptionInput`'s `id` field is
  optional:

  - omit `id` → every option is regenerated with a fresh id, and **every item on the
    board silently loses its Status** (`fieldValueByName` returns `null`)
  - pass each existing option's `id` → ids are preserved and item assignments survive;
    new options are added by omitting `id` on just those entries

  Measured 2026-09-13: adding one column to a 10-column board regenerated all ten ids and
  nulled an item that had been set to `Implement`.

  Since Status **is** Fabrik's pipeline state, the omit-id form on a live board is a
  catastrophic data-loss event — every item falls out of every column at once. Repair must
  read the current options and echo their ids back.
- **R5** — Ship a test that places an item in a column, adds a new column, and asserts the
  item's Status survives. The failure mode is silent, so nothing else will catch a
  regression here.
- **R6** — Repair must never delete a column that has items in it. Removing a column is
  out of scope entirely; extra columns are reported, not removed.
- **R7** — Creation and repair are **organization-only**. Fabrik already resolves
  `OwnerType` (`"organization"` or `"user"`) when fetching an existing board; creation and
  repair must check it and refuse explicitly (naming the reason) for a user-owned target
  rather than attempting the mutation and surfacing whatever error the GitHub API happens
  to return. Creating a brand-new board has no existing board to resolve `OwnerType` from,
  so Research must establish where that value comes from in the creation path (config,
  flag, or an owner-type lookup against the org/user login itself) before repair's
  existing-board path and creation's new-board path can share the same refusal check. For
  a user-owned board under App auth the operations are unavailable (see #770); under a PAT
  they may work, but that distinction must be explicit in Fabrik's refusal message rather
  than discovered at API-failure time.

## Scope

**In scope**: `cmd/init.go`, board validation in the startup path, new Projects v2
mutations in `github/`, and their tests.

**Out of scope**:
- App authentication itself (#770 chain)
- `fabrik init --github-app` setup UX (separate issue)
- Creating or modifying any field other than Status
- Deleting columns, or repairing a user-owned board (R7)

## Acceptance Criteria

- **AC1** — `fabrik init` on an org account creates a board whose columns match the stage
  configs exactly, with no manual board editing, and the engine starts against it without
  a validation error.
- **AC2** — Validation output names each missing and each extra column.
- **AC3** — Adding a stage column to a board that already has items preserves every
  item's Status, proven by an automated test.
- **AC4** — Repair is never performed without explicit opt-in.
- **AC5** — `go test -race ./...` passes.

## Prior Art / Context

- GitHub's Projects v2 GraphQL mutations (`createProjectV2`, `updateProjectV2Field`,
  `linkProjectV2ToRepository`, `updateProjectV2`) were verified against the live API on
  2026-09-13, including the option-id data-loss behavior in R4 — this is empirical, not
  speculative.
- `github/project.go` already contains the read-side GraphQL query this issue's mutations
  will complement; `github/wire_contract_test.go` + `github/testdata/` is the existing
  pattern for schema-validated, fixture-recorded tests that new mutations should follow.
- `engine/startup.go`'s `checkStageColumnAlignment` is the existing validation entry
  point; R2's work is scoped to that function and its report-building helpers, not a new
  validation path — see the note under R2.

## Risks / Dependencies

- The R4 option-id-preservation behavior is the single highest-risk part of this issue:
  getting it wrong silently detaches every item on a live board from its Status. R5's test
  is the primary safeguard and must not be skipped or weakened.
- Part of #770's onboarding requirements (O1–O5), but **deliberately not blocked on App
  auth** — every mutation here works with a PAT, so the onboarding win ships independently
  and to all users. Only O5's App-specific user-owned-board refusal lives in the App-auth
  issue.

arbeithand:
Closes #1714

## Summary

Teach `fabrik init` to create a fully-configured project board directly from the stage
configs it just wrote — Status options in stage order, linked to the repo, described —
so a fresh onboarding never requires hand-retyping column names. Separately, give an
existing board an explicit, opt-in **repair** path that adds missing stage columns
without disturbing any item's current Status, and sharpen board validation to name
exactly which columns are missing or extra instead of only reporting failure.

## Problem

Fabrik's most common onboarding failure is a project board whose column names do not
exactly match the stage names in `.fabrik/stages/*.yaml`. It is the first entry under
*Common Issues* in `CLAUDE.md`:

> **Startup board validation failure**: Stage names in `.fabrik/stages/*.yaml` must match
> the column names on your GitHub Project board exactly. Check both for typos.

The failure class exists because the board is hand-built: a human reads the stage configs
and retypes nine names into GitHub's UI. Fabrik already knows the correct names — it wrote
them.

The Projects v2 API can create and fully configure a board. Verified 2026-09-13:
`createProjectV2`, `updateProjectV2Field` (setting the exact stage column set in order),
`linkProjectV2ToRepository` and `updateProjectV2` all succeed. **This works with an
ordinary PAT as well as an App token**, so it is not blocked on #770 and benefits every
user immediately.

### Approach

This issue adds the first "board-structure mutation" capability to Fabrik. The design keeps the new GraphQL/REST plumbing in `github/`, keeps the CLI-facing decisions (flags, prompts, refusal messages) in `cmd/`, and reuses as much existing machinery as possible (`writeConfigTemplate`, `stages.LoadAll`, the `refresh-stages`-style dry-run/`--apply` UX, the wire-contract test discipline).

Two independent entry points, per R1/R3:

- **`fabrik init --create-board --owner <org> --repo <repo> [--title <title>]`** — creates a brand-new, fully-configured board and folds it into the same config-writing path `init` already uses for a pasted project URL.
- **`fabrik repair-board [--apply]`** — a new, separate subcommand (not a flag on the daemon or on `init`) that diagnoses and (opt-in) repairs an *existing*, already-configured board. Modeled directly on `refresh-stages`'s "print a plan, require `--apply` to act" UX, since repair targets a live board with real items on it — the highest-consequence surface in this issue (R4).

Both paths share a small new helper file (`cmd/board_admin.go`) for token/client construction, the R7 org-only refusal check, and the stage→column-name diffing logic, so the two entry points can never disagree about what "missing" or "user-owned" means.

All new GraphQL operations live in a new `github/project_admin.go` (mirroring the Research note that this is a new category — administration — distinct from `project.go`'s fetch concerns and `status.go`'s per-item status concerns), with `github/status.go`'s `FetchStatusField`/`StatusField` extended additively (new `OptionDetails` field; `Options` untouched) to carry the `color`/`description` every existing option's echo requires.

**Key simplification vs. Research's four-mutation list**: `createProjectV2`'s input accepts an optional `repositoryId` directly, so the initial link happens atomically in the same call that creates the project — no separate `linkProjectV2ToRepository` call, and no window where the project exists but is unlinked. `linkProjectV2ToRepository` is not implemented at all in this issue: nothing would call it, and an untested, uncalled mutation is dead code the wire-contract completeness check would just as soon not carry. R1's "link it to the repo" requirement is satisfied by the `repositoryId` field on `createProjectV2Input`, confirmed present in the vendored schema.

### New/Modified Files

| File | Change |
|------|--------|
| `github/project_admin.go` (new) | `ResolveOwner(login) (id, ownerType string, err error)` via `repositoryOwner(login:)`; `FetchRepositoryID(owner, repo) (string, error)` via a `repository(owner:, name:) { id }` query (mirrors `prNodeIDQuery`); `CreateProjectV2(ownerID, title, repositoryID string) (projectID string, number int, err error)`; `SetProjectDescription(projectID, shortDescription string) error`; `StatusOptionInput{ID *string, Name, Color, Description string}` and `SetStatusFieldOptions(fieldID string, options []StatusOptionInput) error` (the `updateProjectV2Field` mutation) |
| `github/status.go` | Extend `fetchStatusFieldQuery` to also select `color`/`description`; add `StatusOption{ID, Name, Color, Description string}` and `StatusField.OptionDetails map[string]StatusOption` (additive — `Options`/`OrderedOptionNames` unchanged, so no existing caller needs to change) |
| `github/project_admin_test.go` (new) | httptest-based unit tests for every new function, plus the R4/R5 id-preservation payload test (see Key Decisions) |
| `github/status_test.go` | Extend existing `FetchStatusField` tests to cover the new `color`/`description` fields |
| `github/wire_contract_test.go` | Register the 4 new query/mutation consts (`repositoryOwnerQuery`, `repositoryIDForBoardQuery`, `createProjectV2Mutation`, `updateProjectV2ShortDescriptionMutation`, `updateProjectV2FieldMutation` — 5 total; name to avoid clashing with existing `prNodeIDQuery`/`fetchStatusFieldQuery` idents) |
| `cmd/board_admin.go` (new) | Shared helpers: `loadGitHubToken() (string, error)`, `newBoardGHClient(ghesHost string) (*gh.Client, error)`, `refuseIfUserOwnedBoard(ownerType string) error` (R7), `requiredStageColumnNames(allStages []*stages.Stage) []string` (ordered, excludes `CleanupWorktree`/`Unmanaged`, **includes** holding stages unconditionally), `missingStageColumns(required []string, sf *gh.StatusField) []string` |
| `cmd/board_admin_test.go` (new) | Unit tests for the above helpers, especially the R7 refusal and the missing-column diff |
| `cmd/init.go` | Add `--create-board`, `--owner`, `--repo`, `--title` flags; new `runCreateBoard(...)` path wired into `runInit` after stage extraction, before `writeConfigTemplate`; extend `writeConfigTemplate`'s signature to accept `repo` in the URL-flow branch too (mechanical, no behavior change for existing callers since that branch already passes `""`) |
| `cmd/init_test.go` | Tests for the new flags: mutual exclusivity with a positional project URL, `--repo` required with `--create-board`, and the create-board flow's config-write result (mocked `gh.Client` via `httptest`) |
| `cmd/repair_board.go` (new) | `runRepairBoard(args []string) error` + testable core `repairBoardCore(client *gh.Client, cfg config.ProjectConfig, allStages []*stages.Stage, apply bool, w io.Writer) error`, following `refreshStagesWithReader`'s shape |
| `cmd/repair_board_test.go` (new) | Dry-run vs `--apply` behavior, R6 (never a delete path — by construction, no test needed to prove a negative beyond "existing options are always a strict prefix of the sent list"), and the R4/R5 id-preservation assertion via `repairBoardCore` |
| `cmd/root.go` | Dispatch `repair-board` to `runRepairBoard`; add it to the usage banner |
| `engine/startup_test.go` | Verify (not necessarily add) that `checkStageColumnAlignment`'s fatal-report and extra-column-warning paths already name every missing/extra column by name (R2/AC2) |
| `docs/USER_GUIDE.md` | Document `fabrik init --create-board ...` as the recommended org-account path (manual setup instructions kept as the user-owned-board fallback); document `fabrik repair-board [--apply]`; update "Startup Board Validation" to point at `repair-board` as the fix-it path |
| `docs/llms-full.txt` | Regenerate via `scripts/generate-llms-full.sh` (required because `USER_GUIDE.md` changes) |
| `adrs/1714-board-admin-create-and-repair.md` (new) | ADR — see Key Decisions |
| `CLAUDE.md` | Add `github/project_admin.go`, `cmd/repair_board.go`, `cmd/board_admin.go` to the Architecture file list; update the `cmd/init.go` bullet |

### Key Decisions

- **`repositoryId` inline at `createProjectV2`, no separate `linkProjectV2ToRepository` call.** Fewer round-trips, no partial-linkage window, no dead/uncalled mutation for the wire-contract registry to carry. The schema's `repositoryId: ID` field on `CreateProjectV2Input` exists for exactly this purpose.
- **`fabrik init --create-board` requires `--owner` and `--repo` explicitly**, and is mutually exclusive with the existing positional project-URL argument (that argument is for linking to a project that already exists; `--create-board` is for when none does). `--title` defaults to `"<repo> Fabrik Pipeline"`.
- **Repair is a separate subcommand, not an `init` flag.** Repair targets a board that already has live items and is typically run long after onboarding, on a totally different invocation cadence than `init`. It follows `refresh-stages`'s existing "print plan, require `--apply`" convention rather than inventing a new UX — same convention, same audience (an operator diagnosing a stale board), satisfies AC4 (never automatic) by construction, since there is no code path in `repairBoardCore` that mutates without `apply == true`.
- **`OwnerType` resolution for creation vs. repair uses different sources but the same refusal check.** Repair reads `board.OwnerType` from the already-resolved `FetchProjectBoard`. Creation has no board yet, so it resolves ownership fresh via `repositoryOwner(login:)`'s `__typename` — never from `.fabrik/config.yaml`'s `owner_type` key, which is advisory/legacy for the read path and must not be trusted for a mutating decision (a stale or hand-edited value there would let R7's gate silently pass). Both paths call the same `refuseIfUserOwnedBoard(ownerType string) error` so the refusal wording and behavior can never diverge between the two entry points.
- **`FetchStatusField` gains `OptionDetails` additively rather than reshaping `Options`.** Every existing caller (`checkStageColumnAlignment`, `UpdateProjectItemStatus`'s id lookups) only ever needed the id-by-name map; reshaping it would touch call sites that have nothing to do with this issue. New code (repair, creation) is the only consumer of `OptionDetails`.
- **Repair always includes holding stages in "missing" (no ADR-1421 deferral nuance).** The deferral in `checkStageColumnAlignment` exists to keep *startup* non-fatal when `merge_train` is off on an unrelated board someone hasn't gotten around to fixing. Repair is an explicit, opt-in operator action — there's no reason to withhold a column the operator explicitly asked to add just because `merge_train` happens to be off today. Creation includes holding stages unconditionally for the same reason Research flagged: `fabrik init` extracts `queued.yaml` unconditionally, so a fresh board should carry that column from minute one (avoids reproducing the exact #1082/ADR-1421 gap on the very first `merge_train: on` flip).
- **New options get `color: GRAY, description: ""`.** The issue doesn't specify option styling; GRAY is a schema-valid neutral default and matches what a hand-created column looks like before a human customizes it. Not user-configurable in this issue — no requirement asks for it.
- **R4's automated test proves the outbound payload, not GitHub's server-side id-preservation behavior.** There is no live sandbox wired into this test layer (per Research's Risks section and `CLAUDE.md`'s test-layer table — Projects v2 admin mutations have no wire-contract fixture precedent yet and R7 scopes this org-only, which the CI environment can't credibly assert against live GitHub). The test that satisfies AC3/R5 asserts, via `httptest`, that `repairBoardCore`'s constructed `singleSelectOptions` payload echoes every existing option's exact `id`/`color`/`description` unchanged and appends only the new stage names with `id` omitted — i.e., it proves Fabrik sends the payload shape the issue's own empirical finding (2026-09-13) established as safe. This is stated explicitly in the ADR so a future reader doesn't wonder why the test doesn't hit a real board.
- **ADR-worthy**: yes. This introduces (a) the first board-structure-mutating code Fabrik has ever shipped, (b) a non-obvious, safety-critical id-preservation contract for anyone touching `updateProjectV2Field` in the future, and (c) an org-only restriction (R7) that isn't discoverable from the code without reading the refusal message. `adrs/1714-board-admin-create-and-repair.md`.

### Task Checklist

- [x] Extend `github/status.go`: add `StatusOption`/`OptionDetails` to `StatusField`, extend `fetchStatusFieldQuery` to fetch `color`/`description`; update/add tests in `github/status_test.go`
- [x] Create `github/project_admin.go`: `ResolveOwner`, `FetchRepositoryID`, `CreateProjectV2` (with inline `repositoryId`, returning `projectID` + `number`), `SetProjectDescription`, `StatusOptionInput`, `SetStatusFieldOptions`
- [x] Add `github/project_admin_test.go`: httptest coverage for each new function (success + error paths), including the id/color/description echo-payload assertion for `SetStatusFieldOptions`
- [x] Register all new consts in `github/wire_contract_test.go`'s `wireContractRegistry`; run the suite to confirm schema validation passes with no schema refresh needed (already confirmed by Research against the vendored SDL)
- [x] Create `cmd/board_admin.go`: `loadGitHubToken`, `newBoardGHClient`, `refuseIfUserOwnedBoard`, `requiredStageColumnNames`, `missingStageColumns`; `cmd/board_admin_test.go` for each
- [x] Extend `cmd/init.go`: add `--create-board`/`--owner`/`--repo`/`--title` flags, mutual-exclusivity check against the positional URL, `--repo` required with `--create-board`; implement the create-board flow (resolve owner → refuse if user-owned → resolve repo id → load just-written stage configs → create project → set description → set Status options in stage order → write config via the extended `writeConfigTemplate`)
- [x] Extend `writeConfigTemplate`'s signature to accept `repo` in the URL-provided branch (mechanical; existing call site passes `""`, unchanged behavior)
- [x] Add `cmd/init_test.go` cases for the new create-board flow and its error paths (no token, user-owned refusal, missing `--repo`, URL+`--create-board` conflict)
- [x] Create `cmd/repair_board.go`: `runRepairBoard` + `repairBoardCore` (load config → load stages → fetch board/status field → refuse if user-owned → diff missing columns → print plan → if `--apply`, echo existing options + append new, call `SetStatusFieldOptions`)
- [x] Add `cmd/repair_board_test.go`: dry-run prints plan and makes no mutation call; `--apply` sends exactly one mutation with the id-preserving payload (R4/R5/AC3); no-missing-columns case is a no-op even under `--apply`
- [x] Wire `repair-board` into `cmd/root.go`'s subcommand dispatch and usage banner
- [x] Verify AC2/R2: read or extend `engine/startup_test.go` to confirm `checkStageColumnAlignment`'s fatal report and extra-column warning already name every missing/extra column; add a test only if a genuine gap is found (expected: none — treat this as verification, not new work)
- [x] Update `docs/USER_GUIDE.md`: add `fabrik init --create-board` as the recommended org-account path (keep manual instructions as the user-owned-board fallback); document `fabrik repair-board [--apply]`; update "Startup Board Validation" to reference `repair-board`
- [x] Regenerate `docs/llms-full.txt` via `scripts/generate-llms-full.sh` in the same commit as the USER_GUIDE changes
- [x] Update `CLAUDE.md`'s Architecture section for the new/changed files
- [x] Create ADR 1714: Board Administration (Create & Repair) via Projects v2 — cover the id-preservation contract, the org-only restriction, and why the R4/R5 test asserts payload shape rather than live server behavior
- [x] Run `go build ./...`, `go vet ./...`, and `go test -race ./...` (AC5)

### Risks

- **R4 is still the dominant risk.** Getting the option-echo wrong (omitting `color`/`description` on an existing option) fails loud with a GraphQL validation error — safe but confusing; getting it wrong the other way (omitting `id`) is silent data loss on a live board. The dedicated payload-assertion test is the only guard; it must check all three of `id`, `color`, `description` on every echoed option, not just `id`.
- **`repositoryOwner(login:)`'s exact response shape is unverified against the live API** (Research flagged this — only the other four mutations were live-verified on 2026-09-13). Implement should sanity-check the query against a real org/user login early, since it's load-bearing for both `--create-board`'s `ownerId` and R7's gate on that path.
- **No live-recorded fixture exists for any of these mutations.** Coverage is `httptest`-only plus wire-contract schema validation, consistent with Research's recommendation given R7's org-only scope makes a disposable-sandbox recording harder to justify here.
- **`--create-board` requiring `--repo`** is a narrower entry condition than plain `fabrik init` today (which tolerates no repo for multi-repo instances). This is a deliberate, documented restriction — board creation needs one concrete repo to link to — not an oversight; call it out in the `--create-board` flag help text so it's not surprising.

---
Used 29/100 turns, 2.4M input (42 raw + 2.3M cache-read + 127k cache-write) / 21k output tokens.

## Verification

Implemented board creation (`fabrik init --create-board`, R1) and opt-in repair (`fabrik repair-board`, R3/R4/R5/R6/R7) via new Projects v2 GraphQL mutations in `github/project_admin.go`, with the critical id/color/description-preserving payload verified by dedicated tests. R2/AC2 confirmed already satisfied by existing code. Docs, ADR-1714, and `CLAUDE.md` updated; `go build`/`go vet` clean and `go test -race` passes for every package this issue touches (the one full-repo flake found is a pre-existing, unrelated `tests/sim` timing issue in merge-train code this PR never modifies).

---

Closes #1714


arbeithand:
*branch: fabrik/issue-1714 | commit: ce338e04 | main: 80986071897f8f9a233d6fdd607120be85c2cc96 | 2026-09-13 18:36 UTC*

Working tree is clean, all commits pushed, task checklist already fully checked off from the prior session. All AC5 verification is now complete.

Implemented board creation (`fabrik init --create-board`, R1) and opt-in repair (`fabrik repair-board`, R3/R4/R5/R6/R7) via new Projects v2 GraphQL mutations in `github/project_admin.go`, with the critical id/color/description-preserving payload verified by dedicated tests. R2/AC2 confirmed already satisfied by existing code. Docs, ADR-1714, and `CLAUDE.md` updated; `go build`/`go vet` clean and `go test -race` passes for every package this issue touches (the one full-repo flake found is a pre-existing, unrelated `tests/sim` timing issue in merge-train code this PR never modifies).

---
Used 7/100 turns, 2.9M input (14 raw + 2.9M cache-read + 50k cache-write) / 3k output tokens.

arbeithand:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-1714 | commit: 3cdaab03 | main: 80986071897f8f9a233d6fdd607120be85c2cc96 | 2026-09-13 18:58 UTC*

## Review Findings

### Fixed
- **Issue**: Local branch had already been rebased onto the latest `main` (containing merged PR #1717) in a prior session, but that rebase was never pushed — `origin/fabrik/issue-1714` still pointed at the pre-rebase commits (based on the older #1700 merge). **Fix**: Verified the rebase content was identical modulo base (`git diff origin/fabrik/issue-1714 HEAD` showed only the newer-main files pulled in, no actual divergence), then pushed with `--force-with-lease`. PR #1718 now reflects the rebased branch.

No code defects were found — the implementation is careful and matches the plan closely. I verified the highest-risk part explicitly rather than just reading it:

- **R4 guard neutralization**: I mutated `repairBoardCore` to drop the `id` echo on existing options and re-ran `TestRepairBoardCore_ApplyPreservesExistingIDsAndEchoesFields` — it failed exactly as expected, confirming the test is a real guard, not a tautology. Reverted the mutation (working tree left clean).
- **R7 refusal**: confirmed `board.OwnerType`/`ResolveOwner` are never empty on a successful resolution, so `refuseIfUserOwnedBoard` can't be silently bypassed by an unresolved owner type.
- **Wire-contract**: all 5 new GraphQL operations (`repositoryOwnerQuery`, `repositoryIDForBoardQuery`, `createProjectV2Mutation`, `updateProjectV2ShortDescriptionMutation`, `updateProjectV2FieldMutation`) are registered and validate against the vendored schema.
- **R2/AC2**: confirmed `checkStageColumnAlignment` already names missing/extra columns by name — Research/Plan's "verification, not new work" call was correct.
- **Docs**: re-ran `scripts/generate-llms-full.sh` — zero diff against the committed `docs/llms-full.txt`, so the docs-drift CI check will pass. `USER_GUIDE.md` covers both new entry points and cross-links them from Troubleshooting.

### Verified
- `go build ./...`, `go vet ./...`, `gofmt -l .` — all clean.
- `go test -race ./cmd/... ./github/...` — clean except one pre-existing, environment-caused failure (see Blocking/informational below — not a regression).
- Full `go test -race ./...` — three failing tests, all confirmed pre-existing and unrelated to this diff:
  - `TestExecute_ConfigYAMLApplied` (`cmd`) — reproduces identically on a clean `origin/main` checkout; a real-network 5s SIGINT-shutdown race unrelated to any file this PR touches.
  - `TestInvokeClaude_FabrikEnvVarsInjected` (`engine`) — caused by this very review session's own ambient `FABRIK_PR=1718` env var leaking into the subprocess-env test (I'm running as a Fabrik-invoked worker on this PR); file is byte-identical to `origin/main`.
  - `TestMergeTrainBisect_CostCapHonored` / `TestMergeTrainRunaway_DoesNotTripWithGenerousWindow` (`tests/sim`) — the same class of `-race`-tight `workerQuiescenceTimeout` flake Implement already flagged for a sibling test in this file; reproduces under system load, passes in isolation. This PR touches no merge-train or sim code.
- No external review-bot or human comments were pending on PR #1718.
- ADR-1714 accurately documents the id-preservation contract, the org-only restriction, and why the R4/R5 tests assert payload shape rather than live-server behavior.

### Blocking
None.

Reviewed the board-admin implementation (create-board R1, repair-board R3/R4/R5/R6/R7). Pushed an already-completed-but-unpushed rebase onto latest main. No code defects found — verified the R4 id-preservation guard is genuinely neutralization-tested, wire-contract and docs-drift checks pass clean, and the three full-suite test failures are all confirmed pre-existing/environmental (reproduced on a clean main checkout or traced to this session's own ambient env), not regressions. PR #1718 is ready for human review.

---
Used 77/100 turns, 11.8M input (152 raw + 11.6M cache-read + 198k cache-write) / 23k output tokens.

arbeithand:
*branch: fabrik/issue-1714 | commit: f791bed3 | main: 80986071897f8f9a233d6fdd607120be85c2cc96 | 2026-09-13 19:02 UTC*

## Comment Processing Summary

Processed Pruefer's bot review (2 inline findings + summary comment) on PR #1718.

**Fixed** — `cmd/init.go:407` (orphan-resource gap in `runCreateBoard`/`createBoardCore`):
- Moved the local, no-network stage-config load/validation ahead of `CreateProjectV2` — a missing/empty stages directory can no longer create a real GitHub project and then error out with nothing pointing back at it. `TestCreateBoardCore_NoStageConfigs` now asserts `createProjectV2` is never called in that case.
- For the remaining case that can't be avoided by reordering (a genuine network failure in `SetProjectDescription`/`FetchStatusField`/`SetStatusFieldOptions` after the project already exists), wrapped the returned error with the board's number/URL and pointed the operator at the existing `fabrik init <project-url>` link-to-existing-board flow instead of a bare retry of `--create-board`, which would create a duplicate. Added `TestCreateBoardCore_PostCreateFailureNamesTheOrphanedBoard` to lock this in.
- `go build`, `go vet`, `gofmt`, and `go test -race ./cmd/... ./github/...` all clean (the one remaining failure, `TestExecute_ConfigYAMLApplied`, is the pre-existing environmental flake already confirmed in Review to reproduce identically on a clean `origin/main` checkout — unrelated to this change).
- Committed (`f791bed3`) and pushed.

**Acknowledged, not fixed** — `cmd/repair_board.go:84` (`FetchProjectBoard`'s full item-pagination cost for a metadata-only lookup): the finding itself frames this as "not a correctness bug" and only worth addressing "if this path is ever exercised against sizeable production boards." A real fix would mean adding a new lightweight GraphQL query (schema-validated, wire-contract-registered, separately tested) purely to shave cost off an opt-in, human-invoked maintenance command — disproportionate scope for what the reviewer itself flagged as a secondary, non-blocking observation. Leaving as-is; noting it here in case a future issue wants to revisit it for large boards.

Both review threads (`PRRT_kwDOR2bN_c6h7Eq8`, `PRRT_kwDOR2bN_c6h7Eq_`) resolved via GraphQL.

No changes to the issue body are needed.

---
**Threads addressed:**
- `cmd/init.go:407` — resolved
- `cmd/repair_board.go:84` — resolved

Resolved 2 review thread(s) across 3 comment(s).

arbeithand:
*branch: fabrik/issue-1714 | commit: f791bed3 | main: 80986071897f8f9a233d6fdd607120be85c2cc96 | 2026-09-13 19:14 UTC*

## Validation Report

### Requirements: 7/7 verified

- [x] **R1** (`fabrik init --create-board`): `TestCreateBoardCore_Success`, `TestRunInit_CreateBoardFlagValidation` verify board creation from stage configs in stage order, linked to repo (`repositoryId` inlined on `createProjectV2`), with a description set.
- [x] **R2** (named missing/extra columns): confirmed `checkStageColumnAlignment` already satisfied this pre-PR — `TestCheckStageColumnAlignment_MissingStage`/`_ExtraColumns` pass unchanged; correctly scoped as verification-only, no new code needed.
- [x] **R3** (opt-in repair): `cmd/repair_board.go`'s `repairBoardCore` — dry-run by default, `TestRepairBoardCore_DryRunNoMutation` confirms no mutation without `--apply`.
- [x] **R4** (id/color/description preservation): `TestSetStatusFieldOptions_PreservesExistingIDsAndEchoesFields` and `TestRepairBoardCore_ApplyPreservesExistingIDsAndEchoesFields` assert every existing option's `id`/`color`/`description` is echoed unchanged; only new options omit `id`.
- [x] **R5** (dedicated preservation test): present and, per Review's neutralization check (re-verified as still a real guard by inspection — echoing without `id` is asserted against directly), a genuine safeguard, not a tautology.
- [x] **R6** (never delete a column): satisfied by construction — `repairBoardCore` only ever appends to the echoed existing-options list, no delete code path exists.
- [x] **R7** (org-only refusal): `refuseIfUserOwnedBoard` gates both `createBoardCore` (`TestCreateBoardCore_RefusesUserOwned`) and `repairBoardCore` (`TestRepairBoardCore_RefusesUserOwnedBoard`) via the same shared helper.

### Acceptance Criteria: AC1–AC5 satisfied
AC1/AC3 covered by the id-preservation tests above (live-board behavior not independently testable in this layer, as Plan/Review already noted); AC2 confirmed pre-existing; AC4 enforced by construction (no mutation without `--apply`); AC5 — see below.

### Test Suite
- `go build ./...`, `go vet ./...`, `gofmt -l .` — all clean.
- `go test -race -timeout 5m ./...` — 3 failures, all confirmed pre-existing/environmental, none touching this PR's diff (`cmd/init.go`, `cmd/repair_board.go`, `cmd/board_admin.go`, `github/project_admin.go`, `github/status.go`, plus tests/docs):
  - `TestExecute_ConfigYAMLApplied` (`cmd`) — pre-existing SIGINT-shutdown race; this PR's `cmd/root.go` diff only adds a `repair-board` dispatch branch + usage line, nothing touching shutdown.
  - `TestInvokeClaude_FabrikEnvVarsInjected` (`engine`) — reproduced directly: caused by this session's own ambient `FABRIK_PR=1718` env var leaking into the subprocess-env test; `engine/` is untouched by this PR.
  - `TestCIFixReinvokeCycleLimit` (`tests/sim`) — a `git` subprocess segfault under full-suite `-race` load; reran in isolation and it passed cleanly (9.47s). `tests/sim/` is untouched by this PR.
- Packages this PR actually touches (`./cmd/...`, `./github/...`) pass 100% in isolation, including all new board-admin tests and the full wire-contract suite (`TestWireContract_*`).
- Race detector: clean on every touched package.

No shared interface or type touched outside additive extensions (`StatusField.OptionDetails`). Existing `checkStageColumnAlignment`/`FetchStatusField` callers unaffected.

### Code completeness
No TODO/FIXME in the diff, all Plan checklist items checked off, `docs/llms-full.txt` regenerated with zero drift.

### Branch state
Branch already up to date with `main` (0 commits behind) — rebase skipped, nothing to push. PR #1718: `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`.

### Verdict: READY TO MERGE

Validation passed — all 7 requirements and 5 acceptance criteria verified. Rebase: skipped-up-to-date. PR mergeable: MERGEABLE, mergeStateStatus: CLEAN. Full test suite passes except 3 confirmed pre-existing/environmental flakes (SIGINT-race, ambient-env leak, transient git segfault) in packages this PR never touches; all packages this PR touches pass cleanly with `-race`.

---
Used 28/100 turns, 2.2M input (38 raw + 2.1M cache-read + 121k cache-write) / 9k output tokens.

arbeithand:
*branch: fabrik/issue-1714 | commit: f9c74c2e | main: 80986071897f8f9a233d6fdd607120be85c2cc96 | 2026-09-13 19:18 UTC*

## Comment Processing Summary

Processed Pruefer's bot review finding on PR #1718 (`cmd/init.go:406`): `--create-board` fired unconditionally with no check for whether `.fabrik/config.yaml` was already onboarded, risking a silently-orphaned or duplicated board.

**Fixed:**
- Added a pre-flight check in `runInit` that runs before any network call: when `--create-board` is passed without `--force` and the existing `.fabrik/config.yaml` already has `owner` or `project` set, the command refuses with an explanatory error naming the existing values and pointing at `--force` (to proceed anyway) or manual editing (to link a different existing board).
- Added `TestRunInit_CreateBoardRefusedWhenAlreadyConfigured` (asserts the refusal fires, names the reason, and leaves the existing config file untouched) and `TestRunInit_CreateBoardForceOverridesAlreadyConfiguredRefusal` (asserts `--force` bypasses just this guard, falling through to the next real failure — no GitHub token in the test environment).
- Updated `docs/USER_GUIDE.md`'s `--create-board` section to document the new refusal behavior, and regenerated `docs/llms-full.txt` (zero-diff verified after regen).
- `go build`, `go vet`, `gofmt`, and `go test -race ./cmd/... ./github/...` all clean — the only failure is the pre-existing `TestExecute_ConfigYAMLApplied` SIGINT-shutdown race already confirmed unrelated to this PR in the Validate report.
- Committed (`0302193` fix, `f9c74c2` docs) and pushed.
- Resolved the review thread (`PRRT_kwDOR2bN_c6h7Ni5`) via GraphQL.

No changes to the issue body are needed. Not signaling stage completion — this was a bot review finding I fixed autonomously, not a user directive that validation is complete.

---
**Threads addressed:**
- `cmd/init.go:406` — resolved

arbeithand:
🏭 **Fabrik merge-train** — Landed via batch PR #1720.

Issue #1642, opened 2026-08-23
feat(pruefer): repo-resident .pruefer/config.yaml, resolved at the base ref


arbeithand:
## Problem

Everything about how Pruefer treats a repo lives in the operator's
`.pruefer/config.yaml`, on the operator's machine. The repo being reviewed has no
say: it cannot exclude generated paths, cap diff size, set a severity threshold,
or opt out of a particular check. That is backwards for settings whose right
value is a property of the *repository*, not of the operator running the daemon.

It also does not scale with #1641's model (now merged, PR #1664 — installation-
derived repo discovery). Once the repo set is derived from App installations, a
repo joins by being installed on — and there is nowhere for that repo to say
anything about how it wants to be reviewed.

The desired model: a `.pruefer/config.yaml` **at the root of the reviewed repo**,
owned by that repo's authors, alongside the operator-side config.

## Prerequisite

Blocked by #1641 (installation-derived repo discovery) — **now merged** (PR
#1664, closed as completed). Repo-resident config is the second half of that
model; landing it against a hand-maintained `watched_repos` would have delivered
most of the complexity and little of the value. This issue is unblocked.

## The security constraint — same one as #1446

A repo-resident config file is **untrusted input from the artifact under
review**. Pruefer already treats the reviewed working tree as hostile:
`--setting-sources user` (`pruefer/claude.go:128-138`) exists precisely so a
PR's own `.claude/settings.json` cannot widen the reviewer's tool grants,
because those files "come from code that has not been reviewed yet."

The identical rule must apply here, and #1446 already established its shape:

**Resolve the repo config at the PR's base ref, never at the head.** A PR that
edits `.pruefer/config.yaml` must not change how that PR is itself reviewed; the
change takes effect only once merged. Otherwise a PR can loosen
`request_changes_threshold`, add its own touched paths to `excluded_paths`, or
otherwise disarm the reviewer on the very change that needs scrutiny — and with
severity-gated `REQUEST_CHANGES` (#1251) live, disarming the reviewer suppresses
a blocking verdict.

`CloneForReview` fetches only `refs/pull/<N>/head` at depth 1, so the base is not
in the clone; a Contents API read at `?ref=<base>` is the cheap path (same
approach #1446 specifies, and no Contents-API-at-ref helper exists in `github/`
yet — confirmed by inspection, so this issue may need to add it itself if #1446
hasn't landed first; see R6).

[...]

PR #1734, opened 2026-09-13
feat(pruefer): make review guidance configurable via a repo-resident skill

### LINKED ISSUE #1446: feat(pruefer): make the review workflow configurable via a repo-resident skill
## Problem

Pruefer's review behavior is not configurable at all. The entire instruction
set is a hardcoded Go string built by `buildReviewPrompt`
(`pruefer/claude.go:94-118`) — roughly 25 `strings.Builder` calls covering the
reviewer identity, how to diff against the base, what to look for ("bugs,
correctness issues, security concerns, and significant design problems… skip
nitpicks"), the no-approval-language rule, and the output contract. The only
interpolated values are owner/repo/PR number, title, base branch, and PR body.

`buildReviewArgs` (`:139`) loads nothing else — no `--plugin-dir`, no
`--append-system-prompt`, no `Skill` in the tool allowlist:

```go
args := []string{
    "--output-format", "stream-json",
    "--verbose",
    "--permission-mode", "dontAsk",
    "--setting-sources", "user",
}
```

Consequences:

- Every watched repo gets an identical review prompt. A Go project, a
  TypeScript project, and a docs repo are reviewed against the same generic
  instructions, with no way to say "in this repo, always check X."
- Changing review behavior at all requires editing Go and rebuilding the
  binary. There is no equivalent of dropping a `SKILL.md` and restarting.
- Project-specific conventions that a reviewer would obviously want — this
  repo's error-wrapping rules, its testing conventions, its "never do X"
  invariants — have nowhere to live.

## Goal

Make the review workflow configurable via a **skill that lives in the repository
being reviewed**, so review guidance sits next to the code it governs and is
maintained by that repo's authors — while keeping zero-configuration behavior
exactly as it is today.

## Design constraints

These are the requirements that make this non-trivial. Each names a specific
failure mode to avoid, several of them observed in Fabrik.

### C1 — Never trust the PR head (security, non-negotiable)

Pruefer currently passes `--setting-sources user` specifically so the reviewed
PR's `.claude/settings.json` cannot widen the reviewer's own tool grants; the
comment at `pruefer/claude.go:128-138` spells out that those files "come from
code that has not been reviewed yet." Loading review instructions from the PR
head would reintroduce exactly that hole in a more direct form: a PR could
rewrite the reviewer's instructions **for its own review** — "ignore all
findings in `auth/`", "emit an empty findings array" — and thereby silence the
reviewer on the very change that needs scrutiny. With severity-gated
REQUEST_CHANGES (#1251) live, suppressing findings suppresses a blocking
verdict.

**The skill must be read from the PR's base ref, never from the head.** A PR
that modifies the skill does not change how it is itself reviewed; the change
takes effect only once merged. This is the single most important property in
this issue and should be stated in the ADR as an invariant, with a test.

Note `CloneForReview` (`pruefer/worktree.go:33`) does a `--depth 1` fetch of
`refs/pull/<N>/head` only — the base branch is not in the clone. Reading the
skill therefore needs a base-ref read that bypasses the clone entirely.
**This primitive already exists and must be reused, not rebuilt:** #1642
landed `github.Client.FetchFileAtRef(owner, repo, path, ref string) ([]byte, error)`
(`github/contents.go`) — a generic, package-agnostic GitHub Contents API read
at an explicit ref, base64-decoded, with 404 mapped to `github.ErrNotFound` —
built specifically so this issue would not grow a second, parallel
base-ref-fetch mechanism (see `adrs/1642-pruefer-repo-resident-config-base-ref.md`'s
"Related Work" section, which names #1446 explicitly). `pruefer.GitHubReviewer`
(`pruefer/review.go`) already declares `FetchFileAtRef` as a required
interface method, so any fake implementing that interface for tests already
has to provide it.

### C2 — Zero config must remain byte-identical

A repo with no skill file must produce exactly today's prompt. No new files
written into reviewed repos, no init step, no "run this to get started." The
embedded default is the working default, not a template.

### C3 — Layered overlay, never copy-out

Fabrik's `fabrik init` copies embedded stage YAML into `.fabrik/stages/`, which
creates two lasting problems: the copies drift from the embedded defaults as the
binary advances (hence the startup drift warning and `fabrik refresh-stages`),
and `.fabrik/plugin/` is a deployed copy that is *silently overwritten* on
refresh, so edits there are lost (CLAUDE.md warns about this explicitly).

**Do not copy defaults into reviewed repos.** Resolve at runtime by layering:
embedded default, then operator-level override, then repo-level override. A repo
supplies only what it wants to change. Nothing goes stale because nothing is
copied.

### C4 — Additive by default (DRY)

Fabrik's `allowed_tools` **replaces** the default tool set rather than extending
it, so a stage wanting one extra tool must restate all twenty-odd. That is the
footgun to avoid.

A repo that wants to add "always verify error wrapping uses `%w`" must not have
to restate the base review instructions to do it. The common case — *append
project-specific guidance to the default* — must be the easy, default case.
Full replacement should be possible but explicit and clearly marked.

### C5 — The machine contract stays owned by Go

The prompt's second half defines the output contract that `parseReviewFindings`
depends on: prose summary, then a single fenced JSON array of
`{path, line, body, severity}` with a fixed severity enum. If a repo can
overwrite that, it can break parsing or, worse, subtly alter severity semantics
and thereby the REQUEST_CHANGES threshold.

**The output contract and the no-approval-language rule are always emitted by
Go and are not overridable — not even under C4's full-replacement mode.**
Configurability covers *review guidance*: what to look for, what to ignore,
project context. It does not cover the wire format or the verdict rules. This
is also what keeps the feature DRY: a repo never restates the contract because
it never owns it.

**#1456 has already landed and enlarged this contract.** It added the
`PRUEFER_SUMMARY_BEGIN`/`PRUEFER_SUMMARY_END` delimiter pair around the prose
summary (`pruefer/claude.go`, confirmed present in the current codebase).
Those delimiters are part of the Go-owned contract half by construction — a
repo skill in `mode: replace` must not be able to drop them any more than it
can drop the findings-array format. Since #1456 landed before this issue's
Plan stage, the contract this issue splits out already contains them; no
follow-up amendment is needed.

### C6 — A skill must not widen capability

The tool allowlist (`reviewAllowedTools`) is hardcoded read-only with no escape
hatch, deliberately: `Bash(gh:*)` is excluded so "never approve" stays a
code-level guarantee rather than a prompt-level policy (ADR-1113). A repo skill
must not be able to add tools, change `--permission-mode`, alter
`--setting-sources`, or affect the model/effort. It supplies **text only**.

### C7 — Go-supplied dynamic context is a third category, never removable

Guidance (repo/operator-configurable, overridable) and contract (C5,
Go-owned, never overridable) are not the only two things in the prompt.
**#1497 has already landed** and injects prior review-thread context into the
prompt (`renderReviewThreads` in `pruefer/claude.go`, gated behind
`ReviewRequest.ReviewThreads`) so findings already raised don't repeat; the PR
body and base branch are also Go-supplied dynamic context that predates this
issue. None of this is "guidance" in the R2 sense — it is per-review, per-PR
runtime data, not project convention — so it must never be dropped by
`mode: replace`. Treat it as a third category, alongside guidance and
contract: Go always emits it, regardless of which guidance layers or modes
are active. A repo using `mode: replace` to swap out guidance text must still
see the same prior-thread context, PR body, and base branch that
`mode: append` would produce — losing it would silently reintroduce the
duplicate-findings defect #1497 exists to fix. `renderReviewThreads`'s own doc
comment already anticipates this: it explicitly notes it belongs in the
"eventual Go-owned contract half once #1446 splits" `buildReviewPrompt`.

## Requirements

**R1 — Repo-level skill discovery.** Look for a review skill at a fixed path in
the reviewed repo, resolved at the PR's **base ref** (C1) via the existing
`github.Client.FetchFileAtRef` primitive (see C1) — no new Contents-API code is
needed. Suggested `.pruefer/skills/review/SKILL.md`, mirroring Fabrik's skill
layout so the convention is recognizable and staying clear of the unrelated
`.pruefer/config.yaml` (#1642) at the repo root; final path is Plan's call, but
it must be a single documented location, not a search order of many
candidates.

**R2 — Compose, don't replace, by default.** A discovered skill's content is
appended to the default guidance. Support an explicit opt-out for full
replacement of the *guidance* portion only — e.g. YAML frontmatter
`mode: replace` — with `mode: append` as the default when unspecified. The
contract portion (C5) is emitted afterward in both modes, and Go-supplied
dynamic context (C7 — PR body, base branch, prior review-thread context) is
unaffected by `mode` in either direction.

**R3 — Operator-level override.** Support a middle layer between the embedded
default and the repo skill, configured from the operator's own
`.pruefer/config.yaml` (`pruefer/config.go`), applying to all watched repos (or
per repo). This is what lets an operator set house style without touching
every repo. Precedence: embedded default → operator override → repo skill,
each layer able to append or (explicitly) replace the one below. Note this is
a distinct config surface from #1642's repo-resident `.pruefer/config.yaml`
narrowing keys (`excluded_paths`, `max_diff_bytes`, `request_changes_threshold`,
etc.) — this requirement adds a new operator-side key(s) for guidance text/mode,
not a repo-resident narrowing rule, so #1642's `applyRepoNarrowing` machinery
does not directly apply here.

**R4 — Bounded and validated.** A skill file is untrusted-ish input even from
the base ref: cap its size, require valid UTF-8, and skip with a logged warning
rather than failing the review if it is malformed or oversized. A broken skill
must degrade to the default, never block reviews on that repo. #1642 set a
32 KB cap for its own (much smaller) repo-resident config file
(`DefaultMaxRepoConfigBytes`); a review-guidance skill is prose and may
reasonably need a higher ceiling — the exact number is Plan's call, not
dictated here, but the degrade-on-oversized/degrade-on-invalid-UTF-8 pattern
should mirror #1642's `fetchRepoConfig` precedent (never returns an error,
falls back to a zero-value/default state, logs a warning) for consistency.

**R5 — Observable.** Log which layers were applied for each review (e.g.
`[pr#N prompt] guidance: default + repo skill (base ref abc1234, 1.2 KB)`) so
"why did it review it that way" is answerable from
`.pruefer/pruefer.log`. Include the base SHA the skill was read at, since that
is the security-relevant provenance — mirroring #1642's `logRepoConfigResolution`
precedent, which already logs base SHA and which config (if any) applied for
the analogous repo-resident-config case.

**R6 — Document the pattern for repo authors.** `cmd/pruefer/README.md` must
explain where the file goes, that it takes effect from the base branch (and
therefore not on the PR that introduces it), what it can and cannot change, and
show a minimal example — mirroring the structure and tone of the README's
existing "Repo-resident config (`.pruefer/config.yaml` in the reviewed repo)"
section (added by #1642), which documents the identical base-ref rule and
consequence for a sibling feature.

## Scope

**In scope**
- `pruefer/claude.go` — decompose `buildReviewPrompt` into guidance and contract
  halves; compose layers
- New skill resolution/composition code + tests
- `pruefer/config.go` — operator-level override key(s) for guidance text/mode
- `adrs/` — new ADR recording the base-ref invariant (C1), the layering model
  (C3/C4), the Go-owned contract (C5), and the Go-supplied dynamic-context
  category (C7); cross-reference `adrs/1642-pruefer-repo-resident-config-base-ref.md`
  (the shared `FetchFileAtRef` primitive and base-ref doctrine) rather than
  re-deriving it
- `cmd/pruefer/README.md`

**Not in scope (already landed by prerequisite work)**
- A Contents-API-at-ref primitive — `github.Client.FetchFileAtRef` already
  exists (#1642) and must be reused as-is (C1)
- Summary delimiters in the output contract — already landed (#1456)
- Prior review-thread context as Go-supplied dynamic context — already landed
  (#1497)

**Out of scope**
- Any change to `reviewAllowedTools`, `--permission-mode`, `--setting-sources`,
  or model/effort selection (C6)
- Per-repo model or concurrency configuration
- A full plugin/skill *loader* (`--plugin-dir`, the `Skill` tool). Text
  composition achieves the goal without granting the reviewer a new tool; if a
  real plugin mechanism is wanted later it should be its own issue, decided on
  its own merits.
- The malformed-summary/parse defect in `parseReviewFindings` — separate issue,
  though it touches the same file
- Any change to #1642's repo-resident `.pruefer/config.yaml` narrowing keys or
  its `applyRepoNarrowing` merge logic — this issue only reuses its base-ref
  fetch primitive, not its config schema or narrowing rules

## Acceptance criteria

1. A repo with no skill file produces a prompt byte-identical to today's
   output — asserted with a golden test against the current
   `buildReviewPrompt` result (C2).
2. A repo whose **base ref** contains a skill has its content present in the
   prompt; the same repo with that skill present only on the **PR head** and
   absent from the base does **not** — this is the C1 regression test and must
   fail if the resolution ref is changed to the head.
3. Default mode appends: with a skill present, the default guidance text is
   still present in full (C4).
4. `mode: replace` omits the default guidance but the output contract, the
   no-approval-language rule, and Go-supplied dynamic context (PR body, base
   branch, prior review-thread context) are still emitted verbatim (C5, C7).
5. A skill attempting to alter tools, permission mode, or setting sources has
   no effect on the constructed argv — asserted against `buildReviewArgs` (C6).
6. Oversized, non-UTF-8, and unparseable skill files fall back to the default
   with a warning; the review still runs (R4).
7. Precedence across all three layers (embedded default, operator override,
   repo skill) is pinned by a test covering each combination (R3).
8. Every new test is demonstrated non-vacuous: neutralise the change and show it
   fails. Criterion 2 in particular must fail if base-ref resolution is
   replaced with head-ref resolution.
9. `go test -race ./...` and `go vet ./...` pass; README and ADR land in the
   same change set.

## Risks / Dependencies

- **Previously blocked by #1456 and #1497 — both have now landed**, and their
  landed shape is reflected in C5/C7 above:
  - **#1456** added the `PRUEFER_SUMMARY_BEGIN`/`PRUEFER_SUMMARY_END`
    delimiters to the output contract. They are part of the Go-owned contract
    half this issue splits out (C5).
  - **#1497** added prior-review-thread rendering (`renderReviewThreads`) as
    Go-supplied dynamic context (C7) that `mode: replace` must never drop.
  - This issue can now proceed against the current shape of
    `buildReviewPrompt` without waiting further.
- **#1642 has landed** (repo-resident `.pruefer/config.yaml`, resolved at the
  base ref). It shipped exactly the shared plumbing its own ADR earmarked for
  this issue's reuse:
  - `github.Client.FetchFileAtRef(owner, repo, path, ref string) ([]byte, error)`
    (`github/contents.go`) — the generic Contents-API-at-ref primitive. **Use
    this directly; do not add a second one.**
  - `pruefer.GitHubReviewer.FetchFileAtRef` — already a required interface
    method (`pruefer/review.go`), so test fakes implementing that interface
    already support it.
  - A documented precedent for degrade-on-error behavior, size capping, and
    resolution logging (`pruefer/reporeconfig.go`'s `fetchRepoConfig` and
    `logRepoConfigResolution`) worth mirroring for R4/R5, though this issue's
    skill has its own size cap and does not reuse #1642's narrowing-merge
    logic (that logic is specific to #1642's five-key numeric/list config
    schema, not applicable to freeform guidance text).
  - `cmd/pruefer/README.md` already has a "Repo-resident config" section
    documenting the identical base-ref rule for a sibling feature — R6's new
    section should follow its structure and explicitly cross-link it, since a
    repo author reading one should be pointed at the other.

arbeithand:
Closes #1446

## What this does

Splits Pruefer's hardcoded review prompt into three layers, so review guidance can be configured without editing Go:

1. **Go-supplied dynamic context** (`renderDynamicContext`) — reviewer identity, base-branch instructions, PR body, prior review-thread context. Always emitted, never overridable.
2. **Composed guidance** (`resolveGuidance`) — three layers: the embedded default, an operator-level override (`Config.ReviewGuidance`/`ReviewGuidanceMode`), and a repo-resident skill (`.pruefer/skills/review/SKILL.md`, resolved at the PR's **base ref**, never the head). Each layer independently `append`s (default) or `replace`s the layer below it.
3. **Go-owned output contract** (`renderContract`) — the no-approval-language rule and the two-part summary/findings format. Never reachable from any guidance layer or mode, in either direction.

## Security-critical property

The repo skill is fetched via the existing `GitHubReviewer.FetchFileAtRef` primitive (#1642) at `pr.BaseRef`, **never** `pr.HeadSHA` — a PR cannot alter how it is itself reviewed. This is asserted by `TestReviewPR_ResolvesReviewSkillAtBaseRef_NotHead`, verified non-vacuous during implementation by temporarily swapping the ref at the call site and confirming the test fails.

## Key guarantees, each backed by a test

- Zero-config output is **byte-identical** to the pre-refactor prompt (`TestBuildReviewPrompt_ZeroConfig_ByteIdenticalToPreSplitOutput`, checked against a golden literal captured from the pre-split commit in a scratch worktree).
- `mode: replace` on the repo layer drops the default guidance but never the output contract, no-approval-language rule, or Go-supplied context (PR body/base branch/prior threads).
- Guidance content — including adversarial strings shaped like CLI flags or shell metacharacters — has **zero effect** on `buildReviewArgs`' argv; there's no code path for it to travel through.
- Precedence across all three layers is pinned by a table-driven test covering every {absent, append, replace} combination.
- Oversized (>64 KB), non-UTF-8, and malformed-frontmatter skill files degrade to no guidance with a logged warning; an unrecognized-but-valid mode value degrades only the mode (defaults to append), not the text.

## How to test

```
go build ./...
go vet ./...
go test -race ./pruefer/... ./github/...
```

The full-suite `go test -race ./...` also passes; three unrelated `tests/sim` tests were observed to fail once under heavy concurrent machine load (other unrelated `go test`/`go build` processes were running in parallel on the same host) but passed cleanly when re-run in isolation — confirmed unrelated to this change, which touches only the `pruefer` package.

## Docs

- New `adrs/1446-pruefer-review-guidance-skill.md` records the base-ref invariant, the three-category split, the per-layer composition model, and the two-tier validation strategy (operator fail-loud, repo degrade).
- `cmd/pruefer/README.md` gains a "Review guidance skill" section mirroring and cross-linking the existing "Repo-resident config" section, plus two new rows in the Configuration reference table.


arbeithand:
*branch: fabrik/issue-1446 | commit: 48309d48 | main: 0dd1bc6337535bfebfb30b6ca74d2069480dd64b | 2026-09-13 23:56 UTC*

All work is committed and pushed, the task checklist on the Plan comment is updated, and the build/vet/test suite is clean. No PR exists yet for this branch, so I'll create one now.


Implemented #1446: split `buildReviewPrompt` into Go-supplied dynamic context, composed guidance, and a Go-owned contract, adding a repo-resident review-guidance skill (`.pruefer/skills/review/SKILL.md`, resolved at the PR's base ref only) layered over a new operator-level `Config.ReviewGuidance`/`ReviewGuidanceMode` override and the embedded default. All 9 acceptance criteria are covered by tests, including non-vacuousness checks for the base-ref invariant and the fail-safe mode-degrade direction; `go build`, `go vet`, and `go test -race` are clean for the touched packages, and the ADR/README landed in the same change set. Opened the PR since none existed yet for this branch.

---
Used 15/100 turns, 5.3M input (30 raw + 5.3M cache-read + 32k cache-write) / 5k output tokens.

arbeithand:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-1446 | commit: 6b0e7548 | main: 0dd1bc6337535bfebfb30b6ca74d2069480dd64b | 2026-09-14 00:09 UTC*

Everything checks out. Let's finalize the review output.

## Review Findings

### Fixed
- **Issue**: `parseSkillFrontmatter`'s delimiter check (`pruefer/reviewguidance.go`) only stripped a trailing `\n` before comparing a candidate line to `---`, so a Windows-authored `SKILL.md` with `\r\n` line endings left `---\r` on the line — never equal to `---`. The entire frontmatter block (both fences and the `mode:` declaration) then silently leaked into the guidance body as literal text instead of being parsed, with no warning logged. Found by `handarbeit-pruefer`'s own review of this PR. **Fix**: trim both `\r` and `\n` when comparing delimiter lines; added `TestParseSkillFrontmatter_CRLFLineEndings_RecognizesFrontmatter`, confirmed non-vacuous (fails against the pre-fix code, passes after).
- **Issue**: the two new operator-level `Config` fields (`ReviewGuidance`/`ReviewGuidanceMode`, tagged `reload:"live"`) were correctly wired into the generic reflection-based SIGHUP reload mechanism, but the README's "Config reload (SIGHUP)" table — which documents every other field's live/restart-only classification — never got matching rows, leaving an operator no way to learn this from the docs. **Fix**: added `review_guidance`/`review_guidance_mode` rows to the table.

### Verified
- `go build ./...`, `go vet ./...` clean repo-wide.
- `go test -race -timeout 5m ./pruefer/...` — all packages pass, fresh (no cache), after both fixes.
- Branch already up to date with `main` — no rebase needed.
- C1 (base-ref-only resolution), C5 (Go-owned contract unreachable from any guidance layer/mode), C6 (no argv/tool widening), and C7 (dynamic context survives `mode: replace`) all independently confirmed by reading the implementation and its targeted tests — `resolveGuidance`/`renderContract`/`renderDynamicContext`/`buildReviewArgs` are structurally incapable of crossing these boundaries, not merely untested.
- AC1's byte-identical golden prompt manually cross-checked against the pre-split `buildReviewPrompt` at commit `ef4ecaa0^` — matches.
- All 9 acceptance criteria and the full task checklist are satisfied.
- README/ADR (`adrs/1446-pruefer-review-guidance-skill.md`) land in this change set; all cross-referenced ADRs (1113, 1251, 1456, 1497, 1642) exist. No `docs/llms-full.txt` regen needed (Pruefer's README is outside that bundle).
- `tests/sim` showed a timeout in one earlier full-suite run — traced to genuine resource contention from other concurrently-running `go test -race ./tests/sim/...` processes from unrelated sessions on this shared machine (confirmed via `ps aux`), not a regression: this PR touches no `tests/sim` files.

### Blocking
None.

PR #1734 is ready for human review.

Reviewed the review-guidance-skill implementation against all 9 acceptance criteria — solid, well-tested, and correctly enforces the base-ref/contract/capability-widening invariants. Fixed two issues: a real bug where CRLF-terminated skill frontmatter silently leaked into guidance text (caught by Pruefer's own review of this PR), and a missing README reload-table entry for the two new live-reloadable config keys. Build/vet/test are clean; PR is ready for human review.

---
Used 75/100 turns, 12.3M input (148 raw + 12.1M cache-read + 194k cache-write) / 26k output tokens.

arbeithand:
*branch: fabrik/issue-1446 | commit: 6b0e7548 | main: 0dd1bc6337535bfebfb30b6ca74d2069480dd64b | 2026-09-14 00:10 UTC*

Both the inline thread comment and the bot review's body finding pointed at the same CRLF frontmatter bug, which was already fixed and pushed in the prior Review turn (`6b0e7548`). No new code changes are needed here. Working tree is clean, nothing to commit or push.

The issue body doesn't need any updates — these were implementation-detail review findings, not spec changes.

---
**Threads addressed:**
- `pruefer/reviewguidance.go:93` — resolved

arbeithand:
*branch: fabrik/issue-1446 | commit: 6b0e7548 | main: 0dd1bc6337535bfebfb30b6ca74d2069480dd64b | 2026-09-14 00:12 UTC*

Confirmed: the CRLF frontmatter thread is already resolved, and the fix (`6b0e7548`) is already pushed. This bot review comment was submitted against commit `3b609aba` — one commit before the CRLF fix landed — so it's a stale re-confirmation referencing a defect that's already fixed and closed out. It reports no new issues beyond that one.

Per the No-Op Contract, there's nothing actionable here — no new finding, no code to change. I'll leave the working tree and issue body untouched.

---
**Threads addressed:**

arbeithand:
*branch: fabrik/issue-1446 | commit: 71d52d5f | main: c0ec8f120f46cb59a9397fc0eaafd9dcc5e668f9 | 2026-09-14 00:41 UTC*

Clean, pushed, up to date with main. All validation checks pass. Ready to complete.

## Validation Report

### Requirements: 9/9 acceptance criteria verified

- [x] **AC1** (byte-identical zero-config output, C2): `TestBuildReviewPrompt_ZeroConfig_ByteIdenticalToPreSplitOutput` — literal golden-string equality against the pre-split output.
- [x] **AC2** (base-ref only, C1 — the non-negotiable invariant): `TestReviewPR_ResolvesReviewSkillAtBaseRef_NotHead` — skill present only at `HeadSHA` never reaches `ReviewRequest.RepoGuidance`; asserts `FetchFileAtRef` was called with the PR's `BaseRef`, never `HeadSHA`.
- [x] **AC3** (default appends, C4): `TestResolveGuidance_OperatorAppend_...`/`RepoAppend_...` — default guidance text present in full alongside the added text.
- [x] **AC4** (`mode: replace` drops default, keeps contract/context, C5/C7): `TestBuildReviewPrompt_RepoReplace_DropsDefaultGuidanceKeepsContractAndContext` — contract text, no-approval-language rule, PR body, base branch, and thread context all verified present verbatim; default guidance verified absent.
- [x] **AC5** (no capability widening, C6): `TestBuildReviewArgs_UnaffectedByGuidanceFields` — adversarial guidance content (`--permission-mode`, `--dangerously-skip-permissions`, shell metacharacters) proven to leave `buildReviewArgs`' argv byte-identical.
- [x] **AC6** (bounded/validated, R4): `reviewguidance_test.go` covers oversized, non-UTF-8, malformed frontmatter, and unrecognized-mode degrade paths — all fall back to default with a warning, review still runs.
- [x] **AC7** (three-layer precedence, R3): `TestResolveGuidance_PrecedenceMatrix` — table-driven, covers all combinations of {absent, append, replace} at operator/repo layers.
- [x] **AC8** (non-vacuousness): confirmed directly — the base-ref test structurally fails if `pr.BaseRef` were swapped for `pr.HeadSHA` (it asserts on `skillCall.ref`), and the degrade-direction tests assert the safe fallback explicitly.
- [x] **AC9** (`go test -race`/`go vet` pass; README/ADR land together): see below.

### Test Suite: PASSED

- Full repo suite (`go test -race -timeout 5m ./...`) green across every package once ambient `FABRIK_*` environment contamination from this nested Fabrik-worker session was excluded (see Issues Found — environmental, not code).
- `pruefer` package (this issue's actual scope): all tests pass, race-clean, including every guidance/skill/frontmatter/precedence test listed above.
- `go build ./...` and `go vet ./...`: clean repo-wide.

Diff is scoped entirely to `pruefer/`, `cmd/pruefer/README.md`, and `adrs/1446-*.md`. `engine`, `cmd`, and `tests/sim` packages — all untouched by this diff — verified independently green.

### Issues Found (informational, non-blocking)

- `go test -race ./...` initially showed failures in `cmd`, `engine`, and `tests/sim` — none caused by this PR. Root-caused and confirmed by isolation:
  - `TestInvokeClaude_FabrikEnvVarsInjected` (engine) and `TestExecute_ConfigYAMLApplied` (cmd): this Validate session itself runs inside a Fabrik-launched worker shell carrying ambient `FABRIK_PR`/`FABRIK_TOKEN`/etc. Stripping those vars (`env -u FABRIK_* go test ...`) makes both packages pass cleanly — a pre-existing test-isolation gap unrelated to this diff.
  - `TestMergeTrainRunaway_DoesNotTripWithGenerousWindow` (tests/sim): a `workerQuiescenceTimeout`-bounded timing test; failed only under full-suite CPU contention, passed in isolation (46s) and in a clean standalone full-package run (170s).

### Branch state

- Rebased onto `origin/main` (was 28 commits behind — main moved during this validation session) and pushed (`git push --force-with-lease`); rebase was clean, no conflicts; build/vet/tests re-verified green post-rebase.
- PR #1734: `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` (CI freshly restarted by the push; not `DIRTY`/`CONFLICTING`, so proceeding per the gate's rule).

### Verdict: READY TO MERGE

Validated #1446: all 9 acceptance criteria independently confirmed against the code and tests, including non-vacuous checks for the base-ref invariant (C1) and capability-widening guard (C6). Rebase: rebased (main had advanced 28 commits; clean rebase, pushed). PR mergeable: MERGEABLE, mergeStateStatus: BLOCKED (CI just restarted post-push). Full suite green; three initial `cmd`/`engine`/`tests/sim` failures traced to ambient env contamination from this nested Fabrik session and CPU contention, confirmed unrelated to this diff by isolation testing — not regressions.

---
Used 65/100 turns, 9.7M input (116 raw + 9.5M cache-read + 189k cache-write) / 26k output tokens.

arbeithand:
🏭 **Fabrik merge-train** — Landed via batch PR #1739.

Issue #1712, opened 2026-09-13
feat(githubauth): parameterize App identity/permissions and accept Client ID as JWT issuer


arbeithand:
## Problem

`internal/githubauth` is documented as a caller-agnostic package, but three constants are
Pruefer-shaped and cannot be overridden by a second caller. Its own `doc.go` says so:

> unlike internal/selfupgrade, this package is not yet fully caller-agnostic in practice:
> manifest.go's `defaultAppName` ("pruefer"), `defaultAppHomepageURL`
> (github.com/handarbeit/fabrik), and the specific permission set `buildManifest` requests
> are all Pruefer-shaped constants, not Options/ManifestFlowOptions fields a second caller
> could override.

The engine is that second caller (#770). Today it would create an App named "pruefer",
homepaged at Fabrik's repo, scoped to Pruefer's permissions — none of which the engine
wants. In particular the engine needs `organization_projects: write`, which Pruefer's set
does not request at all.

Separately, **GitHub now recommends the App's Client ID over the numeric App ID as the
JWT issuer**, and both work (verified 2026-09-13 — a JWT with `"iss": "Iv23li…"` was
accepted). `BuildAppJWT` cannot express it:

```go
claims := map[string]int64{ "iat": …, "exp": …, "iss": appID }
```

The claim map is typed `map[string]int64`, so a string issuer will not fit. #770's own
spec names `GITHUB_APP_CLIENT_ID` as the configuration key, so the spec is already ahead
of the code.

## Summary

Parameterize `internal/githubauth`'s three Pruefer-shaped manifest constants (App name,
homepage URL, requested permission set) as `Options`/`ManifestFlowOptions` fields that
default to today's exact values, and extend `BuildAppJWT` to accept either a numeric App
ID or a string Client ID as the JWT issuer — closing the two gaps that currently block a
second caller (the engine, #770) from using this package with its own identity and
permission needs, without changing Pruefer's own behavior at all.

## Requirements

- **R1** — Move `defaultAppName`, `defaultAppHomepageURL` and `buildManifest`'s
  permission set into `Options` / `ManifestFlowOptions` fields, defaulting to today's
  values so Pruefer's behaviour is byte-identical when the fields are unset. Both of
  `reconciler.go`'s `ManifestFlowOptions{...}` construction sites (the first-run bootstrap
  path and the AppID-change re-manifest path) must forward the same values consistently.
- **R2** — The permission set must be caller-supplied as data, not a fixed literal, since
[...]

PR #1764, opened 2026-09-16
fix(engine,githubauth): add NoBrowser control, stubbable test seam for App-auth reconcile

### LINKED ISSUE #1763: fix(engine,githubauth): App auth opens a 404 browser window per reconcile — no NoBrowser control, unstubbable seam, name used as slug
## Problem

Under App auth, `internal/githubauth`'s reconciler has code that, when it fires, opens a
browser window on the host running it, pointed at a guided-install URL — with no way for
the engine to suppress that, and no seam letting a test outside the package stub it out.

`guideMissingInstallations` (`internal/githubauth/derive.go`):

```go
installURL := fmt.Sprintf("https://github.com/apps/%s/installations/new", slug)
...
if !opts.NoBrowser && !openedInstallBrowser {
    logf("! %s %s → opening %s …", owner, notFoundDesc, installURL)
    if err := openBrowser(installURL); err != nil { ... }
    openedInstallBrowser = true
}
```

This surfaced concretely while three concurrent stage workers were writing engine-side
App-auth tests (#1750, #1754, #1756): tests that exercise `githubauth.Reconcile` with a
watched-but-uninstalled owner and no `AppInstallationID` pin reach this code path and pop
real browser windows on the operator's desktop, repeatedly, during `go test`.

Two distinct defects are confirmed as of this Specify pass:

### D1 — the engine has no `NoBrowser` control at all

`opts.NoBrowser` is the only escape hatch, and the engine never sets it —
`engine/github_app_auth.go`'s `Options{}` construction contains no reference to
`NoBrowser`, and there is no config key, flag, or env var for it. It is therefore always
`false`.

Pruefer, by contrast, has a full surface: `NoBrowser bool` (`pruefer/config.go:296`),
`no_browser` YAML (`:387`), reload handling (`:629`), a `-no-browser` flag, and
`PRUEFER_NO_BROWSER` (`:866`). The engine got the dependency on `internal/githubauth`
without adopting its control.

A long-running daemon is a bad place for an unconditional browser-open default: it is
frequently headless, containerized, or on a remote box.

### D2 — the test seam is unreachable from the consumers it exists for

`internal/githubauth/browser.go`:

```go
// openBrowser is a package-level var (not a plain function) so tests can ...
var openBrowser = defaultOpenBrowser
```

It is **unexported**, so only tests *inside* `internal/githubauth` can stub it. But that
package's `doc.go` explicitly designs it for external callers ("callers supply their own
paths/config … the rest of the caller's code depends only on the narrow GitHubAuth
interface"). Any external caller's test that reaches `Reconcile` → `derive` →
`guideMissingInstallations` — via a `WatchedRepos` entry naming an owner with no
installation, and no `AppInstallationID` pin — **opens a real browser**, with
`opts.NoBrowser` (which a test author must know to set explicitly) as the only guard.

This is the confirmed, currently-reproducible path: `engine/github_app_auth_test.go`
already calls `githubauth.Reconcile` directly (line 334), and any such test that doesn't
special-case installation coverage is exposed.

### D3 (superseded — see Verification Notes) — slug/name conflation

The original report of this issue described the guided-install URL as being built from a
creation-time App *name* default (`GitHubAppName = "fabrik"`, `engine/github_app_auth.go`)
rather than the App's real slug, producing a URL that would 404 for every engine
deployment. **Code review during Specify did not confirm this against current `main`** —
see Verification Notes below. R4 is retained as a verify-and-fix (not assumed-broken)
requirement.

## Summary

Give the engine an explicit `NoBrowser` control (mirroring Pruefer's, defaulted to *no
browser* since the engine is a daemon), and make `internal/githubauth`'s browser-opening
seam usable by tests in packages other than `internal/githubauth` itself — so external
callers (the engine today, anyone else in the future) can exercise the reconcile path in
tests without side effects on the developer's desktop, and so a real deployment never pops
an uncontrollable browser window. Additionally, verify and — only if actually
reproducible — fix the slug/name conflation described in the original report, and
reconsider whether/how often guided-install browser-opening should fire from the reconcile
path at all.

## Requirements

- **R1** — Give the engine a `NoBrowser` control: config key, flag, and env var, mirroring
  Pruefer's (`no_browser` / `-no-browser` / `FABRIK_NO_BROWSER`), and plumb it into
  `githubauth.Options.NoBrowser` (that field already exists on `Options` — Pruefer sets
  it; the engine currently does not).
- **R2** — **Default it to no-browser for the engine.** Opening a browser should be
  opt-in for this daemon, unlike Pruefer's first-run setup flow where it is the point.
  Guidance must still be *logged* either way — only the automatic open changes.
- **R3** — Make the browser seam reachable/stubbable from external callers' tests (e.g.
  `engine`'s own test package). Either export a stubbable hook, have `Options` carry the
  opener, or guarantee `NoBrowser` defaults safe when `Options` is zero-valued. A package
  documented for external use must be testable by external users with zero side effects on
  the developer's desktop — this is the fix for the concretely-observed test pollution
  (D2), independent of R4/R5's outcome.
- **R4** — Verify the slug/name conflation described in the original report (D3) against
  current `main`, and fix it only if confirmed. Specify-stage review of
  `internal/githubauth/reconciler.go` found `guideMissingInstallations(opts, slug, set,
  logf)` is called with a `slug` local variable that traces back exclusively to
  `gh.FetchAppSlug(opts.BaseURL, jwt)` / `retryIdentityCheckAfterCreation` (never from
  `opts.AppName`) at every point in its history (`git log -S`), and `Options.AppName` is
  used only by the manifest-creation flow (`bootstrap.go`'s `buildManifest`) — a
  structurally different, unreachable-under-engine-compat-mode path. If Research confirms
  this reading, R4 is satisfied by a regression test locking in correct-slug usage rather
  than a code change; if Research finds an actual conflation (e.g. in an error/fallback
  branch not covered by the trace above), fix it there. If no real slug is resolvable in
  whatever branch is found to be at issue, log the guidance without a fabricated URL rather
  than emitting one that could 404.
- **R5** — Reconsider whether the reconcile path should attempt guided installation at
  all, and how often. Specify-stage review found `guideMissingInstallations` is called
  from exactly one site — inside `Reconcile()`'s non-pinned branch — and that both the
  engine (`New()`, once per process) and Pruefer (`execute.go`, once per daemon startup)
  call `Reconcile()` only once per process lifetime; periodic re-derivation
  (`Reconciler.Derive` / Pruefer's `rederiveRepos`) never calls it. So
  `openedInstallBrowser`'s existing per-call bound is *already* a per-process-lifetime
  bound in steady state — the realistic repeat-trigger is a crash-looping or
  frequently-restarted process, not a live long-running reconcile interval. Research should
  confirm this mechanism precisely (including whether the engine's compat/pinned-only
  config makes this code path unreachable in production entirely today — see Verification
  Notes) before deciding whether R5 needs code changes beyond R1–R3, or whether it's
  already effectively satisfied and only needs documenting.
- **R6** — Regression coverage: a test asserting no browser is opened when the engine
  reconciles with a watched-but-uninstalled owner (non-pinned discovery path). It must be
  written so that a failure does not itself open a browser (i.e. it must use R3's seam, not
  rely solely on `NoBrowser` being set correctly).

## Scope

**In scope**: `internal/githubauth/derive.go`'s `guideMissingInstallations` and its
browser/slug handling, `internal/githubauth/browser.go`'s seam,
`engine/github_app_auth.go`'s Options construction and the new config surface
(`cmd/root.go`, `config/config.go`), docs, tests.

**Out of scope**: Pruefer's existing `no_browser` behaviour — it is correct for a
first-run setup flow and should not change; the manifest-creation browser flow in
`bootstrap.go`, which is legitimately interactive and user-initiated.

## Open Questions

None outstanding that require a human decision — the remaining uncertainties (exact
current reachability of `guideMissingInstallations` under the engine's compat/pinned
config, and whether D3 reproduces at all) are implementation-verification items for
Research, tracked under Verification Notes below, not product/scope decisions.

## Verification Notes for Research (Specify-stage code review, not yet confirmed by Research)

- **Reachability under engine's compat mode**: the engine's App-auth config is
  all-or-nothing (`AppID` + `AppPrivateKeyPath` + `AppInstallationID` all set together —
  `validateGitHubAppConfig`), so `Reconcile()`'s pinned branch
  (`opts.AppInstallationID != 0`) always applies and `return`s (reconciler.go, before the
  non-pinned branch's `guideMissingInstallations` call) — meaning this code path appears
  structurally unreachable from the engine's actual production runtime today. The
  concretely-reproducible trigger is **tests** (in `engine` or any other external package)
  that call `githubauth.Reconcile` directly with `WatchedRepos` and no `AppInstallationID`
  pin — exactly D2's scenario, and exactly what R6 must cover. Please confirm this reading
  before scoping R5's fix, since "the engine's production runtime is not currently
  affected" changes the urgency/shape of R5 relative to R1–R3, which remain needed
  regardless (test hygiene, future non-pinned engine support, and Pruefer's own
  non-pinned/discovery mode are all still exposed).
- **Slug source**: confirmed via `git log -S"guideMissingInstallations(opts"` that the
  call site has never passed anything other than the `gh.FetchAppSlug`-derived `slug`
  variable. Please double-check for any branch/fallback not visible from a static read
  (e.g. a future edit, or an error path that leaves `slug` in a stale/zero state before
  reaching this call) rather than assuming D3 is fully void — but do not assume it's
  broken either.

## Prior Art / Context

- Pruefer's `NoBrowser` surface (`pruefer/config.go:292-296,387,629,719,866`) is the
  direct precedent for R1 — mirror its shape (config/flag/env, `reload:"restart"`
  semantics) rather than inventing a new one.
- `internal/githubauth/doc.go` explicitly documents the package as caller-agnostic and
  designed for external callers to depend only on the narrow `GitHubAuth` interface — this
  is the design intent R3 is restoring for the test/side-effect boundary specifically.
- Same family as #1750, #1752, #1755, #1756: behaviour that differs under App auth and was
  inferred rather than exercised. This one is distinctive in that the defect is in the
  *shared* package (`internal/githubauth`) rather than the engine's use of it, and
  Pruefer's version of the same surface (`NoBrowser`) is the correct one — the engine
  adopted the dependency without adopting its control.

## Risks / Dependencies

- R4/R5's final shape depends on Research's verification of the reachability/slug findings
  above — plan for the possibility that these requirements resolve to "add regression
  coverage confirming existing-correct behavior" rather than "change behavior."
- R3's seam design (exported hook vs. `Options`-carried opener vs. safe zero-value default)
  is an implementation choice for Research/Plan, not prescribed here — any of the three
  satisfies AC3 as long as it requires no caller-side `NoBrowser` knowledge to avoid a real
  browser open in tests.

## Acceptance Criteria

- **AC1** — An engine reconciling under App auth with a watched-but-uninstalled owner opens
  **no** browser window, by default, and logs the guidance instead (already true for the
  engine's current pinned-only production runtime per the Verification Notes above; this
  AC's real teeth are R6's regression test covering the non-pinned/discovery path, which
  the engine's tests and any future non-pinned engine usage do exercise).
- **AC2** — The engine exposes a documented way to re-enable browser-opening if wanted.
- **AC3** — A test in the `engine` package can exercise the reconcile path without opening a
  browser and without needing to know about `NoBrowser`.
- **AC4** — No guided-install URL is emitted from a creation-time name default; either the
  real slug is used or no URL is emitted. (Per R4, confirm this already holds before
  treating it as a code-change requirement.)
- **AC5** — Pruefer's first-run browser behaviour is unchanged.
- **AC6** — `go test -race ./...` passes and opens no browser windows.

## Notes

Found live while three stage workers iterated on App-auth tests — repeated 404 browser
windows were arriving on the operator's desktop during test runs (D2, confirmed). This
blocks comfortable further App-auth test work: any worker touching this path risks firing
tabs until R3/R6 land.

arbeithand:
Closes #1763

## Summary

Fixes #1763's two confirmed defects (D1/D2) around `internal/githubauth`'s App-auth reconcile opening an uncontrollable browser window, and adds regression coverage confirming the two suspected-but-unconfirmed issues (D3 slug/name conflation, R5 call-frequency) don't actually exist in current code.

- **R1/R2** — The engine daemon gets its own `--no-browser` / `FABRIK_NO_BROWSER` / `no_browser` control (`cmd/root.go`, `config/config.go`, `engine.Config.NoBrowser`), wired into `githubauth.Options.NoBrowser`. Defaults to `true` (suppressed) — the inverse of Pruefer's own `no_browser` default — since the engine is a long-running daemon rather than a first-run interactive setup flow. Guidance is still logged either way; only the automatic browser launch is gated.
- **R3/AC3** — `internal/githubauth`'s `Options` gains a new `OpenBrowser func(url string) error` field. `guideMissingInstallations` prefers it over the package-level `openBrowser` var when set, so an external caller's test (e.g. `engine`'s own tests) can exercise the non-pinned discovery reconcile path with zero side effects, without needing to know that `Options.NoBrowser`'s zero value is otherwise unsafe.
- **R4** — Confirmed via regression test (`TestReconcile_GuidedInstallURL_UsesSlugNotAppName`) that the guided-install URL is always built from the live-fetched App slug, never from `Options.AppName` — the originally reported slug/name conflation does not exist in current code. No production code changed for this requirement.
- **R5** — Documented (ADR 1763) rather than code-changed: `guideMissingInstallations`'s per-call browser-open bound is already a per-process-lifetime bound in steady state, since both the engine and Pruefer call `Reconcile()` exactly once per process lifetime.
- **R6/AC1/AC6** — New engine-side regression test (`TestReconcile_NonPinnedDiscovery_NoBrowserOpensViaOptionsSeam`) exercises the non-pinned discovery path directly via `githubauth.Reconcile`, leaving `NoBrowser` at its unsafe zero value and relying solely on the `OpenBrowser` seam to stay side-effect-free.
- **AC2** — Documented in `docs/USER_GUIDE.md`'s new "Browser-opening during reconcile" subsection, distinguishing this daemon-level control from the pre-existing `fabrik init --github-app --no-browser` setup-time flag.
- **AC5** — Pruefer's own `no_browser` surface and default are untouched.

## Key changes

- `internal/githubauth/reconciler.go`, `derive.go` — `Options.OpenBrowser` seam
- `engine/engine.go`, `engine/github_app_auth.go` — `Config.NoBrowser` plumbed into `githubauth.Options`
- `cmd/root.go`, `config/config.go` — `--no-browser` flag / `FABRIK_NO_BROWSER` env / `no_browser` YAML (tri-state, mirrors `TUI`'s pattern), default `true`
- `engine/github_app_auth_test.go` — extends the fake App server with `GET /app/installations` (list) and `GET /installation/repositories`, needed only by the non-pinned discovery path
- `docs/USER_GUIDE.md` (+ regenerated `docs/llms-full.txt`), `CLAUDE.md`, `adrs/1763-githubauth-browser-seam-and-noBrowser-default.md`

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -race ./...` — all packages pass except `TestExecute_ConfigYAMLApplied` (`cmd` package), a pre-existing network-dependent flake confirmed to fail identically on the pre-issue-1763 baseline (hits real github.com, unrelated to this change)
- [x] New regression tests in `internal/githubauth` (R4) and `engine` (R6/AC3) pass, confirmed no browser window opens


arbeithand:
*branch: fabrik/issue-1763 | commit: 504d03c3 | main: 85074f0b8871c564f6240dc9a8d8be640f6f7667 | 2026-09-16 14:51 UTC*

Build confirmed clean. Everything is committed, pushed, tested, and documented. Creating the PR now and completing the stage.


All implementation work (R1–R6), docs, and ADR 1763 were already committed and pushed to `fabrik/issue-1763`; the only outstanding step was PR creation, which is done now. `go build`/`go vet`/`go test -race ./...` pass across the whole repo except one pre-existing, network-dependent flaky test in `cmd` unrelated to this change.

---
Used 4/100 turns, 1.3M input (8 raw + 1.3M cache-read + 48k cache-write) / 2k output tokens.

arbeithand:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-1763 | commit: 5f302339 | main: 85074f0b8871c564f6240dc9a8d8be640f6f7667 | 2026-09-16 14:59 UTC*

Still waiting on the `engine` package's `-race` test run (running ~2+ minutes, expected for this suite's size). I'll pick this back up as soon as the background job notifies completion.

---
Used 79/100 turns, 12.4M input (156 raw + 12.3M cache-read + 166k cache-write) / 26k output tokens. Stage incomplete.

arbeithand:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-1763 | commit: dafecb20 | main: 85074f0b8871c564f6240dc9a8d8be640f6f7667 | 2026-09-16 15:16 UTC*

All review threads resolved, build/vet/tests clean, docs in sync, base branch current. The PR is ready for human review.

## Review Findings

### Fixed
- **Issue**: `internal/githubauth/reconciler.go`'s `Options.OpenBrowser` doc comment incorrectly claimed the field overrides *both* browser-open call sites (`guideMissingInstallations` and `RunManifestFlow`) and works "independently of NoBrowser." Neither is true: `RunManifestFlow` is driven by the separate `ManifestFlowOptions` struct (`bootstrap.go`), which has no `OpenBrowser` field at all and always calls the unexported package-level `openBrowser` var directly — this was explicitly left out of scope per this PR's own ADR 1763 (Decision 3). Also, `OpenBrowser` doesn't bypass the `NoBrowser` gate — `NoBrowser` still decides *whether* a browser-open is attempted at all; `OpenBrowser` only decides *what* runs when it is. A future caller relying on the stale wording could be misled into believing `OpenBrowser` alone makes the manifest-bootstrap self-heal path (`reconciler.go:771,1094`) side-effect-free in tests, and would get a real browser popped there. **Fix**: Corrected the doc comments on both `NoBrowser`/`OpenBrowser` in `reconciler.go`. (Pruefer, the automated reviewer, independently flagged the same defect in an inline review thread — now resolved, since the underlying comment is fixed.)
- **Issue**: `internal/githubauth/doc.go`'s package overview documents each caller-agnostic `Options` extension with an "As of #NNNN, ..." paragraph (see the #1712/#1722 precedents), but the new `Options.OpenBrowser` field (#1763) had no equivalent entry, leaving the package-level doc silent about its own testability seam. **Fix**: Added a paragraph following the file's established convention, scoping the claim correctly (covers `guideMissingInstallations` only, not `RunManifestFlow`).

### Verified
- `go test -race -timeout 5m ./internal/githubauth/... ./engine/... ./config/...` passes, including the new R4 (`TestReconcile_GuidedInstallURL_UsesSlugNotAppName`) and R6 (`TestReconcile_NonPinnedDiscovery_NoBrowserOpensViaOptionsSeam`) regression tests, and the new `cmd` precedence tests (`TestExecute_NoBrowser*`, all 5 passing).
- Two unrelated, pre-existing flakes observed, neither touched by this diff nor a regression:
  - `cmd.TestExecute_ConfigYAMLApplied` — a SIGINT-timing test that does real (network) label-seeding calls; flaky under slow network conditions, `cmd/root_test.go` untouched by this PR.
  - `engine.TestInvokeClaude_FabrikEnvVarsInjected` — fails only when `go test` is run inside a Fabrik-orchestrated worker's own shell (this very Review-stage invocation has ambient `FABRIK_PR=1764` set by the outer engine), because the test doesn't scrub ambient `FABRIK_*` vars before asserting their absence. `engine/claude.go`/its test file are untouched by this PR; this is a self-hosting/meta artifact, not something CI (which has no such ambient vars) would hit.
- `docs/llms-full.txt` regenerated — no drift from the committed bundle.
- Branch is current with `main` (0 behind); no rebase needed.
- All Plan checklist items (R1–R6, docs, ADR 1763) confirmed present and correct in the diff.
- Both automated Pruefer reviews found no other defects.

Reviewed the NoBrowser/OpenBrowser implementation for #1763; found and fixed a doc-comment defect (also independently flagged by Pruefer) where `Options.OpenBrowser`'s comment falsely claimed it covers `RunManifestFlow`, which it doesn't and was never meant to. Resolved the corresponding PR review thread, added a missing package-doc entry for the new seam, and re-verified `go build`/`go vet`/`go test -race` all pass (two unrelated pre-existing environmental flakes noted, not regressions). PR #1764 is ready for human review.

---
Used 22/100 turns, 4.9M input (44 raw + 4.9M cache-read + 46k cache-write) / 12k output tokens.

arbeithand:
*branch: fabrik/issue-1763 | commit: dafecb20 | main: 85074f0b8871c564f6240dc9a8d8be640f6f7667 | 2026-09-16 15:17 UTC*

I've reviewed Pruefer's latest bot finding — no action needed.

## Summary

The comment is a `[Bot Review Finding]` (Pruefer's second automated review, submitted 2026-09-16 15:03). It explicitly states **"No new defects found"** and confirms the fix I already applied in the prior Review-stage turn is correct: *"The `OpenBrowser` seam correctly overrides only the opener while `NoBrowser` still gates whether it's called at all, matching its doc comment."*

Its one remaining note — "the pre-existing open review thread about the `RunManifestFlow`/`OpenBrowser` doc-comment mismatch is unchanged by this diff" — refers to the GitHub review thread's *resolution status* at the time Pruefer's review ran (15:03:04Z), which was before I resolved that thread via `resolveReviewThread` later in that same session. I've re-verified just now: that thread (`PRRT_kwDOR2bN_c6i_CNO` on `internal/githubauth/reconciler.go`) is confirmed `isResolved: true`, and it's the only review thread on the PR — no others have opened since.

Working tree is clean, branch is up to date with origin (fix commit `dafecb20` already pushed). There is nothing actionable in this comment — per the No-Op Contract, I'm making no code changes and no issue body changes.

---
**Threads addressed:**

arbeithand:
*branch: fabrik/issue-1763 | commit: dafecb20 | main: 85074f0b8871c564f6240dc9a8d8be640f6f7667 | 2026-09-16 15:25 UTC*

## Validation Report

### Requirements: 6/6 passed

- [x] **R1** (NoBrowser control): `engine.Config.NoBrowser`, `cmd/root.go`'s `--no-browser` flag, `FABRIK_NO_BROWSER` env, `config/config.go`'s `ProjectConfig.NoBrowser *bool` (`no_browser` YAML key) — all wired end-to-end into `githubauth.Options.NoBrowser` via `setUpGitHubAppAuth`. Verified by reading the diff and by 5 passing precedence tests (`TestExecute_NoBrowser{DefaultTrue,FlagReEnables,EnvReEnables,ConfigYAMLReEnables,FlagBeatsEnvAndConfig}`).
- [x] **R2** (default no-browser): flag default `true`; `TestExecute_NoBrowserDefaultTrue` confirms `cfg.NoBrowser == true` with nothing set. Guidance is still logged regardless (`logf` call in `guideMissingInstallations` is unconditional).
- [x] **R3** (stubbable seam): `Options.OpenBrowser func(url string) error` added to `internal/githubauth`; `guideMissingInstallations` prefers it over the package var when non-nil. Verified by `TestReconcile_NonPinnedDiscovery_NoBrowserOpensViaOptionsSeam` (engine package), which deliberately leaves `NoBrowser` at its unsafe zero value and relies solely on the seam — passes.
- [x] **R4** (slug/name conflation, void): `TestReconcile_GuidedInstallURL_UsesSlugNotAppName` locks in that the guided-install URL always uses the live-fetched slug, never `Options.AppName`, with a distinguishable `AppName` value to avoid a vacuous assertion — passes.
- [x] **R5** (reconcile frequency, documented not changed): covered in ADR 1763 Decision 5 — no code change, matching Research's structural finding.
- [x] **R6** (regression coverage, safe-by-construction): the engine-side seam test above never touches a real browser API regardless of `NoBrowser` correctness.

AC1–AC6 all verified: AC2 documented in USER_GUIDE.md's new "Browser-opening during reconcile" subsection + flag/env reference tables; AC3 satisfied structurally by the `OpenBrowser` field design; AC4 via R4's test; AC5 — Pruefer's own `NoBrowser`/call sites untouched (confirmed via diff, no changes to `pruefer/` or `bootstrap.go`); AC6 — see below.

### Test Suite: PASSED (with 2 known pre-existing, unrelated flakes)

- `go build ./...`, `go vet ./...`: clean.
- `go test -race -timeout 8m ./...`: all packages pass except `engine` and `cmd`, both due to flakes already identified and documented in this PR's own prior Review-stage output, confirmed still unrelated to this diff:
  - `engine.TestInvokeClaude_FabrikEnvVarsInjected` — fails only inside this very Fabrik-orchestrated worker's shell (ambient `FABRIK_TOKEN`/`FABRIK_USER`/etc. are set in this environment; the test asserts their absence). `engine/claude.go`/its test file untouched by this diff.
  - `cmd.TestExecute_ConfigYAMLApplied` — pre-existing SIGINT-timing test making real network calls; `cmd/root_test.go` untouched by this diff.
  - All new tests (R2/R3/R4/R6 regression tests, 5 `--no-browser` precedence tests) pass in isolation and within full-package runs.
- No browser window opened during any test run (verified structurally: all reconcile-path tests either stub `OpenBrowser` or set `NoBrowser: true`).

No changes to shared interfaces beyond additive fields (`Options.OpenBrowser`, `Config.NoBrowser`, `ProjectConfig.NoBrowser`). Pruefer's own `NoBrowser` surface and manifest-bootstrap flow are untouched.

### Documentation

`docs/llms-full.txt` regenerated — no drift. `docs/USER_GUIDE.md`, `CLAUDE.md`, and ADR 1763 all present, consistent with the implementation.

### Branch State

Working tree clean, 0 behind `origin/main`, all commits pushed.

### Verdict: READY TO MERGE

Validation passed. Rebase: skipped-up-to-date. PR mergeable: MERGEABLE, mergeStateStatus: CLEAN. All 6 requirements (R1–R6) and AC1–AC6 verified; full test suite passes except two known pre-existing flakes unrelated to this diff (ambient-env test and a network-timing test, both in untouched files). Docs bundle in sync, no regressions.

---
Used 29/100 turns, 3.6M input (48 raw + 3.4M cache-read + 150k cache-write) / 9k output tokens.

arbeithand:
🏭 **Fabrik merge-train** — Landed via singleton fast path PR #1764. That is this PR: the pinned base was already an ancestor of this head, this PR was mergeable, and its own CI was green and complete, so it was landed directly — no trial branch was assembled and no separate integration PR exists. See ADR-1644.

Issue #1699, opened 2026-09-09
Vendored plugin drift is undetectable: plugin version is static, so a customised .fabrik/plugin can sit 40 commits stale with no warning


verveguy:
## Summary

A vendored `.fabrik/plugin/` can fall arbitrarily far behind the embedded skills with **no signal of any kind**. The plugin version is the only thing that could indicate staleness, and it has been `0.1.0` since 2026-05-06 — unchanged across every release since. So a project that vendors the plugin and customises it (the supported path, given `upgrade` refuses to clobber customisations) has no way to learn it is out of date short of speculatively running `fabrik upgrade --reconcile`.

## What happened

A downstream project sat **40 commits and ~67KB behind** on its vendored skills for two months and only discovered it while investigating repeated stage failures.

Measured at the point of discovery, against `plugin/fabrik-workflows/skills/`:

| skill | vendored | embedded | |
|---|---|---|---|
| `fabrik-validate` | 9,321 | 36,983 | **25%** |
| `fabrik-review` | 10,479 | 24,814 | **42%** |
| `fabrik-implement` | 11,040 | 18,440 | 60% |
| `fabrik-plan` | 13,617 | 19,999 | 68% |
| **total (12 skills)** | **97,794** | **165,291** | **59%** |

Throughout, `.fabrik/warnings.json` was `{"version":1,"entries":[]}`, `fabrik refresh-stages` reported nothing (the stage YAMLs genuinely were current), and both `plugin.json` files read `"version": "0.1.0"`.

## Why it mattered

The vendored copy predated 2026-08-08's `feat(skills): add standalone never-wait-on-CI/background imperative to 8 skills`. Without it, a Validate stage waited in-session on a ~21-minute local computation, ended without emitting `FABRIK_STAGE_COMPLETE`, and burned all three retries — each one re-deriving the identical stall, exactly as the new text predicts:

> "'wait for a completion notification' is never a valid terminal strategy in a headless stage ... because the reasoning is deterministic, every retry re-derives the identical stall."

The engine behaved correctly. The skill guidance that would have prevented it existed upstream for a month and simply never arrived.

## Why the existing mechanisms don't cover it

- **`plugin.json` version is static.** Its only commits are `0b5ac2ef` (2026-05-06, creation) and `1af3aa03` (2026-08-15, attribution). Any version-comparison check is a no-op by construction.
[...]

PR #1808, opened 2026-09-19
feat: report plugin staleness independently of customization

### LINKED ISSUE #1787: vendored plugin drift is invisible: customisation short-circuits the staleness signal, and the version never moves

arbeithand:
Closes #1787

Before this change, a customised `.fabrik/plugin/` never told the operator it was also behind the embedded skills. `checkPluginState` returned as soon as it saw the customisation, so staleness was never computed. This PR splits the two into independent facts and reports both. #1699 is the community report that motivated it, mentioned for context only.

**Key changes**
- `plugin/refresh.go`: `checkPluginState`/`CheckPluginState` gain a third return value, `stale`. It is true when `installedVer` is non-empty and differs from `embeddedVer`, and it is computed independent of the customisation and known-version trust logic. The existing `customWorkflow`/`upgradeNeeded` branching is unchanged, so `stale` never affects whether an upgrade auto-refreshes.
- `plugin/known_embedded_versions.go`: new `VersionsBehind`, an optional "N known releases behind" count. It is omitted when the installed fingerprint isn't a known entry.
- `cmd/root.go`: a new `evaluatePluginStartupState` helper is shared by the daemon-startup and re-exec paths, and its logic is unit-testable. `pluginCustomizationWarning` now also names the drifted files and `fabrik upgrade --reconcile` when the plugin is stale.
- `cmd/upgrade.go`: the `fabrik upgrade` refusal and the TTY and non-TTY messages add a staleness note. Nothing about the refusal itself changes.
- `tui/`: the header shows a combined badge, `[u] custom workflow (N stale)`. `SkillsStaleEvent` and `CustomWorkflowEvent` no longer clear each other's field, and the `u` dialog names outstanding staleness.
- `docs/USER_GUIDE.md` (and the regenerated `docs/llms-full.txt`): new combined-state row in the upgrade-protection table, updated TUI descriptions, and a statement that `plugin/fabrik-workflows/.claude-plugin/plugin.json`'s `version` is not consulted by any staleness logic.
- `adrs/1787-plugin-staleness-independent-of-customization.md`: records these decisions.

**Testing**
- New tests cover the customised-and-stale state in `plugin/refresh_test.go`, `cmd/plugin_startup_test.go`, `cmd/upgrade_test.go`, `tui/header_test.go` and `tui/model_test.go`.
- Existing tests that asserted the old mutual exclusivity were rewritten.
- Run `go test -race ./...` and `go vet ./...`. `TestExecute_ConfigYAMLApplied` in `cmd` times out on SIGINT; it fails the same way on a clean `main` checkout.


arbeithand:
*branch: fabrik/issue-1787 | commit: 7e6b16d8 | main: befc86c4c67f88c3ef39883aafd1315c817e8cf0 | 2026-09-19 05:12 UTC*

All #1787 checklist items were already implemented in earlier attempts, so this run only verified them, and the rebased branch is now pushed (`7e6b16d8`).

Checks:
- `go build` and `go vet` are clean, and `docs/llms-full.txt` is in sync with `docs/USER_GUIDE.md`.
- `go test -race ./...` passes except `TestExecute_ConfigYAMLApplied` in `cmd`, which times out waiting for SIGINT. That test fails the same way on a clean `origin/main` checkout, so it predates this work.
- I didn't check whether the new tests fail against `main`, so AC1's red-first check was not re-done here.
- The branch was already rebased on `main`. The remote still held the old SHAs, so I updated it with `--force-with-lease`.
- I skipped checking off tasks in the Plan comment, because the `gh` lookup for it returned nothing.

Customisation and staleness are now reported as independent facts. `checkPluginState` returns a third `stale` value, and the startup warning, `fabrik upgrade` refusal, TUI badge and `u` dialog all show both signals and point at `fabrik upgrade --reconcile`. The refusal to overwrite customised plugins is unchanged. I also added `VersionsBehind` for the optional "N versions behind" count, documented in `docs/USER_GUIDE.md` that `plugin.json`'s `version` is not consulted, and added ADR 1787. Build, vet and the race test run pass, except `TestExecute_ConfigYAMLApplied`, which times out on SIGINT and fails the same way on clean `main`.

---
Used 13/100 turns, 1.2M input (22 raw + 1.1M cache-read + 119k cache-write) / 3k output tokens.

arbeithand:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-1787 | commit: 7e6b16d8 | main: befc86c4c67f88c3ef39883aafd1315c817e8cf0 | 2026-09-19 05:21 UTC*

## Review Findings

### Fixed
- Nothing needed fixing. The diff (15 files) matches the Plan checklist and the issue's R1–R5, and I made no commits in this review.

### Verified
- **Branch state:** the branch was already up to date with `main` (0 commits behind), so I skipped the rebase. The PR base is `main`, and everything is already pushed.
- **R1 (independent facts):** `checkPluginState` now returns a third `stale` value, computed once as `installedVer != "" && installedVer != embeddedVer`. The customised branch returns it instead of skipping the comparison. The startup warning, the `fabrik upgrade` refusal, the TTY and non-TTY messages, the TUI badge and the `u` dialog all show both signals. `evaluatePluginStartupState` replaces the two duplicated inline call sites in `cmd/root.go`, so they can't diverge.
- **R2:** file naming reuses `diffingPluginFiles`. `VersionsBehind` adds the "N known release(s) behind" count and omits it when `installedVer` isn't a known entry.
- **R3:** `docs/USER_GUIDE.md` now states that `plugin.json`'s `version` is not consulted, and adds a row for the combined state. ADR 1787 records the decision.
- **R4:** the startup warning, the refusal message and the TTY/non-TTY messages name `fabrik upgrade --reconcile`. The TUI `u` dialog offers "Reconcile via Claude Code" and now names the outstanding stale-file count, but doesn't quote the CLI command, to keep the status line short. I judged that acceptable.
- **R5:** no auto-refresh decision reads `stale`. `runUpgrade` still refuses whenever `customWorkflow` is true, and the auto-refresh path for a pristine stale plugin is unchanged.
- **Non-vacuous tests (AC6):** I neutralised the fix by making the customised branch return `stale=false`. Five new tests then failed: `TestCheckPluginState_CustomizedAndStale`, `TestEvaluatePluginStartupState_CustomizedAndStale`, the non-TTY and TTY `checkPluginSkillsWithReader` variants, and `TestRunUpgrade_CustomWorkflowAndStaleError`. I restored the file afterwards; the tree is clean.
- **Build and checks:** `go build` and `go vet` are clean, and `docs/llms-full.txt` regenerates with no diff.
- **Tests:** `go test -race ./...` passes across 31 packages (2969 tests) except `cmd.TestExecute_ConfigYAMLApplied` ("Execute did not return after SIGINT"). I ran that test on a clean `origin/main` worktree and it fails the same way, so it predates this PR and is unrelated.

### Notes (non-blocking)
- When the disk already matches the embedded plugin but `installedVer` differs, `stale` is true but the differing-file count is 0. The staleness note is then omitted, which is correct because there is no file drift to report.
- In `runUpgrade`, if `diffingPluginFiles` errors while `stale` is true, the staleness addendum is silently dropped. The refusal itself is unaffected.

Reviewed #1787: `checkPluginState` now returns an independent `stale` fact, and the startup warning, `fabrik upgrade` refusal, TUI badge and `u` dialog show both customisation and staleness, pointing at `fabrik upgrade --reconcile`. No fixes were needed. Neutralising the fix turned five new tests red. `go build` and `go vet` are clean, and `go test -race ./...` passes except `TestExecute_ConfigYAMLApplied`, which fails identically on clean `main`. The branch was already up to date with `main`, and the PR is ready for human review.

---
Used 11/100 turns, 1.4M input (22 raw + 1.2M cache-read + 132k cache-write) / 4k output tokens.

arbeithand:
*branch: fabrik/issue-1787 | commit: 4daabf74 | main: befc86c4c67f88c3ef39883aafd1315c817e8cf0 | 2026-09-19 05:23 UTC*

I fixed the Pruefer finding about `evaluatePluginStartupState`, declined the second edge case it raised, and resolved the review thread. The fix is committed and pushed as `4daabf74`.

**Fixed: a failed file diff no longer drops the plugin state** (`cmd/root.go`)
- **Problem:** `evaluatePluginStartupState` returned an error when `diffingPluginFiles` failed. That happened even though `CheckPluginState` had succeeded and the file count is only a reporting figure. In `handleReexecPluginRefresh` this skipped the `upgradeNeeded` auto-refresh. In `Execute` it dropped `customWorkflow`, so the TUI badge disappeared. Both cut against the promise that staleness never changes refresh behaviour (R5).
- **Fix:** a diff failure now writes an `[upgrade] warning: plugin skill check failed: …` line to stderr and reports `skillsStaleCount = 0`. `customWorkflow` and `upgradeNeeded` are kept as `CheckPluginState` returned them.
- **Test:** `TestEvaluatePluginStartupState_DiffFailurePreservesState` has a pristine-stale case and a customised-and-stale case. It forces the diff failure by putting a directory where an embedded `SKILL.md` should be. With the old error-return restored, both cases fail; with the fix, both pass.
- **Checks:** `go build` and `go vet` are clean. `go test -race ./cmd/` fails only `TestExecute_ConfigYAMLApplied`, which I showed earlier fails the same way on clean `main`.

**Declined: report staleness when the file diff is empty**
- **Situation:** the disk was hand-synced to match the embedded content, but `.installed-version` is old. Then `stale` is true, `customWorkflow` is true, and the diff is empty, so the count is 0 and nothing is shown.
- **Why no change:** the plugin content is not behind, so there is no drift to report, and a "0 files stale" badge would be a false alarm. The file count is the disk-vs-embedded comparison the issue specifies (R2). The gap this issue fixes is real content drift being hidden behind the customisation signal. The only stale item in this edge case is the fingerprint file. Running `fabrik upgrade --reconcile` or `--force` refreshes the fingerprint anyway.
- **Security:** the finding is not security-relevant.

No issue body changes are needed.

---
**Threads addressed:**
- `cmd/root.go:1119` — resolved

arbeithand:
*branch: fabrik/issue-1787 | commit: 4daabf74 | main: befc86c4c67f88c3ef39883aafd1315c817e8cf0 | 2026-09-19 06:06 UTC*

## Validation Report

### Requirements: 6/6 passed (verified against the issue spec, R1–R5 and AC1–AC6)
- [x] **R1, independent facts:** `checkPluginState` returns a third `stale` value, and `evaluatePluginStartupState` replaces the two duplicated inline call sites in `cmd/root.go`. The `fabrik upgrade` refusal, the TTY and non-TTY messages, the TUI badge and the `u` dialog all show both signals.
- [x] **R2, names what drifted:** file naming reuses `diffingPluginFiles`. `VersionsBehind` adds the "N releases behind" count and omits it when the installed fingerprint isn't a known entry.
- [x] **R3, version policy:** `docs/USER_GUIDE.md` states that `plugin.json`'s `version` is not consulted and adds a combined-state table row. ADR 1787 records the decision.
- [x] **R4, points at `fabrik upgrade --reconcile`:** the startup warning, the refusal and the TTY/non-TTY messages name it. The TUI `u` dialog names the stale count but doesn't quote the command. Review judged that acceptable, and I agree.
- [x] **R5, refusal not weakened:** no auto-refresh decision reads `stale`, and `runUpgrade` still refuses whenever `customWorkflow` is true.
- [x] **AC6, non-vacuous tests:** Review neutralised the fix and five new tests went red. I did not repeat that check here.

### Test Suite: `go build` and `go vet` clean; tests pass except three failures that touch none of this PR's code
- **Full run:** `go test -race ./...` ran 3299 passed, 2 failed, 1 skipped across 31 packages.
- **`cmd.TestExecute_ConfigYAMLApplied`:** "Execute did not return after SIGINT". Review reported it fails identically on clean `main`. I did not re-verify that here.
- **`tests/sim` failures:** the failing tests changed on every run:
  - Run 1: `TestFailureShape_UsageLimitExit`.
  - Run 2: 5 merge-train tests.
  - Run 3: `TestCruiseFullPipeline_NonVacuous` and `TestMergeTrainRunaway_DoesNotTripWithGenerousWindow`.
- **Cause of the sim failures:** two of the failures came from this machine's environment:
  - `git config` died with `signal: segmentation fault` in test setup (Apple Git 2.54.0).
  - A worker exceeded the 1m quiescence timeout under full-suite load.
- **Isolation rerun:** `TestFailureShape_UsageLimitExit` passed 3 of 3 when run alone, and the two merge-train tests I ran together passed. I did not rerun the other failures alone.
- **Scope:** the sim package doesn't exercise plugin-state, `cmd/upgrade` or TUI code.
- **CI:** the required check "Test and vet" is `pass` on the PR.
- **Not done:** I skipped a same-conditions rerun of the sim package on clean `main`, because the worktree approach was declined.

The diff is 15 files, all within scope: plugin, cmd, tui, docs, ADR and tests.

### Branch state
- The branch is clean and at `4daabf74`, matching `origin/fabrik/issue-1787`.
- It is 0 commits behind `main`.
- The dependency-install step was skipped, because `CLAUDE.md` specifies no install command for this Go project.

### Verdict: READY TO MERGE

Rebase: skipped-up-to-date (0 commits behind `main`). PR mergeable: MERGEABLE, mergeStateStatus: CLEAN, not in a merge queue. Required checks: Test and vet: pass, Verify llms-full.txt is up to date: pass, Analyze (go): pass (listed twice). Requirements R1–R5 and AC1–AC6 were verified against the issue spec. `go build` and `go vet` are clean. Locally, `go test -race ./...` ran 3299 passed and 2 failed: `cmd.TestExecute_ConfigYAMLApplied` (SIGINT timeout; Review reported it fails identically on `main`, not re-verified here) and `tests/sim` tests whose failures changed between runs (`git` segfault in setup, and a worker timeout under load) and passed when rerun in isolation. None of those touch this PR's code. The engine's CI gate decides whether the PR advances.

---
Used 5/100 turns, 1.9M input (28 raw + 1.8M cache-read + 165k cache-write) / 4k output tokens.

arbeithand:
*branch: fabrik/issue-1787 | commit: 4daabf74 | main: befc86c4c67f88c3ef39883aafd1315c817e8cf0 | 2026-09-19 06:06 UTC*

I made no code changes for this finding and resolved the review thread.

Review already looked at this edge case and declined it. Pruefer raised it again on a new thread (`cmd/root.go:1126`).

**The edge case:** the disk has been hand-synced to match the embedded plugin, but `.installed-version` is old. Then `stale` is true, the file diff is empty, and `skillsStaleCount` is 0. The staleness badge and warning suffix stay silent.

**Why I declined it again:**
- **Nothing is behind:** the plugin files already match the embedded ones. The only outdated item is the fingerprint file, so a "0 files stale" notice would be a false alarm.
- **It matches the spec:** R2 defines the staleness signal as the disk-vs-embedded file diff, plus the optional releases-behind count. The gap this issue closes is real file drift hidden behind the customisation warning, and that is now reported.
- **Suggested fixes:**
  - `max(len(diffing), 1)` would report "1 file(s) differ" when nothing differs.
  - Returning `stale` as a new field would change the function signature and add a message variant for the case the spec did not ask for.
- **Refresh is unaffected:** `fabrik upgrade --reconcile` or `--force` rewrites the fingerprint file. Refresh decisions never read `stale`, so R5 is unaffected.
- **Not security-relevant:** the finding concerns whether a status notice is shown, not credentials or permissions.

The diff-failure fallback to a count of 0, which Pruefer also mentioned, already logs `[upgrade] warning: plugin skill check failed: …` to stderr. `customWorkflow` and `upgradeNeeded` are kept, and the earlier fix in `4daabf74` covers it.

The tree is clean and `origin/fabrik/issue-1787` is unchanged at `4daabf74`. I did not re-run the test suite, because nothing changed since the Validate run. No issue body changes are needed.

---
**Threads addressed:**
- `cmd/root.go:1126` — resolved

arbeithand:
🏭 **Fabrik merge-train** — Landed via singleton fast path PR #1808. That is this PR: the pinned base was already an ancestor of this head, this PR was mergeable, and its own CI was green and complete, so it was landed directly — no trial branch was assembled and no separate integration PR exists. See ADR-1644.

PR #1816, opened 2026-09-20
fix(engine): resume a paused issue only on a human comment that postdates the pause

### LINKED ISSUE #1813: the paused-resume gate re-fires on the same unprocessed comment every poll, so a circuit-breaker pause can never hold
## Problem

A circuit-breaker pause cannot hold when the thing that tripped it is an unprocessable comment,
because that same comment immediately lifts the pause. The breaker then re-arms from zero, trips
again, lifts again, and loops indefinitely.

**Issue #1752, 2026-09-16.** GitHub's own timeline, during the loop:

```
13:36:46  comment   🏭 Fabrik — comment-processing circuit breaker tripped
13:36:47  labeled   fabrik:paused
13:36:48  labeled   fabrik:awaiting-input
13:36:49  unlabeled fabrik:paused          ← 2 seconds later
13:36:50  unlabeled fabrik:awaiting-input
13:36:52  labeled   fabrik:editing         ← straight back to work
...
13:38:05  comment   🏭 Fabrik — no-op comment-processing circuit breaker tripped
13:38:06  labeled   fabrik:paused
13:38:12  unlabeled fabrik:paused
...
```

The breaker tripped and was overridden **ten times** between 13:36 and 13:46, each time reporting
"10 consecutive" — because the unpause path calls `clearFailedStage`, which zeroes the counter. In
total #1752 burned **139** Validate comment-review invocations in 20 continuous minutes. 86ed's #574
did 93 in the same hour. Neither breaker bound held for a single cycle.

**The mechanism.** `processItem` (`engine/item.go:525`) treats *any* unprocessed human comment on a
paused issue as an implicit "resume and handle this":

```go
if len(filterHuman(e.findNewComments(item))) > 0 {
    e.logf(item.Number, "unpause", "user commented on paused issue — unpausing\n")
    e.removeLabel(item, "fabrik:paused")
    e.clearFailedStage(item, stage)   // zeroes the breaker counter
    break
}
```

`findNewComments` defines "new" as *not yet marked processed* — no `CommentProcessed` record, no 🚀
reaction. A comment only becomes processed when an invocation successfully processes it. So when
processing can **never** succeed, the comment is permanently "new", and the resume gate re-fires on
it every single poll, forever.

On #1752 the comment in question is a genuine one — `arbeithand`, 12:12:30Z, *"## D2's blast radius
is wider than filed…"*. It is not Fabrik output (it does not carry the `🏭 **Fabrik` body prefix
that `findNewComments` filters on), and `arbeithand` is a plain GitHub user account, so
`IsBotLogin` correctly returns false. Nothing here is misclassified. The comment deserved to be
processed and deserved to lift the pause **once**. The defect is that it lifts it every time.

The trigger on #1752 was a Claude session limit misreported as a transient `api_error` (sibling
issue), which made every invocation fail at turn 1. But the loop is **not specific to that cause**:
any condition under which comment processing cannot complete — a poisoned session, a persistently
failing worktree, a comment that reliably crashes the stage — reaches the same state. The pause
being defeated is the general defect; the 429 was just what exposed it.

This also means **ADR-069's "honorable pause" guarantee does not currently hold** for any
engine-applied pause on an issue carrying an unprocessable human comment. The operator kill-switch
is real for a clean issue and illusory for exactly the issues most likely to need it.

## Summary

The paused-resume gate authorises a resume from a *state* ("an unprocessed human comment exists")
rather than an *event* ("a human commented since the pause"). Those are the same thing only while
comment processing succeeds. This issue changes the gate to require a comment that genuinely
postdates the pause, so one comment authorises at most one resume, and a breaker pause survives a
comment that can never be processed.

## Requirements

**R1 — One comment authorises at most one resume.** A paused issue must not be resumed twice by the
same comment. The resume decision must be anchored to something that advances when the resume
happens, not to a processed-marker that only advances when processing succeeds.

**R2 — A comment arriving after the pause still resumes, promptly.** This is the whole point of the
gate (#1083) and must not regress. A human who comments on a paused issue gets a resume on the next
poll, exactly as today, including when an older unprocessable comment is also present.

**R3 — Durable across restarts.** The anchor must survive an engine restart; `itemstate.Store` is
in-memory only. Use GitHub-resident state — the pause's own timestamp, a marker on the pause
comment, or a label — not process memory. A restart must not re-authorise a resume the pre-restart
engine already spent.

**R4 — Apply to `fabrik:awaiting-input` too.** The `isAwaitingInput` branch immediately above has
the identical `filterHuman(findNewComments)` shape and the identical defect; both labels were
applied and both removed within four seconds on #1752. Fix them together or the loop simply moves.

**R5 — A backlog is still handed over whole.** Today, once a resume is authorised, the *full* raw
comment set is passed to `processComments` in the same pass so no bot-comment backlog is stranded.
That behaviour must be preserved; only the *authorisation* changes, not what gets processed.

**R6 — Do not make the pause unliftable.** A human must always have a way back. If the anchor is
ever indeterminate, resume — the failure mode to avoid is a permanently stuck issue, not one extra
resume. State the chosen tie-break explicitly in the ADR.

**R7 — Breaker counters must not be zeroed by a resume the breaker itself provoked.** `clearFailedStage`
zeroing the comment-breaker counter is correct for a genuine human resume and is what let #1752
re-report "10 consecutive" ten times over. With R1 in place this should follow automatically;
verify it explicitly rather than assuming it.

**R8 — Admission and resume must agree.** `itemNeedsWork` (`engine/item.go` ~L363) carries a third
copy of the same `filterHuman(findNewComments)` test for paused items, deciding whether `processItem`
is dispatched at all. It must use the same authorisation predicate as the two resume gates in
`processItem`; otherwise the admission check would keep dispatching an item whose resume is then
refused (or vice versa). One shared predicate, three callers.

## Scope

**In scope:** `engine/item.go`'s paused and `isAwaitingInput` resume gates and the matching
paused-admission check in `itemNeedsWork`, whatever durable anchor R3 requires,
`engine/mutate.go`'s `pauseIssue` if it must record the anchor, `docs/state-machine.md`'s
pause/resume and comment-processing lifecycle sections (and the regenerated `docs/llms-full.txt`),
`CLAUDE.md`'s `fabrik:paused` / `fabrik:awaiting-input` semantics, tests, and an ADR numbered
`adrs/1813-*.md` recording the anchor choice and the R6 tie-break.

**Out of scope:** the 429 misclassification and the review-cycle accounting (sibling issues);
`filterHuman` / `IsBotLogin`'s definition of a bot — it is behaving correctly here; the breakers'
own thresholds; adding backoff to comment-processing dispatch.

## Acceptance

- A paused issue carrying one unprocessable human comment that predates the pause is **not** resumed,
  on any number of consecutive polls (repeat-comment case), and its breaker counter is not zeroed.
- The same issue **is** resumed on the next poll once a new human comment arrives after the pause;
  the full raw comment set is passed to `processComments` (R5).
- The same two behaviours hold for `fabrik:awaiting-input`.
- Behaviour is unchanged across an engine restart (no in-memory state consulted).
- An indeterminate anchor resumes (R6), covered by a test.
- Existing human-only resume tests (`blocked_on_input_test.go`) remain green and are extended with the repeat-comment case.

## Deferred to Research (non-blocking)

**Q1 — What is the durable anchor?** Candidates: (a) compare comment `created_at` against the pause
comment's `created_at`; (b) 🚀-react the triggering comment at pause time so it stops being "new";
(c) a watermark label. (a) is the most direct expression of "arrived after the pause" and needs no
new durable state, but relies on the pause comment being findable — there is precedent for that scan
in `hasCIGatePauseComment`. (b) is cheapest but lies about the comment's processing state and would
suppress it from a later legitimate invocation. The requirements above do not depend on which is
chosen; Research should pick one, weighing (a)'s handling of pauses that post no comment (e.g. an
operator-applied `fabrik:paused`, where the label-applied timestamp is the only anchor) against R6.

## Prior Art / Context

- **#1083 / ADR-069** established that bot chatter must not defeat an operator pause, and introduced
  `filterHuman` at exactly these two sites. This issue continues that line of reasoning: #1083
  closed the "wrong *author*" hole; this closes the "same comment, over and over" hole.
- **`hasCIGatePauseComment`** (`engine/ci.go`) already scans an issue's comment history for a
  Fabrik pause comment to make a gate decision idempotent — precedent for anchor option (a),
  including its unscoped-by-time scan.
- **`findNewComments`'s `🏭 **Fabrik` body-prefix filter** is why Fabrik's *own* pause comment does
  not itself trigger the resume. That filter is load-bearing and must not be disturbed — note that
  any engine-authored comment not carrying the prefix would become a self-unpause trigger.
- **#1575** (*Two of four resume paths clear the cycle counters but not #1089's comment breaker*)
  is adjacent but distinct: it is about *which state a resume clears*, this is about *whether the
  resume should have been authorised at all*. Neither fixes the other. Worth reading together —
  #1575's table of the four resume paths is the map this issue's R4 needs.
- **`settleBotServiceNotices`** solves the adjacent "a comment that is never processed keeps
  re-admitting" problem for bot service notices by watermarking them. The same shape may inform R3.

## Risks / Dependencies

- **The failure mode to fear is a pause nobody can lift.** R6 is the governing constraint; when in
  doubt, resume. An extra resume is a wasted cycle, a stuck issue is an operator emergency.
- **Regressing #1083 would be silent.** The existing human-only resume tests (`blocked_on_input_test.go`)
  must stay green and be extended with the repeat-comment case.
- **Three call sites, one predicate.** Fixing only the `processItem` gates while leaving
  `itemNeedsWork` unchanged (R8) would produce admission/resume disagreement.
- **Independent of the sibling issues** and may land in any order. The 429 fix makes this loop far
  rarer; it does not make the pause hold.

## Notes

139 invocations at roughly 8 seconds apart is what an unbounded loop looks like from the outside: no
error, no alert, just a daemon quietly spending an account's quota on an issue it had already
decided to stop working on.

The ADR for this issue is numbered after it, per `CLAUDE.md`'s ADR-numbering convention.

arbeithand:
Closes #1813

A paused or awaiting-input issue was resumed by any unprocessed human comment. A comment that can never be processed stayed "new" forever, so it lifted the pause every poll. Each resume zeroed the comment-breaker counters, which let #1752 trip and override its breaker ten times in ten minutes.

## What changed

- **One shared predicate**, `resumeAuthorised` in `engine/comments.go`. It replaces the four inline `filterHuman(findNewComments)` tests: both `itemNeedsWork` branches and both `processItem` gates (R4, R8). The dead `humanNewComments` is removed.
- **A resume now needs an event, not a state.** At least one human comment must have `CreatedAt` at or after the latest `fabrik:paused` `labeled` event. That timestamp comes from `FetchLabelAppliedAt` and lives in GitHub's event log, so it survives restarts (R3).
- **The anchor is read live, not from the record-on-write cache.** A label a human removes in the UI would otherwise leave a stale early timestamp that a later pause inherits.
- **R6 tie-break: every indeterminate input resumes.** That covers a fetch error, no event found, a zero `CreatedAt`, or an equal timestamp. Only "every human comment strictly predates the pause" refuses.
- **R5 is unchanged.** An authorised resume still passes the full raw `findNewComments` set to `processComments`.
- **R7 needed no reset changes.** The counter resets simply stop running when the gate refuses; tests assert the counters are untouched.
- Refusals log a distinct "N human comment(s) predate the pause" line.

Trade-off, recorded in the ADR: a human comment posted *during* an invocation that then pauses predates that pause and does not resume it. The human comments again.

## Docs

- ADR 1813 (new), with a supplement pointer added to ADR-069.
- `docs/state-machine.md` (new "Resume anchor" paragraph, plus the label table, transition rows, guard tables).
- `docs/USER_GUIDE.md`, `CLAUDE.md` (new `fabrik:paused` / `fabrik:awaiting-input` bullet), and the regenerated `docs/llms-full.txt`.

## Testing

- `engine/resume_anchor_test.go` covers:
  - the comparison table;
  - the predicate, including R6 and a stale-cache bypass;
  - the repeat-comment case over 5 polls, for both labels, at admission and in `processItem` (R1, R4, R8);
  - breaker counters untouched (R7);
  - a post-pause comment resuming with the full raw set (R2, R5);
  - a fresh engine making the same decision (R3);
  - an indeterminate anchor resuming (R6).
- Sim scenario `tests/sim/pause_resume_anchor_test.go`: the pause holds across 6 polls against a pre-pause comment, then a post-pause comment resumes it. It fails with the predicate neutralised and passes with the fix.
- The `humanNewComments` tests were retargeted at `filterHuman(findNewComments)`. I did not extend `blocked_on_input_test.go` with repeat-comment cases; they live in the new file, and the existing tests there pass unchanged.
- `go build`, `go vet` and `go test -race ./...`: 3801 passed, 1 failed. The failure is `cmd`'s `TestExecute_ConfigYAMLApplied` ("Execute did not return after SIGINT"). It fails identically on the base commit 9289bb44, so it isn't caused by this change.


arbeithand:
*branch: fabrik/issue-1813 | commit: 1746cb49 | main: 9289bb44fa363acf601d845276184b25f8636125 | 2026-09-20 20:16 UTC*

`cmd`'s `TestExecute_ConfigYAMLApplied` fails the same way on the base commit, so it isn't caused by my change; it appears to be environmental (sandbox network/signal handling). Everything else passes.


Added one shared predicate, `resumeAuthorised`, so a paused or awaiting-input issue resumes only when a human comment arrives at or after the latest `fabrik:paused` labeled event. It backs all four gates, including `itemNeedsWork`, and an indeterminate anchor still resumes (R6). Tests cover the predicate, the gates, breaker counters, restart behaviour and a sim scenario, and ADR 1813, the state-machine docs, CLAUDE.md, USER_GUIDE and `llms-full.txt` are updated. `go test -race ./...` has one failure, `cmd`'s `TestExecute_ConfigYAMLApplied` (SIGINT), which fails identically on the base commit.

---
Used 35/100 turns, 4.7M input (68 raw + 4.5M cache-read + 155k cache-write) / 22k output tokens.

arbeithand:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-1813 | commit: 1746cb49 | main: 9289bb44fa363acf601d845276184b25f8636125 | 2026-09-20 20:26 UTC*

## Review Findings

I found nothing to change, so this review made no commits. The branch is already pushed and the worktree is clean.

### Verified
- **Rebase:** the branch is 0 commits behind `origin/main`, so I skipped the rebase. The PR's base is `main`.
- **External feedback:** the PR has no review-bot or human comments to address.
- **Spec coverage:**
  - **R1–R4:** one shared predicate, `resumeAuthorised`, replaces all four gates. Those are the awaiting-input and paused branches in both `itemNeedsWork` and `processItem`. The anchor is the latest `fabrik:paused` `labeled` event, read directly from `client.FetchLabelAppliedAt` and never from the record-on-write cache. That is deliberate: a human removing the label in the GitHub UI would leave a stale early timestamp in the cache and re-arm the loop.
  - **R5:** the full raw `findNewComments` set is still what `processComments` receives.
  - **R6:** a fetch error, a zero anchor, a zero `CreatedAt` or an equal timestamp all resume. Only "every human comment strictly predates the pause" refuses. `commentsPredatePause` implements this, and the tests cover it.
  - **R7:** no code change was needed. `clearFailedStage`, `unblockAwaitingInput` and the breaker resets simply stop running when the gate refuses, and tests assert the counters stay untouched.
  - **R8:** `itemNeedsWork` and `processItem` share the one predicate, so admission and resume cannot disagree.
- **Dead code:** the unused `humanNewComments` was removed, and its three tests were retargeted at `filterHuman(findNewComments(...))`.
- **Tests:**
  - **Unit:** `engine/resume_anchor_test.go` covers the predicate and the gates for paused alone and for paused plus awaiting-input.
  - **Sim:** `tests/sim/pause_resume_anchor_test.go` runs the repeat-comment case end to end. A pre-pause comment holds the pause across 6 polls with zero comment-review calls, and a post-pause comment then resumes it.
  - **Targeted run:** 285 pause/resume/breaker tests pass under `-race`.
- **Docs:** ADR 1813, `docs/state-machine.md`, `CLAUDE.md` and `docs/USER_GUIDE.md` are updated. Regenerating `docs/llms-full.txt` produces no diff.
- **Build:** `go build ./...` and `go vet ./...` are clean.

### Full `go test -race ./...`: 4 failures, none from this change
- **`engine` (2):** `TestInvokeClaude_FabrikEnvVarsInjected` and its `FABRIK_PR_absent` subtest fail identically on a clean `origin/main` checkout. The test trips on the host environment (`NVM_INC` and similar variables).
- **`cmd` (1):** `TestExecute_ConfigYAMLApplied` (no return after SIGINT). The Implement stage reported it fails the same way on the base commit. I did not re-run it on main.
- **`tests/sim` (1):** `TestMergeTrainMode_On` failed in the full run on a git-config step. It passes in isolation on this branch and on main, so I read it as load-related flakiness.

### Accepted trade-off (recorded in the ADR)
A human comment posted while an invocation is running predates the pause that same invocation applies, such as a breaker trip or `blockOnInput`. It does not resume that pause, so the human has to comment again.

Reviewed the post-pause resume anchor. The single `resumeAuthorised` predicate covers all four gates, and every unknown input resumes, so a pause can still be lifted. I found nothing to fix and made no commits. The branch is up to date with main, and the 285 pause/resume/breaker tests pass under `-race`. Docs, ADR and `llms-full.txt` are consistent. Four failures in the full run are unrelated: three fail on a clean `origin/main` or reproduce there, and one sim test is flaky under load and passes in isolation. PR is ready for human review.

---
Used 12/100 turns, 1.1M input (20 raw + 957k cache-read + 121k cache-write) / 3k output tokens.

arbeithand:
*branch: fabrik/issue-1813 | commit: 1746cb49 | main: 9289bb44fa363acf601d845276184b25f8636125 | 2026-09-20 20:27 UTC*

I made no code changes for the Pruefer polling-cost finding and resolved its thread. It is low severity, and the only fix on offer would undo the safeguard the PR is built around. The working tree is clean and nothing was pushed.

**Why no change**
- **Call count is smaller than stated.** `processItem` is only dispatched when `itemNeedsWork` returns true (`engine/poll.go:1917`). In the stuck case, where the pre-pause comment is refused, only the `itemNeedsWork` evaluation runs, so that is one anchor fetch per poll, not two or more. A second fetch happens only on an authorised resume, which happens once.
- **Only some paused items pay it.** Paused items skip the deep fetch unless a new comment puts them in `cycleSet`. The fetch also only runs when a human unprocessed comment exists.
- **A memo would bring back the stale-anchor risk.** A short-TTL anchor cache is the record-on-write staleness ADR 1813 rejects. If a human removes `fabrik:paused` in the UI and the item is re-paused inside the TTL, the cached early anchor makes the old comment look post-pause. That resumes the item again and re-arms the loop for up to the TTL.
- **Keying the memo on `updatedAt` doesn't help.** The `updatedAt` key would need to move on every label event to stay correct, and `ProjectItem` isn't guaranteed to carry a reliable one here.
- **The trade-off is already in the design.** The ADR states the direct fetch costs one paged REST call, only when a paused item already has an unprocessed human comment. A per-poll admission-vs-dispatch share isn't needed, because a refused item never reaches dispatch.
- **The outage point matches R6.** A persistent fetch error resumes by design, because a pause nobody can lift is the failure R6 forbids. An API outage lifting pauses that carry a human comment is the accepted cost, and a wasted cycle is recoverable.
- **Pushing would cost a review cycle.** A commit for a no-defect performance note would draw another bot review.

If polling volume on long-stuck items becomes a problem in practice, that should be a separate follow-up issue. It would need to solve label-removal-in-UI invalidation, for example by keying on the label's event ID rather than a TTL.

---
**Threads addressed:**
- `engine/comments.go:201` — resolved

arbeithand:
*branch: fabrik/issue-1813 | commit: 1746cb49 | main: 9289bb44fa363acf601d845276184b25f8636125 | 2026-09-20 20:42 UTC*

## Validation Report

### Requirements: 8/8 passed
- **R1:** Verified in `engine/comments.go`. `resumeAuthorised` resumes only when a human comment is created at or after the latest `fabrik:paused` `labeled` event. A resume removes the labels, and a re-pause writes a newer event, so one comment can't resume twice.
- **R2:** A comment created after the pause resumes on the next poll, including when an older unprocessable comment is also present. Covered by the predicate table in `engine/resume_anchor_test.go`.
- **R3:** The anchor is read directly from GitHub via `client.FetchLabelAppliedAt`, bypassing the record-on-write cache. It doesn't depend on process memory, and a restart test uses a fresh engine.
- **R4:** The `isAwaitingInput` branch uses the same predicate as the paused branch.
- **R5:** `raw` (the full `findNewComments` set) is still what `processComments` receives.
- **R6:** A fetch error, a zero anchor, a zero `CreatedAt` or an equal timestamp all resume. Only "every human comment strictly predates the pause" refuses. `commentsPredatePause` implements this and it's tested.
- **R7:** No code change was needed. `clearFailedStage` and `unblockAwaitingInput` stop running once the gate refuses, and tests assert the breaker counters are untouched.
- **R8:** `itemNeedsWork` (both branches) and `processItem` (both gates) call the same predicate.

The sim scenario `tests/sim/pause_resume_anchor_test.go` and the ADR, state-machine, USER_GUIDE, CLAUDE.md and `llms-full.txt` updates are present. There are no TODO/FIXME comments in the changed files.

### Test suite: full run had 6 failures, none caused by this change
- **Build and vet:** clean.
- **Full run:** `go test -race ./...` reported 3955 passed, 6 failed.
  - `engine` has 2 failures, `TestInvokeClaude_FabrikEnvVarsInjected` and its `FABRIK_PR_absent` subtest, which trip on host env vars (`NVM_INC`). The Review stage reported the same failure on a clean `origin/main`.
  - `cmd` has 1 failure, `TestExecute_ConfigYAMLApplied` (no return after SIGINT). The Implement stage reported it fails the same way on the base commit. I did not re-run it on main.
  - `tests/sim` had 3 merge-train failures in the full run: `TestMergeTrainLanding_OneAtATimeViaInteractionOnlyPoison`, `TestMergeTrainRunaway_DoesNotTripWithGenerousWindow` and `TestMergeTrainRunaway_TripsAndPausesQueuedMembers`.
- **Re-runs:**
  - The two runaway tests are flaky under `-race`. `TestMergeTrainRunaway_DoesNotTripWithGenerousWindow` failed on this branch in one isolated run and on `origin/main` too, and passed in other isolated runs.
  - Every other sim merge-train test I re-ran passed in isolation, and the targeted pause/resume/breaker engine tests passed (95).
- **Race detector:** no data races reported.

The existing human-only resume tests (`blocked_on_input_test.go`) still pass. The zero-anchor mock default takes the R6 resume path, so they resume unchanged.

### Branch state
Rebase outcome: `skipped-up-to-date`. The branch is 0 commits behind `origin/main`, and the PR was not in the merge queue. The worktree is clean and HEAD `1746cb49` matches the pushed branch. I removed the temporary main worktree I used for comparison.

### Verdict: READY TO MERGE

Rebase: skipped-up-to-date. PR mergeable: MERGEABLE, mergeStateStatus: CLEAN. Required checks: Test and vet: pass, Verify llms-full.txt is up to date: pass, Analyze (go): pass (listed twice). Requirements: 8/8 verified against the issue spec. Local `go test -race ./...` had 6 failures: two `TestInvokeClaude_FabrikEnvVarsInjected` cases caused by host env vars, `cmd`'s `TestExecute_ConfigYAMLApplied` (SIGINT), and three sim merge-train tests. The `TestMergeTrainRunaway_DoesNotTripWithGenerousWindow` failure also reproduces on `origin/main`. The other sim merge-train failures passed in isolation, and the targeted pause/resume/breaker tests all passed.

---
Used 11/100 turns, 1.1M input (20 raw + 997k cache-read + 120k cache-write) / 3k output tokens.

arbeithand:
🏭 **Fabrik merge-train** — Landed via singleton fast path PR #1816. That is this PR: the pinned base was already an ancestor of this head, this PR was mergeable, and its own CI was green and complete, so it was landed directly — no trial branch was assembled and no separate integration PR exists. See ADR-1644.

Issue #1777, opened 2026-09-18
shipped stage skills emit 17 command shapes that dontAsk always denies and no allowlist can permit


arbeithand:
## Problem

Fabrik's own shipped stage skills instruct workers to run command shapes that Fabrik's own default
permission mode (`--permission-mode dontAsk`) refuses, and that **no operator allowlist can permit**
— Claude Code Bash allow rules are prefix matches over the command string, so `x=$(…)`, `if`, `for`
and herestrings are unallowlistable by construction.

Reported in #1742 (@jmatthewpryor). He cited two sites; a full census of the shipped skills found
**17**:

| Skill | Occurrences | Notes |
|---|---|---|
| `fabrik-validate` | **13** | incl. lines 262-265 (three `$( )` assignments + a `read … <<<`) and the whole Pre-Completion Gate block, 272-343 |
| `fabrik-review` | 3 | 36-37 (the same base-branch block), 53 |
| `fabrik-implement` | 1 | 99 (`CURRENT_BODY=$(gh api …`) |

Shapes: 14 × `x=$(…)`, 2 × `if`, 1 × herestring.

### Why this is the highest-frequency denial in the pipeline

`fabrik-validate/SKILL.md:29` is **step 2** — the first real command of the stage:

```bash
base_branch=$(gh pr view --json baseRefName --jq .baseRefName 2>/dev/null)
if [ -z "$base_branch" ]; then
  ...
fi
```

`fabrik-review/SKILL.md:36` opens identically. So the first thing a Validate or Review worker does is
run a command that is guaranteed to be refused. Paired with #1741/#1775 (the denial is reported
tool-wide, and our comment asserts "no retry can fix this on its own"), a worker concludes Bash is
gone and abandons the stage having done nothing.

The reporter's census attributes the denials by command shape: `base_branch=$(gh` denied on **all 9**
appearances, `owner_repo=$(gh` 5/5, `BODY=$(gh` 3/3, `pr_number=$(gh` 2/2. The equivalent simple
commands are allowlisted and ran fine in the same sessions.

**Worst affected is the Pre-Completion Gate** (`fabrik-validate` 262-343) — the block that decides
whether Validate may complete is built almost entirely from these shapes.

## Summary

Rewrite every shell command in the shipped stage skills (`fabrik-validate`, `fabrik-review`,
`fabrik-implement`) that uses a shape `--permission-mode dontAsk` structurally refuses
[...]

Issue #1597, opened 2026-08-13
Validate exhausts FABRIK_MAX_REVIEW_CYCLES against a single review, and resolving the feedback does not stop it


totalslacker:
## `FABRIK_MAX_REVIEW_CYCLES` is exhausted against a single review, and addressing the feedback does not stop it

Two issues in `totalslacker/concept-graph` were paused at **Validate** on the same day with:

> The stage **Validate** has been re-invoked to address PR review feedback 5 time(s), which has reached the maximum configured limit (`FABRIK_MAX_REVIEW_CYCLES=5`).
>
> This usually means a reviewer (bot or human) is repeatedly requesting changes after each fix.

**No reviewer requested changes repeatedly.** Each PR had exactly **one** review submission, from `copilot-pull-request-reviewer[bot]`, and neither was a `CHANGES_REQUESTED` — both were `COMMENTED`.

| issue | PR | review submissions | threads | mergeable | CI |
|---|---|---|---|---|---|
| CG#305 | #307 | 1 (`COMMENTED`, 00:52:45) | 2, **both `isResolved: true`, `isOutdated: true`** | `MERGEABLE` / `CLEAN` | `test: SUCCESS` |
| CG#304 | #308 | 1 (`COMMENTED`, 00:59:15) | 1, `isResolved: false` | `MERGEABLE` / `CLEAN` | `test: SUCCESS` |

## The part that looks most like the bug

**CG#305's feedback was fully addressed and it was paused anyway.** Both threads are resolved *and* outdated — commits at 00:55:50 and 00:58:35 landed after the 00:52:45 review and superseded the lines it commented on. CI was green and the PR was clean. It then burned five Validate invocations and stopped.

So the counter is not measuring "a reviewer keeps rejecting the work". Whatever the loop's termination condition is, **an addressed, resolved, outdated review thread on a green PR does not satisfy it.**

## A second, narrower problem: a review that lands after the stage ends

On CG#304 the timing was:

```
00:58:58   last commit pushed
00:59:15   Copilot review submitted   ← 17 seconds later
```

The stage had already finished. Nothing was ever going to address that comment, so re-invoking Validate to "address PR review feedback" had nothing to act on — but it re-invoked five times regardless, then paused.

Worth considering whether a re-invocation that produces no commit should count differently from one that does, since five no-op invocations is what the limit is spending itself on.

## Impact

Both issues sat paused with `fabrik:paused` + `fabrik:awaiting-input` and needed manual intervention:

[...]

Issue #1555, opened 2026-08-12
Successful no-op comment re-invocations are unbounded (sibling of #1382)


arbeithand:
## Problem

A comment re-invocation that **succeeds** and does nothing is unbounded. Fabrik will redeliver the same already-processed comment to the same stage indefinitely, burning a full Claude invocation each time, with no counter that ever reaches a limit and no escalation.

This is the sibling of **#1382**, which is genuinely fixed. #1382 bounded *failing* comment re-invocations; this issue covers *succeeding* ones.

Observed live in production:

- `handarbeit/fabrik#1254`, Validate stage. The stage's own output: "**Seventh consecutive identical delivery** of the same stale, already-processed bot comment (superseded by the later human consolidation comments showing 53 actual unresolved findings). No new content, no action taken."
- `handarbeit/fabrik#1253`, same shape: "Same HEAD (`7ea6d674`), clean tree — unchanged. This is a **third delivery of the identical bot comment already processed as a No-Op twice**. No new content, nothing actionable… Not signaling `FABRIK_STAGE_COMPLETE`."

Both issues sat at `stage:Validate:complete` with PRs `MERGEABLE / CLEAN`, all checks green, zero commits behind `main` — cycling continuously and producing nothing. Neither carried `fabrik:paused`.

### Why nothing currently bounds it

- **`max_retries` does not apply.** It is documented as *"Max failed **stage** attempts before pausing the issue."* These attempts are neither failed nor plain stage attempts. #1382's own evidence from `history.json` shows the asymmetry that motivated it:

  ```
  Implement  IsComment=false  Success=false  ->   3 runs   <- max_retries: 3 held
  Implement  IsComment=true   Success=false  -> 141 runs   <- unbounded
  ```

- **`MaxReviewCycles` does not apply.** It governs `dispatchReviewReinvoke` — the PR-review-thread path. These invocations are issue-comment processing, headed `🏭 Fabrik — stage: Validate (comment review)`.

- **ADR-1518's guard does not apply.** `engine/catch_up_handlers.go:260` already anticipates a refund-masked loop:

  ```go
  // max, not ReviewCycles alone (ADR-1518): #1045's ReviewCycleDecremented
  // refund can hold ReviewCycles below MaxReviewCycles indefinitely when
  // every reinvoke happens to land no new commit. ReviewBlockedCycles is
  // the never-refunded counterpart ... so a refund-masked loop still
  // reaches the limit here.
[...]

Issue #1045, opened 2026-07-22
bug(review): bot review feedback delivered as a plain PR body comment is neither gated on nor auto-addressed (gap in #357)


verveguy:
## Summary

Fabrik's review-gate reinvoke mechanism (added by #357, refined by #1375/ADR-1375) misses substantive bot review feedback delivered in two forms:

1. **A plain PR body/issue comment with no formal review submission at all** (the original report). `wait_for_reviews` only counts formal review submissions, so this never clears the gate via review arrival, and the comment-processing path doesn't route it to the fixer because it lacks the inline `**File:**`/`**Diff context:**` headers the auto-fix path keys off.
2. **The body of a `COMMENTED` review from a reviewer whose findings are the substantive content of that body** (discovered while refiling this issue). `buildReviewBodyCommentsFromReviews` (`engine/reviews.go:1163`) deliberately excludes `State != "CHANGES_REQUESTED"` reviews from the reinvoke path — correct for bots like Copilot/Gemini that submit a generic "Pull request overview" as their `COMMENTED` body, but wrong for a reviewer (e.g. Pruefer, `handarbeit-pruefer`) that submits **exclusively** `COMMENTED` reviews whose bodies are the actual finding set (Pruefer's raise-only `REQUEST_CHANGES` per #1251, undconfigured threshold on this instance). On such a repo, the gate clears normally (a `COMMENTED` review is a non-`DISMISSED` review from an author) but the findings are never routed to the fixer, and the PR proceeds to merge unread — #1207's failure shape arriving through a different door.

Both delivery shapes must be fixed; fixing only one leaves a live gap.

### Note on the original Gemini/Copilot framing

The issue's original framing (Gemini producing only a sunset notice, `@copilot review` posting body comments) is stale for `handarbeit/fabrik` specifically: Pruefer now submits a formal review there, so failure mode 1 (gate never clears) does not occur on this repo today — a `COMMENTED` review is a non-`DISMISSED` review from an author, so the gate clears. Verified against PRs #1399, #1400, #1403, #1415. The plain-body-comment case (no submission at all) remains a real gap for other repos/bots (e.g. `@copilot review`'s inconsistent behavior) and is still in scope as delivery shape [1] above.

### Withdrawn design direction

[...]

PR #1817, opened 2026-09-20
fix(engine): don't charge a reinvoke cycle for invocations that never ran

### LINKED ISSUE #1812: a review reinvoke that never ran still charges a MaxReviewCycles cycle, pausing healthy issues during any Claude outage
## Problem

Fabrik pauses healthy issues at `FABRIK_MAX_REVIEW_CYCLES` when the reviewer had nothing left to
say. Two community reporters hit it (#1597 totalslacker, and adjacent #1743), and it then reproduced
on this project's own board — which is what makes it diagnosable.

**Issue #1777 / PR #1792, 2026-09-18.** Validate's own output: *"Clean — zero unallowlistable
shapes, build/vet clean, plugin tests pass. No further code changes are needed… Nothing to commit."*
PR `CLEAN`, `MERGEABLE`, 6/6 checks green, `stage:Validate:complete` applied, no requested
reviewers, two `COMMENTED` reviews from `handarbeit-pruefer`, one review thread `isResolved: true`.
Paused anyway at 05:29Z.

The five invocations that spent the budget are all still on disk, in
`.fabrik/logs/handarbeit-fabrik/issue-1777/`:

```
Validate-comment-review-20260918-052917  7498 bytes
Validate-comment-review-20260918-052927  7498 bytes
Validate-comment-review-20260918-052936  7498 bytes
Validate-comment-review-20260918-052946  7497 bytes
Validate-comment-review-20260918-052954  7497 bytes
```

**Five cycles in 37 seconds.** Every one of them is a single result line: `is_error: true`,
`terminal_reason: "api_error"`, `api_error_status: 429`, `num_turns: 1`, `total_cost_usd: 0`,
`"You've hit your session limit · resets 3:30am"`. The stage never ran. Not once. The cycle-limit
pause comment posted at 05:29:59 — five seconds after the last one.

The same signature, at **exactly 5**, on six issues across three daemons and three separate dates:

| daemon | issue | 5 failed reinvokes | cycle-limit pause |
|---|---|---|---|
| fabrik | #1777 | 05:29:17–:54 | 05:29:59 |
| fabrik | #1802 | 16:15:30–:57 | 16:16:02 |
| fabrik | #1750 | 13:27:12–:29 | 13:27:33 |
| 86ed | #268, #266 | ~13s window | — |
| concept-mapping | #617, #530 | ~30s window | — |

Exactly five, every time, because five is the limit. The pause always lands 4–5 seconds after the
last failed invocation.

**This is not the `ReviewBlockedCycles` / authoritative-gate hypothesis**, which was investigated and
is wrong: for #1792 `reviewDecision` is empty and no review is `CHANGES_REQUESTED`, so
`reviewGateAuthorityVerdict` returns *satisfied* and the gate cleared. `blocked` and `timedOut` were
both `false`, so `ReviewBlockedCycles` never incremented. It was plain `ReviewCycles`.

## Summary

`handleReviewGate` applies `ReviewCycleIncremented` **synchronously, before dispatch**
(`engine/catch_up_handlers.go:271`). The compensating refund, `ReviewCycleDecremented` (#1045), sits
in `dispatchReviewReinvoke`'s `after` hook (`engine/reviews.go`) behind `if err != nil { return }`.
So an invocation that **provably never ran** is charged a review cycle it can never get back.

The guard is not careless — its comment reasons that a failed dispatch "never got a chance to act,
so it is not evidence there was nothing to act on." That is correct about *evidence*, and wrong
about *accounting*. `MaxReviewCycles` bounds attempts to converge. A dispatch that died at turn 1
having spent $0.00 was not an attempt.

The engine already knows this. On the stage-dispatch path the identical exit is explicitly exempted,
logging `"api_error exit detected — stage did not run, not charged against max_retries"`
(`engine/claude.go`). `finalizeStageOutcome` routes it to `handleAPIErrorExit` and skips
`StageRetryIncremented`. The reinvoke path has no equivalent, so the same exit is free on one path
and permanently charged on the other.

It compounds: because the reinvoke never ran, #1555's durable `<!-- fabrik:review-ids-addressed -->`
marker is never written, so the same review body is still actionable on the next poll. There is no
backoff anywhere on this path. The loop runs at poll cadence until the budget is gone.

`ReviewCycles` is the counter that pauses issues today, but the defect is shared. `RebaseCycles`
and `CIFixCycles` are incremented by the same `dispatchWithCycleLimit` helper and have **no refund at
all** — `dispatchCIFixReinvoke`'s `after` hook records a no-op *debounce*, not a cycle refund, and
is gated on the same `err != nil`. A long enough outage exhausts those budgets too, with the same
wrong pause at the end.

## Requirements

**R1 — A did-not-run invocation must not be charged a reinvoke cycle.** When a reinvoke's
`processComments` returns an error whose classification means the Claude invocation never executed,
the pre-dispatch cycle increment must be compensated. The set is the one the stage path already
treats this way: `claudeUsageLimitError`, `claudeAPIErrorExit`, and `apiKeyHelperDetectedError`.

**R2 — Apply it to all three cycle counters.** `ReviewCycles`, `RebaseCycles` and `CIFixCycles` are
incremented by the same pre-dispatch path and carry the same defect. The mechanism belongs in the
shared scaffold (`dispatchReinvoke` / `dispatchWithCycleLimit`), not copied three times. `RebaseCycles`
and `CIFixCycles` need `*Decremented` mutations added; `ReviewCycleDecremented` already exists and
must keep its existing #1045 no-op-refund behaviour unchanged.

**R3 — An error that is *not* a did-not-run classification keeps today's behaviour.** A turn-limit
exit, a resume failure, a tools-denied exit, a genuine mid-run crash, a network failure during
output publication — all of these may have done real work, and must continue to be charged exactly
as they are now. Do not widen this to "any error".

**R4 — Never refund a cycle whose invocation did work.** The `#1045` turns/cost reasoning applies:
an invocation that consumed turns and incurred cost is not a did-not-run exit regardless of how it
was labelled. Where the classification is ambiguous, charge the cycle. Under-refunding costs an
unnecessary pause; over-refunding removes the only bound on a genuine non-convergence loop.

**R5 — The refund must be visible.** Emit a log line naming the counter, the stage and the
classification, in the shape of the existing `"no new commit pushed … decrementing review cycle
counter (#1045)"` line. The current defect was only diagnosable because per-issue transcripts
survive; a silent refund would make the next one harder, not easier.

**R6 — Floor at zero and stay idempotent.** `ReviewCycleDecremented` is already floored by the store.
The new mutations must be too, and applying a refund twice for one increment must be harmless.

**R7 — The cycle-limit pause message must stop misdiagnosing this case.** The current text asserts
*"This usually means a reviewer (bot or human) is repeatedly requesting changes after each fix"* —
which was actively misleading on all six issues above. When the cycles that reached the limit were
predominantly did-not-run exits, say so, and name the likely cause. An operator reading #1777's
pause had no way to reach the truth from the comment.

## Scope

**In scope:** `engine/catch_up_handlers.go` (`dispatchWithCycleLimit`), `engine/reinvoke.go`
(`dispatchReinvoke`'s `after` scaffold), `engine/reviews.go` (`dispatchReviewReinvoke`,
`pauseForReviewCycleLimit`'s message), `engine/ci.go` (`dispatchCIFixReinvoke`),
`engine/merge_gate.go` (`dispatchRebaseReinvoke`, the rebase reinvoke dispatch),
`internal/itemstate/` (new `*Decremented` mutations + store cases),
`docs/state-machine.md`'s review-gate and reinvoke sections (regenerate `docs/llms-full.txt` in the
same commit), tests.

**Out of scope:** detecting the 429 session limit correctly in the first place (sibling issue — that
one prevents the storm, this one makes a leaked exit harmless; **both are wanted, neither
substitutes for the other**); the paused-resume loop (its own issue); `MaxReviewCycles`' default
value; the #1555 dedup marker; adding backoff to the reinvoke poll cadence.

## Open Questions

None. The failing invocations, the counter, the pause timestamps and the guard are all identified
and reproduced; the requirements above are decidable without further investigation.

## Relationship to the community report

This is a **work issue** for community report #1597 (@totalslacker) — *"Validate exhausts
FABRIK_MAX_REVIEW_CYCLES against a single review, and resolving the feedback does not stop it."*
On that reporter's CG#305 both review threads were `isResolved: true` **and** `isOutdated: true`,
CI green, PR clean, one `COMMENTED` review from `copilot-pull-request-reviewer[bot]`, never
`CHANGES_REQUESTED` — and it still burned five Validate invocations. That is this defect's exact
shape.

An earlier diagnosis on #1597 attributed it to a v0.0.78-only window (#1045 shipping ahead of
#1555's durable dedup marker) and advised upgrading. A correction was posted 2026-09-18: #1777
happened on `main`, well past v0.0.79, with the marker present. This issue supersedes that
diagnosis entirely — it was never a version window.

**Linkage rules (`CLAUDE.md`, "Handling Community Bug Reports") — these are mandatory:**

- The PR for this issue carries `Closes #<this issue>` and mentions **#1597 by bare number only**.
- **Never** place a closing keyword (`Closes` / `Fixes` / `Resolves`) near #1597. GitHub parses
  adjacency, not grammar — it fires inside a negation too, and has already auto-closed a community
  report on this repo that way. Safe phrasings: "#1597 is the report behind this", "mentioned for
  context: #1597".
- #1597 is closed **by hand**, by the report owner, only once a release containing the fix has
  shipped. Not on merge.

Adjacent, not addressed here: #1743 (@jmatthewpryor), *"stage did not make progress" is asserted
from `!completed`, never measured* — already spec'd into #1775. It touches the same code path.

## Prior Art / Context

- **#1045 / `ReviewCycleDecremented`** established the refund-a-cycle-that-changed-nothing principle
  and the `err != nil` guard this issue narrows. Its no-op-on-HEAD case is correct and untouched;
  this adds a second, disjoint refund trigger.
- **ADR-1518** added `ReviewBlockedCycles` as a never-refunded counterpart so a refund-masked loop
  still reaches the limit. That reasoning is unaffected: a did-not-run cycle never increments
  `ReviewBlockedCycles` (it is gated on `blocked || timedOut`), so nothing here weakens it.
- **ADR-1458 / `handleAPIErrorExit`** is the exact precedent being extended: the stage path already
  treats this exit as "the stage never ran" and exempts it from `max_retries`. This issue applies
  the same judgement to the reinvoke cycle counters.
- **`fabrik:claude-limit` and `fabrik:api-key-helper-detected`** (ADR-1119, ADR-1346) both record
  `StageAttempted` without `StageRetryIncremented` for precisely this reason. The reinvoke path is
  the outlier.

## Risks / Dependencies

- **The real risk is removing the bound.** A refund that is too generous lets a genuinely
  non-converging review loop run forever. R3 and R4 exist to prevent that, and `ReviewBlockedCycles`
  remains as the independent backstop for the blocked-gate case.
- **No dependency on the sibling issues**, and no ordering constraint — they touch different files.
  If the 429 detection lands first, this becomes rarer but not unnecessary: any did-not-run exit
  reaches it.
- **Touches a shared helper used by three call paths.** Rebase and CI-fix regressions are the thing
  to watch in review; their existing cycle-limit tests must stay green.

## Notes

The discriminator between this defect and genuine non-convergence is **elapsed time**. Real review
cycles take minutes each because Claude actually runs. This burns all five in under a minute. That
is worth saying in the pause comment (R7) and worth remembering when triaging the next report.

The ADR for this issue is numbered after it, per `CLAUDE.md`'s ADR-numbering convention
(`adrs/1812-kebab-title.md`).

arbeithand:
Closes #1812

A review, CI-fix or rebase reinvoke whose Claude invocation provably never ran (a 429 or `api_error` exit, a usage limit, an account suspension, or `apiKeyHelper`) is no longer charged a cycle. Such an exit now refunds the cycle counters instead of eating the `MaxReviewCycles` budget.

Before this change, five such exits in under a minute paused a healthy issue, and the pause message blamed a reviewer. That is the #1777 incident. #1597 is the report behind this; it stays open until a release containing the fix has shipped.

**What changed**
- **Refund in the shared scaffold.** `dispatchReinvoke` (`engine/reinvoke.go`) compensates the pre-dispatch charge. This is the one place that also covers the rebase and CI-fix increments made inline in `checkAutoMergeConvergence`.
- **Classification side channel.** `processCommentsClassified` returns a `didNotRunKind` next to the unchanged error, and `processComments` is a thin wrapper, so the user-comment callers are unchanged. This was needed because usage-limit and suspension exits return `nil` today.
- **What counts as did-not-run.**
  - Refunded: `claudeUsageLimitError`, `claudeAPIErrorExit`, the account-suspension gate, and `apiKeyHelperDetectedError`. The `apiKeyHelperDetectedError` arm is forward-compat only, since the comment path never produces it today.
  - Still charged: turn limit, resume failure, tools denied, mid-run crashes and publish failures.
  - Any ambiguity is charged too.
- **`ReviewBlockedCycles` refund.** It is refunded for did-not-run exits only. The spec said a did-not-run cycle never increments it, but that holds only when the gate is not blocked. ADR-1518's no-refund guarantee still applies to a run that executed and no-op'd.
- **`after` hook skipped.** It no longer runs on a did-not-run exit. That removes the bogus CI-fix no-op-SHA debounce and the rebase auto-merge re-enable on those paths, and keeps this refund disjoint from the #1045 one.
- **New itemstate mutations.** `ReviewBlockedCycleDecremented`, `CIFixCycleDecremented`, `RebaseCycleDecremented` and `DidNotRunReinvokeRecorded`. Each decrement floors at zero and emits no `Change` on an already-zero counter.
- **Pause message.** A never-refunded `DidNotRunReinvokes` tally feeds `didNotRunPauseNote` in the review, CI-fix and rebase cycle-limit messages. When the tally reaches the cycle count, the "reviewer keeps requesting changes" text is replaced with the real cause, the elapsed-time hint and where to look in the logs. A zero tally leaves the message byte-identical, and the dedup fragments are unchanged.
- **Refund log line.** Each refund logs `invocation did not run (<kind>) — refunding <counter> cycle counter … (#1812)`.
- **Docs.** `docs/state-machine.md` §6.2 and §7.3a, the regenerated `docs/llms-full.txt`, and `adrs/1812-reinvoke-did-not-run-cycle-refund.md`.

**Known limit:** after the refund, a did-not-run storm is bounded by the comment circuit breaker and the #1555 no-op breaker, both about 10, instead of by `MaxReviewCycles`. Backoff is out of scope, and detecting the 429 upstream (the sibling issue) is what prevents the storm.

**How to test**
- `go test -race ./engine ./internal/itemstate`. The new tests are `engine/did_not_run_refund_test.go` and `internal/itemstate/did_not_run_test.go`.
- They cover a five-in-a-row `api_error` storm, the usage-limit and suspension paths, the blocked-gate variant, CI-fix and rebase, and the negative cases that must stay charged.
- The full `go test -race ./...` passes except `cmd` `TestExecute_ConfigYAMLApplied`, which fails identically on the untouched base commit `9289bb44` (the SIGINT/network case).


arbeithand:
*branch: fabrik/issue-1812 | commit: c4d876da | main: 9289bb44fa363acf601d845276184b25f8636125 | 2026-09-20 20:24 UTC*

Implemented the did-not-run refund in `dispatchReinvoke`: a review, CI-fix or rebase reinvoke that provably never ran (`api_error`, usage limit, account suspension, `apiKeyHelper`) now refunds its cycle counter instead of spending `MaxReviewCycles`. It also refunds `ReviewBlockedCycles` in that case, adds a did-not-run tally so the cycle-limit pause messages name the real cause, and adds tests, `state-machine.md` updates and ADR 1812. `go vet` is clean and the full suite passes except `cmd` `TestExecute_ConfigYAMLApplied`, which fails identically on the base commit. All work is pushed and Plan tasks 1–9 are checked off.

---
Used 71/100 turns, 9.3M input (110 raw + 9.1M cache-read + 199k cache-write) / 40k output tokens.

arbeithand:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-1812 | commit: c4d876da | main: 9289bb44fa363acf601d845276184b25f8636125 | 2026-09-20 20:50 UTC*

## Review Findings

### Fixed
Nothing needed fixing. I found no correctness, security or convention problems in the diff.

### Verified
- **Rebase:** skipped, since the branch is 0 commits behind `origin/main`.
- **Call sites:** every caller of `dispatchReviewReinvoke`, `dispatchCIFixReinvoke` and `dispatchRebaseReinvoke` increments its counter first. That includes the inline increments in `checkAutoMergeConvergence` (`merge_gate.go` around lines 439 and 469). The refund in the shared `dispatchReinvoke` therefore never decrements a counter that wasn't charged.
- **R1/R2:** `usage-limit`, `api-error`, account-suspension and `apiKeyHelper` exits all take the refund path. It covers `ReviewCycles`, `RebaseCycles` and `CIFixCycles`, and also `ReviewBlockedCycles`, which is charged when `blocked || timedOut` is true.
- **R3/R4:** turn-limit, resume-failure, tools-denied and generic errors stay charged. I made `classifyDidNotRun` return a did-not-run kind for every error and `TestClassifyDidNotRun` failed, so the guard is exercised.
- **R5:** each refund logs `invocation did not run (<kind>) — refunding <counter> cycle counter … (#1812)`.
- **R6:** the new decrements are floored at zero, no-op without a `Change` on an already-zero counter, and are cleared by `EngineCyclesCleared`.
- **R7:** the three cycle-limit pause messages name the outage cause when the did-not-run tally is at least the cycle count. A zero tally leaves the message unchanged, so the `#1460` pause dedup still matches.
- **Docs:** `docs/llms-full.txt` matches a fresh regeneration and the tree is clean. ADR 1812 follows the naming convention.
- **Build and vet:** `go build ./...` and `go vet ./...` pass, and the `engine` and `internal/itemstate` tests for this change pass.
- **Failing tests:** `TestInvokeClaude_FabrikEnvVarsInjected` fails because the host's `NVM_INC` env var leaks into the test; it fails the same way on `origin/main`. The full `tests/sim` package also fails on `origin/main` (I didn't identify which tests failed there). `TestMergeTrainAssembly_PoisonMatrix` passes alone on this branch, and fails only in the full-package run.

### Non-blocking notes
- The "never ran" pause note can be wrong in a mixed case. The tally is never refunded and is not per-counter. A burst of outage exits followed by real non-converging cycles could make the pause blame Claude availability instead of the reviewer. The text says "most likely", the ADR records this trade-off, and it can only show up after a genuine outage.
- After the refund, an outage storm is bounded only by the comment circuit breakers (about 10 invocations). The ADR and `state-machine.md` both state this. Backoff and upstream 429 detection are out of scope, as the issue says.

Reviewed the did-not-run reinvoke cycle refund. It covers review, CI-fix and rebase, including `ReviewBlockedCycles`, and the pause text now names the outage cause. Errors that may have done work (turn limit, resume failure, tools denied) stay charged. I found nothing to fix. The branch is up to date with `main`, and build and vet are clean. The remaining test failures (`TestInvokeClaude_FabrikEnvVarsInjected` on a host `NVM_INC` env var, the full `tests/sim` package) also fail on `origin/main`. The PR is ready for human review.

---
Used 19/100 turns, 2.3M input (36 raw + 2.1M cache-read + 136k cache-write) / 6k output tokens.

arbeithand:
*branch: fabrik/issue-1812 | commit: 588e73a6 | main: 9089417fc69ca6110de9a600661331af64c2d781 | 2026-09-20 21:00 UTC*

## Review Findings

### Fixed
- **Pause message could misattribute a real non-converging loop to a Claude outage** (Pruefer thread `PRRT_kwDOR2bN_c6kLzpw`; I had noted it as non-blocking in my first review). `DidNotRunReinvokes` was a per-stage tally cleared only by `EngineCyclesCleared`. With `MaxReviewCycles=5`, five refunded 429s spread over days plus five genuine reviewer cycles would satisfy `tally >= cycleCount`. The pause would then blame Claude availability and hide the real cause. The same tally was shared by the review, CI-fix and rebase reinvokes.
  - **Fix:** new `DidNotRunReinvokesReset` mutation. `dispatchReinvoke` applies it whenever a reinvoke stays charged, meaning it ran or is ambiguous. The tally is now the streak of never-ran reinvokes since the last genuine cycle.
  - **Reset behaviour:** a reset is a no-op (no `Change`) when the tally is already zero. A did-not-run exit extends the streak and does not reset it. A real reinvoke of any of the three kinds ends the streak, which falls back to the original message wording.
  - **Tests:** the new `TestReviewReinvoke_ChargedCycleResetsDidNotRunTally` and `TestReviewReinvoke_DidNotRunKeepsTally`, plus `TestApplyDidNotRunReinvokesReset`. With the reset removed, the charged-cycle test fails in all four subcases.
  - **Docs:** ADR 1812 decision 6, `state-machine.md` §6.2 and the state-ownership table are updated, and `docs/llms-full.txt` is regenerated.

I did not take the per-counter tally option. With reset-on-charge, a stale tally from another counter cannot survive a real cycle anyway.

### Verified
- **Vet and build:** `go build` and `go vet` are clean.
- **Tests:** `./engine` and `./internal/...` pass under `-race` except `TestInvokeClaude_FabrikEnvVarsInjected`. That test fails on the host's `NVM_INC` env var and fails the same way on `origin/main`.
- **Branch:** it is 0 commits behind `main`. It was already rebased, so the push was rejected once; I pushed with a lease pinned to the remote tip I had inspected (`c4d876da`), whose only extra commits were the old copies of my own.

Addressed Pruefer's finding: the did-not-run tally is now reset whenever a reinvoke stays charged, so an earlier outage can no longer make a genuine non-converging loop's pause message blame Claude availability. Added the reset mutation with tests, updated ADR 1812 and `state-machine.md`, and regenerated `docs/llms-full.txt`. Pushed to the branch. The only test failure is an env-var (`NVM_INC`) test that also fails on `main`.

---
**Threads addressed:**
- `engine/reinvoke.go:242` — resolved

arbeithand:
*branch: fabrik/issue-1812 | commit: 588e73a6 | main: 9089417fc69ca6110de9a600661331af64c2d781 | 2026-09-20 21:08 UTC*

## Validation Report

### Requirements: 7/7 passed
- **R1:** a reinvoke that never ran no longer spends a cycle. `dispatchReinvoke` in `engine/reinvoke.go` refunds it for `api_error`, usage-limit, account-suspension and `apiKeyHelper` exits. `processComments` in `engine/comments.go` now exposes the classification, so the usage-limit case, which used to arrive as `nil`, is caught.
- **R2:** the refund covers `ReviewCycles`, `RebaseCycles` and `CIFixCycles`. It also covers `ReviewBlockedCycles`, which Research found could otherwise leave the same false pause. The refund is in the shared scaffold, so it also covers the inline increments in `checkAutoMergeConvergence`. Existing #1045 no-op-refund behaviour is unchanged.
- **R3/R4:** turn-limit, resume-failure, tools-denied and generic errors stay charged. Review confirmed the guard is exercised: it made `classifyDidNotRun` return a did-not-run kind for every error and `TestClassifyDidNotRun` failed. The `TurnsUsed > 0 && CostUSD > 0` guard is inside the existing classifiers.
- **R5:** each refund logs the counter, stage and classification.
- **R6:** the new decrements are floored at zero and emit no `Change` when the counter is already zero, so a double refund is harmless. `EngineCyclesCleared` also clears the new tally.
- **R7:** the cycle-limit pause messages name the likely outage cause when the did-not-run tally reaches the cycle count. The tally is reset by any reinvoke that stays charged, which fixes the Pruefer finding from Review. A zero tally leaves the original wording.
- **Docs and ADR:** `adrs/1812-reinvoke-did-not-run-cycle-refund.md` and `docs/state-machine.md` are updated. `docs/llms-full.txt` is regenerated, and the "Verify llms-full.txt is up to date" check reports pass.

### Test Suite: 2 known failures, unrelated to this PR
- I ran `./engine`, `./internal/...`, `./github/...`, `./boardcache/...`, `./stages/...` and `./plugin/...` under `-race`. I did not re-run `cmd` or `tests/sim` in this pass. Build and vet are clean.
- Result: 4206 passed, 2 failed, 1 skipped.
- The two failures are `TestInvokeClaude_FabrikEnvVarsInjected` and its subtest. The host's `NVM_INC` env var leaks into the test. Review reported that it fails identically on `origin/main`, and the diff touches nothing near env-var injection.

- Existing cycle-limit, convergence and `mutation_test.go` decrement tests are in the passing set.
- The diff has no TODO or FIXME comments.

### Branch state
- The branch is 0 commits behind `origin/main`, and the working tree is clean.
- Local HEAD `588e73a6` matches `origin/fabrik/issue-1812`.
- I did not rebase (`skipped-up-to-date`) and made no code changes.

### Verdict: READY TO MERGE

Rebase: skipped-up-to-date. PR #1817 mergeable: MERGEABLE, mergeStateStatus: BLOCKED. Required checks: Test and vet: pass, Verify llms-full.txt is up to date: pass, Analyze (go): pass. The engine's CI gate decides what happens next. Requirements: 7/7 verified against the issue spec. A reinvoke that never ran (api_error, usage-limit, suspension, apiKeyHelper) now refunds its review, CI-fix or rebase cycle, and `ReviewBlockedCycles` too, and the pause messages name the outage cause. Errors that may have done work stay charged. Local test suite: 4206 passed, 2 failed across 12 packages under -race, build and vet clean. The two failures are `TestInvokeClaude_FabrikEnvVarsInjected` and its subtest, caused by the host's `NVM_INC` env var and also failing on `origin/main`.

---
Used 6/100 turns, 460k input (10 raw + 350k cache-read + 110k cache-write) / 2k output tokens.

arbeithand:
🏭 **Fabrik merge-train** — Landed via singleton fast path PR #1817. That is this PR: the pinned base was already an ancestor of this head, this PR was mergeable, and its own CI was green and complete, so it was landed directly — no trial branch was assembled and no separate integration PR exists. See ADR-1644.

PR #1824, opened 2026-09-20
feat(engine): registry-independent worker-session sweep for orphan reaping

### LINKED ISSUE #1814: bug(cleanup): orphans with no registry entry are invisible to both reap paths — #1798 closes the common case, not the race
## Problem

#1798's session-scoped reaper substantially reduced orphan leakage, but it did not close it. Orphans were observed surviving **after** the fix was deployed and running, on three separate occasions. The gap is structural, not an implementation bug: **a descendant that was never recorded in the registry is invisible to both reap paths.**

Both paths operate on `.fabrik/state/descendants.json`:

- `reapTrackedDescendants` (invocation end) loads *entries for that worker* and kills them.
- `sweepStaleDescendants` (the hourly `proc-janitor`) loads the *entire registry* and kills entries whose worker is confirmed dead.

Neither can find a live orphan that has no entry. `trackWorkerDescendants` records on a `descendantScanInterval` tick (3s in production), so a descendant spawned in the final seconds before its worker exits, or during a transient `ps` failure, is never recorded. Nothing reaps it, ever.

### Observed, post-deployment

Engine running `dev(5a4ed3b)` with the reaper live and `proc-janitor` reporting clean hourly cycles.

**2026-09-18 ~23:23**: one orphan, 3 minutes old (created ~3.5h after the reaper went live):

```
pid 26446  ppid=1  sid=14372 (leader exited)  96.2% CPU
cwd /private/var/folders/.../TestCrossRepoSpawn.../worktrees/acme-gadgets/issue-1
.fabrik/state/descendants.json == []
```

**2026-09-19 ~05:38**: two more:

```
pid 64286  ppid=1  sid=62076 DEAD  1.0% CPU  age 5h42m
           cwd .../TestRestartRecovery_KillAfterPRCreatedBeforeReady.../worktrees/acme-widgets/issue-1
pid 82584  ppid=1  sid=58248 DEAD  0.6% CPU  age 31m
           cwd .../TestCIFixReinvokeCycleLimit.../worktrees/acme-widgets/issue-1
.fabrik/state/descendants.json == []
```

In every case the information needed to identify and kill these existed. Each was a **live process whose session leader is dead**, discoverable with the same `unix.Getsid` primitive `sessionScopedDescendantsFromProcs` already uses. Nothing consults it, because discovery is registry-driven.

What *is* working: every one of these carried a real, distinct SID, and live workers were confirmed as their own session leaders (`sid == own pid`). `Setsid: true` (`engine/procattr_unix.go`) is doing its job. The lineage is intact and legible. It is simply not read at sweep time.

### A second orphan shape, not just test binaries

**2026-09-19 ~05:59**, same host: **56 orphaned Claude Code Bash-tool shells**, all `PPID 1`, all with **zero children**, all ~35 minutes old, at 34–99% CPU each:

```
/bin/zsh -c source ~/.claude-v3rv/shell-snapshots/snapshot-zsh-<id>.sh
cwd ~/dev/86ed/.fabrik/worktrees/verveguy-86ed/issue-711/apps/capture
```

These came from a **different daemon on a pre-#1798 binary** (no reaper at all), so they are not a failure of this mechanism. They do bear on its design:

1. The leaked process is not always a test binary. It can be **the Bash tool's own shell**, spinning with no children. Any registry-independent sweep must cover this shape.
2. **Blast radius crosses instances.** Those 56 shells drove host load from ~12 to **263** and starved an unrelated daemon's Validate stage into a 30-minute `max_wall_time` timeout after only 11 turns. That was a false failure on a healthy change. One instance's leak corrupts another instance's test signals.

## Summary

Make orphan discovery independent of per-descendant registry entries. Fabrik must durably record each worker at spawn time, when its PID and identity are known with no race. A periodic sweep, and an immediate post-exit sweep at invocation end, must enumerate live processes and reap any whose session ID equals the PID of a recorded worker that is confirmed dead. The existing per-descendant registry path stays as a fast path. Ownership is established from the worker side, and a process is signalled only when it is provably a descendant of a dead Fabrik worker.

## Requirements

- **R1: Registry-independent discovery.** The periodic sweep must find an orphan that has *no* registry entry, by enumerating live processes and identifying those whose session leader is a dead, recorded Fabrik worker. Per-descendant registry entries remain a fast path; they must stop being the only path.
- **R2: Ownership positively established before signalling.** Same discipline as #1798's R5. A candidate must be provably a Fabrik worker descendant. A dead session leader alone is not sufficient evidence, because arbitrary user processes can have dead session leaders. Failing open (leaking) is strictly preferable to killing a developer's process. Ownership means the process's SID equals the PID of a worker recorded by this instance, and that worker is confirmed dead.
- **R3: Cover non-test descendants.** Ownership is by session lineage, not by process name or type. Bash-tool shells (`/bin/zsh -c source .../shell-snapshots/...`) with zero children must be reaped, not just test binaries. No comm/args allow-list may gate the kill.
- **R4: Survive engine restart.** Whatever identifies ownership must be durable on disk, so an orphan from a previous engine run is still reapable by the next run.
- **R5: Close the discovery race, don't just narrow it.** Shrinking `descendantScanInterval` reduces the window but cannot eliminate it. Correctness must not depend on catching a descendant within N seconds of its creation. In particular, at invocation end (after the worker has exited) the engine must also sweep by the worker's SID, so a descendant spawned in the final seconds is caught at that point and not left for a later janitor cycle.
- **R6: No regression of #1798.** `reapTrackedDescendants` and the existing registry path keep working. This change is additive.
- **R7: Observability.** Every registry-independent reap logs pid, comm, the dead session leader (worker PID) it was matched to, and why, using the existing `[#N kill]` convention.
- **R8: PID-reuse safety on the worker side.** A recorded worker is "dead" only when its PID is no longer alive *or* the live process at that PID no longer matches the recorded identity fingerprint (`comm`/`lstart`). If a recycled PID now belongs to a *new* live worker (which is also a session leader with that same PID as its SID), processes that belong to the new worker must not be reaped. The spec does not prescribe how to distinguish them, but it must not rely on SID equality alone in that case. A transient fingerprint-lookup failure must be treated as inconclusive (fail open), consistent with the fail-open discipline established in ADR-1798.
- **R9: Worker record lifecycle.** A worker record exists from spawn. It must not be discarded while a live process with that SID could still exist. It must also not accumulate forever, so a record with no remaining live members and a confirmed-dead worker is eventually pruned. If the worker's own fingerprint cannot be captured at spawn, the record must not cause a sweep to signal anything on a liveness-only basis while the PID is alive. It must fail open.
- **R10: Documentation.** `docs/stage-lifecycle.md` (Subprocess Cleanup) and `docs/state-machine.md` (janitor section) are updated to describe the new path, and `docs/llms-full.txt` is regenerated (`bash scripts/generate-llms-full.sh`) in the same change.

## Scope

**In scope**
- Recording a durable per-invocation worker record at spawn.
- A registry-independent, SID-based sweep, run both periodically (the existing `proc-janitor` cadence, `JanitorIntervalHours`) and immediately after a worker exits at invocation end.
- Expected touch points: `engine/descendant_reap_unix.go`, `engine/descendant_registry.go`, `engine/procattr_unix.go`, `engine/janitor.go`, plus tests and docs.
- A new ADR (see AC7).

**Out of scope**
- Windows (unchanged no-op precedent).
- Reaping *other* Fabrik instances' orphans, or orphans of workers this instance never recorded. Each daemon reaps its own. The cross-instance blast radius is motivation, not a requirement. The 56-shell incident came from a different daemon on a pre-#1798 binary and would not be reaped by this change.
- Orphans that predate the introduction of worker records. They have no worker record and are not covered. This is an accepted fail-open gap.
- Descendants that call `setsid()` themselves. Their SID diverges permanently. This is the known gap already recorded in ADR-1798 and covered only partially by ADR-063.
- Why orphans busy-spin rather than idle (#1805).
- Changing `Setsid: true` semantics or `killProcGroup`.

**Assumptions**
- "Within one interval" (AC2) means one run of the existing hourly `proc-janitor` cadence, with the invocation-end sweep (R5) covering the just-exited-worker case immediately. No new config knob is introduced.
- `unix.Getsid` remains the SID source, since `ps`'s session column is masked on macOS (ADR-1798).
- Kill scope stays a single PID per process, as in ADR-1798. A reaped process's own children are each discovered independently, because every descendant carries the worker's SID.

## Acceptance

- **AC1**: A descendant spawned within one `descendantScanInterval` of its worker's exit (never recorded) is reaped. A test spawns a real detached child in that window and asserts it is gone.
- **AC2**: An orphan with **no registry entry**, whose session leader is a dead recorded worker, is reaped by the periodic sweep within one interval. This includes one that survived a previous engine run (worker record loaded from disk).
- **AC3**: A process whose session leader is **alive** is never signalled, even with a matching SID (that invocation is still in flight). This includes the recycled-PID case in R8, where a live new worker now holds a dead worker's former PID.
- **AC4**: A process not descended from any recorded Fabrik worker is never signalled. This is pinned by a test that fails if the ownership check is removed (neutralization test).
- **AC5**: An orphaned Bash-tool shell with zero children is reaped, not only test binaries.
- **AC6**: After a full sim-suite run under the engine, no orphan remains with a dead session leader.
- **AC7**: An ADR (`adrs/1814-*.md`, numbered after this issue) records the decision. It states plainly why registry-driven discovery alone was insufficient (a descendant never recorded is invisible to every reap path), so the assumption is not re-adopted. It also records the PID-reuse handling and the fail-open gaps listed under Scope.
- **AC8**: All existing #1798 tests (`reapTrackedDescendants`, `sweepStaleDescendants`, registry) still pass unmodified (R6).

## Open Questions

None outstanding. The sweep cadence assumption above is the one point a reviewer may want to challenge.

## Prior Art / Context

- **ADR-1798**: session-scoped reaper. Its SID lineage insight and `Setsid: true` are load-bearing and confirmed working. This issue extends it. It rejected PPID-walk discovery because reparenting to init is sub-second. Registry-driven *recording* has the same timing weakness on a smaller scale (the 3s tick).
- **ADR-063 / #975**: cwd-rooted reaper. Its assumption (orphan cwd inside the worktree) is also violated by every cwd observed here, since each is a `t.TempDir()`.
- **ADR-054**: kill escalation.
- **#1805**: why orphans spin.
- **Original issue's suggested approach (non-prescriptive)**: record the *worker* rather than chasing each descendant. Its PID and fingerprint are known at spawn with no race, and `setsid()` guarantees every descendant carries that PID as its SID for life. A sweep then enumerates live processes, calls `Getsid` on each, and reaps any whose SID matches a recorded worker that is confirmed dead by fingerprint. It reuses `unix.Getsid`, `pidFingerprint`, and the existing atomic registry writer. This inverts the model to one durable record per invocation instead of N racing descendant records. The Plan stage owns the actual design.
- **Convention**: a PID-reuse hazard on the worker side was already found and fixed once in #1798 (`workerIdentityStillMatches`). R8 generalizes it to the new path.

## Risks / Dependencies

- **PID reuse ↔ SID collision (R8).** A recycled worker PID that becomes a new worker's session leader shares its SID value with the dead worker's orphans. Identity/age must disambiguate them. This is the highest-risk correctness point, and a wrong answer means killing a live worker's descendants.
- **False kills of developer processes.** A dead session leader is common for unrelated processes. The ownership requirement (R2/AC4) is the mitigation, and AC4 must be a real neutralization test.
- **Process-table scan cost.** Enumerating and calling `Getsid` on every live process costs something on a loaded host. It should reuse the existing shared process-table scan (`sharedProcessTableScan`) rather than adding a redundant scan, given the `ps` and `lsof` CPU-cost findings in ADR-1798 and #1805.
- **Registry growth and write contention.** Worker records add durable writes at every invocation start. They share the existing registry mutex and atomic writer, and must be pruned (R9).
- **Test realism.** AC1/AC2/AC5 require real detached processes with real SIDs. Tests must be `-race`-clean and must not themselves leak processes.
- **Dependency**: builds on #1798's merged code (`descendant_reap_unix.go`, `descendant_registry.go`, `janitor.go`). No other blocking issues.

arbeithand:
Closes #1814

## What this does

#1798's reaper only finds descendants that have a `descendants.json` entry. One spawned in the last seconds before its worker exits, or during a transient `ps` failure, is never recorded, and no path ever reaps it. This PR makes discovery independent of that registry.

- **Worker record at spawn.** `runClaude` now writes one durable record per invocation to `.fabrik/state/workers.json`, right after `cmd.Start()`. It holds the worker's PID and `comm`/`lstart` fingerprint. The record has a unique ID and is never keyed by PID, so a recycled PID cannot overwrite a dead worker's record.
- **SID sweep.** Every process whose `Getsid` equals the PID of a recorded, confirmed-dead worker is reaped. Ownership is by session lineage only, with no `comm`/args allow-list, so zero-children Bash-tool shells are covered.
- **Two call sites.**
  - At invocation end, after `reapTrackedDescendants`, with a fresh uncached scan and a bounded rescan (about 2s).
  - From `runProcessSweepJanitor`, reading all records from disk so orphans from a previous engine run are still reapable.
- **Additive.** `reapTrackedDescendants`, `sweepStaleDescendants`, `trackWorkerDescendants` and `descendants.json` are unchanged. The janitor logs a second summary line and leaves the existing line and its format alone.

## Safety and fail-open behaviour

- No record means no kill, and a dead session leader alone is never enough.
- A live worker with a matching `lstart` is never touched.
- A recycled PID (alive, `lstart` differs) reaps only members that started at or after the dead worker and strictly before the current holder. Equal or unparseable timestamps are skipped.
- Fingerprint errors, an empty recorded fingerprint while the PID is alive, and scan errors all signal nothing and keep the record.
- Each candidate is re-verified (fresh fingerprint and `Getsid`) right before a single-PID `SIGKILL`.
- Records are pruned only after the worker is dead and two consecutive clean scans find no members. At invocation end these are back-to-back rescans; in the janitor, `EmptyScans >= 2` is persisted across passes.

## Tests

- New tests cover AC1 (real `runClaude` with the tracker tick set to an hour), AC2 (record loaded from disk), AC3, AC4 (neutralization: no record, or another PID's record, means no kill), AC5 (zero-children shell), R8 bounds, and R9.
- **AC6:** the sim bed never goes through `runClaude`, so it cannot observe this. AC6 is covered by a real-`runClaude` integration test with mixed detached children, using a `Getsid` enumeration as the oracle.
- `TestMain` points `workerRecordsPathOverride` at a temp file so about 50 existing `InvokeClaude` tests don't write into the package directory.

## Docs

`adrs/1814-registry-independent-worker-session-sweep.md` records why registry-driven discovery alone was insufficient, the PID-reuse handling, and the fail-open gaps. `docs/stage-lifecycle.md` and `docs/state-machine.md` §11.7 are updated, and `docs/llms-full.txt` is regenerated.

## Known gaps

Documented in the ADR: orphans that predate worker records, descendants that call `setsid()` themselves, other Fabrik instances' orphans, the empty-fingerprint residual, and the fork-during-sweep window (the periodic pass is the backstop).


arbeithand:
*branch: fabrik/issue-1814 | commit: 14896264 | main: e4025890d31deaceb8e0fb0ef23e40a97835d06d | 2026-09-20 22:53 UTC*

The worker-session sweep is implemented, tested and pushed on `fabrik/issue-1814`. `go test -race` passes for the new tests and the unmodified #1798 suite (AC8), and `go vet` is clean.

Two failures showed up in the full `go test -race ./...` run:
- **`cmd` `TestExecute_ConfigYAMLApplied`:** it fails the same way on the base commit, so it is not caused by this change.
- **sim `TestReviewAuthorityReinvokesOnChangesRequested`:** it passed when run alone, so I treated it as a flake.

- **Not re-run after the docs commit:** the full `./engine/` and `./tests/sim/` suites. My last attempt timed out and I stopped the retry when you interrupted. The last full run passed engine, and only docs changed after it.
- **Windows build:** `GOOS=windows go build ./...` fails in `engine/poll.go` and `pruefer` on `syscall.Flock`/`syscall.Kill`. I did not check whether that also fails on the base commit. It reports no errors from the new files.
- **Plan checklist:** I could not tick it. The lookup for the Plan stage comment returned no ID, so I skipped that step.

**AC6:** the sim bed never goes through `runClaude`, so it has no worker or SID sweep to check. I met AC6 with a real-`runClaude` integration test that fakes `claude` with a script and uses a `Getsid` enumeration to confirm no member of the dead worker's session survives. Please confirm that interpretation.


Implemented a registry-independent orphan sweep. Each Claude invocation gets a durable worker record at spawn (`workers.json`), and processes whose session ID is a dead recorded worker's PID are reaped at invocation end (fresh scan) and by the proc-janitor. Recycled-PID handling uses start-time bounds and everything fails open. Tests cover AC1–AC6, R8 and R9; the #1798 tests are unmodified and pass. ADR 1814 and the docs are updated, with `llms-full.txt` regenerated. AC6 is met by a real-`runClaude` integration test rather than the sim, and the full engine and sim suites were not re-run after the docs commit.

---
Used 5/100 turns, 5.9M input (72 raw + 5.7M cache-read + 213k cache-write) / 50k output tokens.

arbeithand:
🏭 **Fabrik — stage: Review**
*branch: fabrik/issue-1814 | commit: 14896264 | main: e4025890d31deaceb8e0fb0ef23e40a97835d06d | 2026-09-20 23:04 UTC*

## Review Findings

### Fixed
- No defects found that needed code changes. The worktree is clean, nothing is left to push, and no rebase was needed (0 commits behind `main`).

### Verified
- **Design.**
  - Worker records are keyed by a unique ID, not by PID (R8).
  - The recycled-PID discriminator uses an `lstart` window (lower and upper bounds), and only `lstart`, not `comm`, marks a worker as recycled. Parse failures and equal-second timestamps skip the candidate.
  - Every fingerprint or scan error fails open (R2, R8, R9).
  - An empty-fingerprint record is never swept while its PID is alive (R9).
  - Each candidate is re-verified (fresh fingerprint, `Getsid`, bounds) before a single-PID `SIGKILL`.
  - The invocation-end sweep uses a fresh, uncached scan and rescans, pruning the record only after two clean scans with nothing skipped (R5).
  - The periodic sweep reads records from disk (R4) and prunes at `EmptyScans >= 2`.
  - Each reap logs the `[#N kill]` line required by R7.
  - The #1798 files (`descendant_reap_unix.go`, `descendant_registry.go`, `procattr_unix.go`) are untouched (R6/AC8).
- **Neutralization, run by me.**
  - Mutation: `classifyWorker` treats an empty-fingerprint live PID as dead. The R9/classify tests go red.
  - Mutation: removing the `want[sid]` filter in `collectSessionMembers` does not fail AC4. That filter is only an optimisation. Ownership is enforced by the per-record `bySID[r.PID]` lookup plus the `sid != rec.PID` re-check in `reapWorkerSessionMembers`. The AC4 test covers the behaviour with a record-less session and a record for a different PID, so I left the test as is.
  - I restored the source after each mutation.
- **Tests.** The new worker sweep, registry and janitor tests plus the #1798 suites pass with `-race` (229 in the targeted run). `go build ./...` and `go vet ./...` are clean.
- **Full engine + sim run.** Three failures appeared and none come from this change:
  - `TestInvokeClaude_FabrikEnvVarsInjected` fails only because this Review session's environment has `FABRIK_PR` set. It passes with `env -u FABRIK_PR`.
  - `TestPartialMutation_SpawnChildren_ProjectAddFails` and `TestNoOpCommentCycles_TripsBreakerAndResetsOnProgress` failed under the full run but pass when run alone, so I treat them as load flakes.
- **Windows.** `GOOS=windows go build ./engine/` fails only on `syscall.Flock` in `engine/poll.go`, which this branch doesn't touch. The new files produce no errors.
- **Docs.** Re-running `scripts/generate-llms-full.sh` leaves the tree clean, so `llms-full.txt` matches the docs. The ADR states why registry-driven discovery was insufficient and records the PID-reuse handling and the fail-open gaps (AC7).
- **AC6.** I agree with the interpretation. The sim bed doesn't go through `runClaude`, so a real-`runClaude` integration test with a `Getsid` enumeration is the meaningful check. The maintainer may still want to confirm it.

### Notes for the human reviewer
- Each invocation end now costs at least two fresh `ps` scans plus a 50ms gap.
- The empty-fingerprint residual and the other fail-open gaps are documented in the ADR, not fixed.

Reviewed the worker-session sweep. The design is sound: it fails open on every ambiguity, handles recycled PIDs with an `lstart` window, and leaves the #1798 code untouched. I found nothing that needed changing. I ran two mutation checks, and the R9 mutation is caught by the tests. The ownership guard is the per-record lookup, and AC4 passes regardless of the `want` filter in `collectSessionMembers`. Build, vet and the new and #1798 tests pass with `-race`. Three failures in the full engine and sim run are unrelated (one is this session's ambient `FABRIK_PR`, two pass in isolation). The PR is ready for human review.

---
Used 18/100 turns, 2.0M input (32 raw + 1.9M cache-read + 130k cache-write) / 7k output tokens.

arbeithand:
*branch: fabrik/issue-1814 | commit: 61de7d37 | main: e4025890d31deaceb8e0fb0ef23e40a97835d06d | 2026-09-20 23:06 UTC*

I fixed the corrupt `workers.json` finding from Pruefer, pushed it as `61de7d37`, and resolved the thread. The new test and the worker-sweep tests pass with `-race`. I did not re-run the full engine suite after this change.

- **Finding:** an unparseable `workers.json` made every `appendWorkerRecord` and `allWorkerRecords` call fail, which silently disabled the sweep for that repo until someone deleted the file by hand.
- **Fix:** `loadWorkerRecords` in `engine/worker_registry.go` now renames an unparseable file to `workers.json.corrupt`, logs a warning, and carries on with an empty set. If the rename itself fails, the error is still returned.
- **Trade-off:** the records in the quarantined file are lost, so their orphans won't be reaped (a leak, never a false kill). The file is kept for inspection.
- **Test and docs:** `TestWorkerRecords_CorruptFileIsQuarantined` writes a truncated file, checks it is quarantined and preserved, then confirms a new record persists afterwards. The new gap is listed under "Known gaps" in ADR 1814 and in `docs/state-machine.md` §11.7, and `docs/llms-full.txt` is regenerated.
- **Security:** the finding is a reliability issue, not a security one.

The bot's summary review found nothing else to act on.

---
**Threads addressed:**
- `engine/worker_registry.go:78` — resolved

arbeithand:
*branch: fabrik/issue-1814 | commit: 61de7d37 | main: e4025890d31deaceb8e0fb0ef23e40a97835d06d | 2026-09-20 23:47 UTC*

## Validation Report

### Requirements: 10/10 R and 8/8 AC verified
- **R1, R5 (registry-independent discovery, race closed):** `sweepWorkerSessionAtInvocationEnd` runs a fresh uncached scan and reaps by session ID. `sweepOrphanedWorkerSessions` runs periodically. Both are covered by AC1 and AC2 tests using real detached children.
- **R2, AC4 (ownership):** a process is signalled only when its session ID equals a recorded worker's PID and that worker is dead. The neutralization tests cover a record-less dead-leader session and a record for a different PID. They stay green with the `want[sid]` filter removed, because that filter is only an optimisation and ownership is enforced by the per-record lookup.
- **R3, AC5:** ownership is by session ID alone, with no comm/args allow-list, and a zero-child shell orphan is reaped in tests.
- **R4:** records are stored in `workers.json` and the periodic sweep loads them from disk.
- **R6, AC8:** `descendant_reap_unix.go`, `descendant_registry.go` and `procattr_unix.go` are unmodified, and the #1798 suites pass.
- **R7:** each reap logs the `[#N kill]` line with pid, comm, worker PID and reason.
- **R8, AC3:** worker records have unique IDs, and recycled PIDs are handled with an `lstart` window. Parse failures, equal timestamps and fingerprint errors fail open.
- **R9:** an empty-fingerprint record with a live PID is never swept. Pruning needs a dead worker, clean scans and no skips. A corrupt `workers.json` is quarantined.
- **R10, AC7:** `docs/stage-lifecycle.md`, `docs/state-machine.md` §11.7 and ADR 1814 are updated. Re-running `scripts/generate-llms-full.sh` leaves the tree clean.
- **AC6:** met by a real-`runClaude` integration test with a `Getsid` enumeration, not the sim bed, because the sim never goes through `runClaude`. This is still the maintainer's interpretation to confirm.

### Test Suite
- **Build and vet:** `go build ./...` and `go vet ./...` are clean.
- **Targeted engine run:** worker-sweep, registry, janitor and #1798 tests plus `TestInvokeClaude_FabrikEnvVarsInjected` pass with `-race` in 84.6s. That env test needs `FABRIK_PR` and `FABRIK_ISSUE` unset in this session.
- **Full `./...` run:**
  - `cmd` `TestExecute_ConfigYAMLApplied` fails identically on `origin/main`, so it is not caused by this branch.
  - The `tests/sim` failures (`TestMergeTrainLanding_OneAtATimeViaInteractionOnlyPoison`, `TestRestartEnv_RoundTrip`, a timeout) differed between runs and passed together in isolation. I treat them as load flakes.
  - The last full `./engine/` run I completed with a 9-minute timeout had no failures. A later 5-minute run hit its own timeout under load, and I did not repeat a full engine pass after that (my 9-minute re-run was interrupted).
- **Windows:** not re-checked in this stage. The Review stage found `GOOS=windows go build ./engine/` fails only on `syscall.Flock` in `engine/poll.go`, which this branch doesn't touch.

The #1798 code is untouched and its tests pass.

### Verdict: READY TO MERGE
Rebase outcome: `skipped-up-to-date` (0 commits behind `main`, not in the merge queue). PR `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`. The worktree is clean and matches the remote.

Rebase: skipped-up-to-date. PR mergeable: MERGEABLE, mergeStateStatus: CLEAN. Required checks: Test and vet: pass, Verify llms-full.txt is up to date: pass, Analyze (go): pass (two entries, both pass). Requirements: R1–R10 and AC1–AC8 verified against the spec, with AC6 met by a real-`runClaude` integration test rather than the sim. Local: build and vet clean, and the targeted worker-sweep and #1798 tests pass with `-race`. In the full run, `cmd` `TestExecute_ConfigYAMLApplied` also fails on `origin/main`, and the sim failures varied between runs and passed in isolation. I did not complete a final full `./engine/` run after the last docs and fix commits.

---
Used 2/100 turns, 1.5M input (26 raw + 1.3M cache-read + 148k cache-write) / 4k output tokens.

arbeithand:
🏭 **Fabrik merge-train** — Landed via singleton fast path PR #1824. That is this PR: the pinned base was already an ancestor of this head, this PR was mergeable, and its own CI was green and complete, so it was landed directly — no trial branch was assembled and no separate integration PR exists. See ADR-1644.

